PR #265, opened 2024-05-31
Add magic routing hint support

### LINKED ISSUE #243: Unable to initiate payment between two sdk-liquid wallets
Reason is because boltz doesn't allow submarine swap to a reverse swap payment. 
For that boltz introduced the magic hints: https://docs.boltz.exchange/v/api/magic-routing-hints
In short When we detect such hints we pay directly to an onchain liquid address.

ok300:
- [x] Receive Support: `receive_payment` creates an invoice with a magic routing hint.
- [x] Send Support: `send_payment` recognizes magic routing hints in invoices and, if present, defaults to onchain L-BTC payments

Closes #243

---

TODOs
- [x] Send: `PrepareSendResponse` should return the onchain fee, if the invoice has MRH
- [x] Send: don't allow self-transfer (check by invoice)
- [x] Send: onchain tx fees are estimated to 251 sats, but the actual fees are 250 sats
  - #293
- [x] Send: store onchain tx fee. If no swap present (e.g. on restore), use that as payment fee.
- [x] Receive: multiple payments can be made via direct onchain tx to the MRH of a single Receive Swap, which result in multiple Receive Payments. Show them in `list_payments` as separate Receive Payments (e.g. with no `swap_id`) and 0 fees.
- [x] Send / Receive: reflect correct payment state when using direct onchain tx
  - tx in mempool = Pending Payment (emit event, transition payment state)
  - tx is confirmed = Completed Payment (emit event, transition payment state)
- Send / Receive: emit correct events when using direct onchain tx
  - tx in mempool = emit `PaymentPending` event
    - [x] Send
    - [x] Receive
  - tx is confirmed = emit `PaymentSucceed` event
    - [x] Send
    - [x] Receive

Open issues
- [x] Receive: create new Receive Swaps using same MRH (same address, amount)


ok300:
> There is also the option to validate the routing hint is correctly added to the invoice

@dangeross good idea, done: 13e99e2c0d7b7f40fef7166658f45deaf82f2cb8

lib/core/src/sdk.rs
```diff
@@ -689,7 +708,58 @@ impl LiquidSdk {
...
+                    payment_type: PaymentType::Send,
+                    is_confirmed: false,
+                };
+                self.persister.insert_or_update_payment(tx_data.clone())?;
+                self.emit_payment_updated(Some(tx_id)).await?; // Emit Pending event
+
+                Ok(SendPaymentResponse {
+                    payment: Payment::from(tx_data, None),
+                })
+            }
```
roeierez, line 751: Since the send_payment_via_swap was extracted to its own method perhaps it make sense to extract this block to send_onchain_payment ?

lib/core/src/sdk.rs
```diff
@@ -689,7 +708,58 @@ impl LiquidSdk {
...
+
+            // If no MRH found, perform usual swap
+            None => self.send_payment_via_swap(req).await,
+        }
+    }
+
+    /// Performs a Send Payment by doing a swap (create it, fund it, track it, etc).
+    async fn send_payment_via_swap(
```
roeierez, line 759: NIT: `send_swap_payment` (vs `send_onchain_payment`)

roeierez (changes requested):
Looks good, dropped some comments.

lib/core/src/send_swap.rs
```diff
@@ -72,6 +72,7 @@ impl SendSwapStateHandler {
                             tx_id: lockup_tx_id.clone(),
                             timestamp: None,
                             amount_sat: swap.payer_amount_sat,
+                            fees_sat: 0, // Will be overwritten once LWK fetches the real tx
```
roeierez, line 75: Why not writing here the real fee, we create the tx so we should know it.

roeierez (changes requested):

lib/core/src/sdk.rs
dangeross: Maybe `send_onchain_payment` should be saved for the interface for sending via chain swap. What do you think?

roeierez: I had that thought, indeed a bit confusing.
The real problem here is that we are dealing with multiple chains on different swaps right?
Perhaps the the chain swap to BTC should be more explicit? e.g `send_to_btc` `receive_via_btc` ?
Anyway we don't have to decide now and my suggestion was about a private method so I am happy with any option as long as it is use the same pattern of the swap version.
For example we can leave this as `send_payment_via_swap` and use `send_payment_via_mrh` or similar.

lib/core/src/sdk.rs
ok300: Done: 66443586567070da511d565f8cbdc063141fe8d9

lib/core/src/sdk.rs
ok300: Done, extracted to `send_payment_via_mrh` for consistency: 66443586567070da511d565f8cbdc063141fe8d9

lib/core/src/send_swap.rs
ok300: Done: cabe96f48dc6d86d5a0aea9a32ae3b59c5297a0f

Although I don't think it brings much. This is a Send via swap, meaning the fees are anyway stored in the Swap table, so `list_payments` retrieves them from there. This field here is simply fallback for when we don't have the swap data and we need to show the tx fees, like for direct onchain payments. That case is covered here

https://github.com/breez/breez-liquid-sdk/pull/265/files#diff-f09a3f89b9a7fd7422e34d3a519a8ab878445ab33debc6bc60870fd348347a90R670-R679

and it already sets the tx fees.

dangeross (approved):
LGTM!

roeierez (approved):
LGTM

PR #287, opened 2024-06-03
Abstract swapper stream as part of swapper trait

roeierez:
This PR moves the boltz status stream behind its own trait and as part of the swapper API.
Also took the opportunity to move persistent access out of the swapper trait by adding a ReconectHandler so the user of the trait has the opportunity (hook) to subscribe new list on every reconnection.


dangeross (approved):
LGTM!

Issue #285, opened 2024-06-03
Split to testable modules


roeierez:
- [x] Move swapper to its own trait
- [x] Move boltz stream as part of swapper API
- [x] Move persister dependency out of Boltz stream
- [ ] SendSwapHandler - for implementing the state machine for send swaps
- [ ] ReceiveSwapHandler - for implementing the state machine for receive swaps
- [ ] Abstract Electurm as ChainService

PR #291, opened 2024-06-05
Abstract onchain wallet behind its own module and trait

roeierez:
As an ongoing effort and part of #285 now it is the onchain wallet that gets abstracted away.

### DIFF
--- a/lib/core/src/wallet.rs
+++ b/lib/core/src/wallet.rs
@@ -1,0 +1,132 @@
+use anyhow::{anyhow, Result};
+use boltz_client::ElementsAddress;
+use lwk_common::{singlesig_desc, Singlesig};
+use lwk_signer::{AnySigner, SwSigner};
+use std::{str::FromStr, sync::Arc};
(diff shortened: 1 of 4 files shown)


lib/core/src/wallet.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /// Get the current tip of the blockchain the wallet is aware of
+    async fn tip(&self) -> Tip {
+        self.wallet.lock().await.tip()
+    }
+
+    /// Get the public key of the wallet
+    async fn pubkey(&self) -> String {
+        self.lwk_signer.xpub().to_string()
```
dangeross, line 134: Seems like the CI is failing because of this, should it be?
```suggestion
        self.lwk_signer.xpub().public_key.to_string()
```

dangeross: also I think the function doesn't really need to be `async`

dangeross (approved):
LGTM

ok300 (approved):

PR #292, opened 2024-06-05
Receive: Add zero-conf checks and properly manage state updates

### LINKED ISSUE #187: Receive (0-conf): Add pre-conditions before broadcasting claim tx
On Receive, currently broadcast the claim tx as soon as the lockup tx is in the mempool.

We should add a few checks before doing this, like an upper bound on the swap amount.

If any of these checks fail, we fallback to only broadcasting the claim tx after the lockup tx is confirmed.

### LINKED ISSUE #227: Show Receive as pending in `list-payments` even when we don't broadcast the claimtx immediately
Right now we insert a pseudo-tx for the claim

https://github.com/breez/breez-liquid-sdk/blob/1777a7bfe2cdb32b9325ef1deae3c1c7cb39a7af/lib/core/src/sdk.rs#L839-L847

which creates a `payment_tx_data` entry for the (now known) claim tx id.

However this only happens when we claim, and right now we claim immediately after seeing the lockup tx in the mempool.

If we later decide to wait for the lockup tx to be confirmed before claiming, then the pending Receive payment will not be included in `list_payments` until the lockup is confirmed (up to 1 min). We should make sure to include the pending Receive payment even in this case.

hydra-yse:
Closes #187.

### DIFF
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -11,6 +11,10 @@ use rusqlite::ToSql;
 use serde::{Deserialize, Serialize};
 
 use crate::error::PaymentError;
+use crate::receive_swap::{
+    DEFAULT_ZERO_CONF_MAX_SAT, DEFAULT_ZERO_CONF_MIN_FEE_RATE_MAINNET,
+    DEFAULT_ZERO_CONF_MIN_FEE_RATE_TESTNET,
+};
 use crate::utils;
 
 /// Configuration for the Liquid SDK
@@ -25,7 +29,13 @@ pub struct Config {
     pub network: Network,
     /// Send payment timeout. See [crate::sdk::LiquidSdk::send_payment]
     pub payment_timeout_sec: u64,
+    /// Zero-conf minimum accepted fee-rate in sat/vbyte
+    pub zero_conf_min_fee_rate: f32,
+    /// Maximum amount in satoshi to accept zero-conf payments with
+    /// Defaults to [crate::receive_swap::DEFAULT_ZERO_CONF_MAX_SAT]
+    pub zero_conf_max_amount_sat: Option<u64>,
 }
+
 impl Config {
     pub fn mainnet() -> Self {
         Config {
@@ -34,6 +44,8 @@ impl Config {
             working_dir: ".".to_string(),
             network: Network::Mainnet,
             payment_timeout_sec: 15,
+            zero_conf_min_fee_rate: DEFAULT_ZERO_CONF_MIN_FEE_RATE_MAINNET,
+            zero_conf_max_amount_sat: None,
         }
     }
 
@@ -44,7 +56,14 @@ impl Config {
             working_dir: ".".to_string(),
             network: Network::Testnet,
             payment_timeout_sec: 15,
+            zero_conf_min_fee_rate: DEFAULT_ZERO_CONF_MIN_FEE_RATE_TESTNET,
+            zero_conf_max_amount_sat: None,
         }
+    }
+
+    pub fn zero_conf_max_amount_sat(&self) -> u64 {
+        self.zero_conf_max_amount_sat
+            .unwrap_or(DEFAULT_ZERO_CONF_MAX_SAT)
     }
 }
 
--- a/lib/core/src/receive_swap.rs
+++ b/lib/core/src/receive_swap.rs
@@ -6,15 +6,22 @@ use boltz_client::swaps::boltzv2;
 use log::{debug, error, info, warn};
 use tokio::sync::broadcast;
 
-use crate::ensure_sdk;
 use crate::model::PaymentState::{Complete, Created, Failed, Pending, TimedOut};
-use crate::model::{PaymentTxData, PaymentType, ReceiveSwap};
+use crate::model::{Config, PaymentTxData, PaymentType, ReceiveSwap};
+use crate::{ensure_sdk, utils};
 use crate::{
     error::PaymentError, model::PaymentState, persist::Persister, swapper::Swapper,
     wallet::OnchainWallet,
 };
 
+/// The minimum acceptable fee rate when claiming using zero-conf
+pub const DEFAULT_ZERO_CONF_MIN_FEE_RATE_TESTNET: f32 = 0.1;
+pub const DEFAULT_ZERO_CONF_MIN_FEE_RATE_MAINNET: f32 = 0.01;
+/// The maximum acceptable amount in satoshi when claiming using zero-conf
+pub const DEFAULT_ZERO_CONF_MAX_SAT: u64 = 100_000;
+
 pub(crate) struct ReceiveSwapStateHandler {
+    config: Config,
     onchain_wallet: Arc<dyn OnchainWallet>,
     persister: Arc<Persister>,
     swapper: Arc<dyn Swapper>,
@@ -23,12 +30,14 @@ pub(crate) struct ReceiveSwapStateHandler {
 
 impl ReceiveSwapStateHandler {
     pub(crate) fn new(
+        config: Config,
         onchain_wallet: Arc<dyn OnchainWallet>,
         persister: Arc<Persister>,
         swapper: Arc<dyn Swapper>,
     ) -> Self {
         let (subscription_notifier, _) = broadcast::channel::<String>(30);
         Self {
+            config,
             onchain_wallet,
             persister,
             swapper,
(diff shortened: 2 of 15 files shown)


lib/core/src/sdk.rs
```diff
@@ -515,10 +549,70 @@ impl LiquidSdk {
...
+                    return Err(anyhow!("Could not find lockup transaction in the mempool"));
+                };
+
+                let electrum_client = self.config.get_electrum_client()?;
+                let transactions = electrum_client.get_transactions(&[lockup_tx_history.txid])?;
+                let Some(lockup_tx) = transactions.first() else {
+                    return Err(anyhow!("Could not find lockup transaction in the mempool"));
+                };
+
+                if swap_state == RevSwapStates::TransactionMempool.to_string() {
```
roeierez, line 568: I think it is better to handle the TransactionMempool in its own case branch.

lib/core/src/sdk.rs
```diff
@@ -44,9 +45,10 @@ use crate::{
     utils,
 };
 
-/// Claim tx feerate, in sats per vbyte.
-/// Since the  Liquid blocks are consistently empty for now, we hardcode the minimum feerate.
-pub const LIQUID_CLAIM_TX_FEERATE_MSAT: f32 = 100.0;
+/// The minimum acceptable fee rate when claiming using zero-conf
+pub const LIQUID_SDK_ZERO_CONF_FEE_RATE: f32 = 100.0;
```
roeierez, line 49: fee rate is usually expressed in sat per vbyte. Was the intention to be 0.1 sat per vbyte?

lib/core/src/sdk.rs
```diff
@@ -515,10 +549,70 @@ impl LiquidSdk {
...
+                        debug!("Lockup transaction signals RBF. Waiting for confirmation...");
+                        return Ok(());
+                    }
+
+                    // If the fees are higher than our estimated value
+                    let tx_fees: u64 = lockup_tx
+                        .all_fees()
+                        .values()
+                        .sum();
+                    let lower_bound_estimated_fees = lockup_tx.vsize() as f32 * LIQUID_SDK_ZERO_CONF_FEE_RATE / 1000.0 * 0.8;
```
roeierez, line 601: Doesn'tn boltz send lowball here? Does this condition allow it?

roeierez (changes requested):

lib/core/src/sdk.rs
```diff
@@ -44,9 +45,10 @@ use crate::{
     utils,
 };
 
-/// Claim tx feerate, in sats per vbyte.
-/// Since the  Liquid blocks are consistently empty for now, we hardcode the minimum feerate.
-pub const LIQUID_CLAIM_TX_FEERATE_MSAT: f32 = 100.0;
+/// The minimum acceptable fee rate when claiming using zero-conf
+pub const LIQUID_SDK_ZERO_CONF_FEE_RATE: f32 = 100.0;
+/// The maximum acceptable amount in satoshi when claiming using zero-conf
+pub const LIQUID_SDK_ZERO_CONF_LIMIT_SAT: u64 = 1_000;
```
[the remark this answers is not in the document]
hydra-yse: My bad, pushed to do some tests with Carlos, will fix soon

lib/core/src/sdk.rs
hydra-yse: Indeed, to avoid rounding errors, we represent it in msat/vbyte

lib/core/src/sdk.rs
hydra-yse: Agreed, we should probably have two cases for mainnet/testnet for now, one with 0.1 sat/vbyte and another for 0.01

lib/core/src/sdk.rs
```diff
@@ -456,20 +457,92 @@ impl LiquidSdk {
...
+                if receiver_amount_sat > LIQUID_SDK_ZERO_CONF_LIMIT_SAT {
+                    debug!("[Receive Swap {id}] Amount is too high to claim with zero-conf ({receiver_amount_sat} sat > {LIQUID_SDK_ZERO_CONF_LIMIT_SAT} sat). Waiting for confirmation...");
+                    return Ok(());
+                }
+
+                debug!("[Receive Swap {id}] Amount is within valid range for zero-conf ({receiver_amount_sat} < {LIQUID_SDK_ZERO_CONF_LIMIT_SAT} sat)");
+
+                // If the transaction has RBF, see https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki
+                let rbf_explicit = lockup_tx.input.iter().any(|input| input.sequence.is_rbf());
+                // let rbf_implicit = lockup_tx_history.height < 0;
```
ok300, line 496: Should this be enabled? Or was it a test and can be removed?

lib/core/src/sdk.rs
```diff
@@ -456,20 +457,92 @@ impl LiquidSdk {
...
+                }
+
+                debug!("[Receive Swap {id}] Lockup tx fees are within acceptable range ({tx_fees} > {lower_bound_estimated_fees} sat). Proceeding with claim.");
+
+                match receive_swap.claim_tx_id {
+                    Some(claim_tx_id) => {
+                        warn!("Claim tx for Receive Swap {id} was already broadcast: txid {claim_tx_id}")
+                    }
+                    None => {
+                        self.try_handle_receive_swap_update(&receive_swap.id, Pending, None)
```
ok300, line 525: Isn't the correct method here `ReceiveSwapStateHandler::update_swap_info`?

ok300, line 525: Should this state transition and event be triggered before our 0-conf preconditions?

In other words, as soon as we see the lockup tx in the mempool?

lib/core/src/sdk.rs
```diff
@@ -456,20 +457,92 @@ impl LiquidSdk {
...
+                        }
+                    }
+                };
+                Ok(())
+            }
             // The lockup tx is confirmed => try to claim
-            | RevSwapStates::TransactionConfirmed) => {
+            Ok(RevSwapStates::TransactionConfirmed) => {
+                // We ensure that the lockup tx is actually confirmed
+                // else if lockup_tx_history.height <= 0 {
```
ok300, line 543: Test code that needs to be removed?

ok300 (commented):
Overall looks good, but had some comments and questions.

Before you make too many changes, I'd suggest merging the latest `main` into your branch.

lib/core/src/model.rs
```diff
@@ -735,3 +735,68 @@ macro_rules! get_invoice_amount {
...
+        event: String,
+        channel: String,
+        args: Vec<Update>,
+    },
+    Error {
+        event: String,
+        channel: String,
+        args: Vec<RespError>,
+    },
+}
```
roeierez, line 802: It seems that all these models belong to the swapper module.

roeierez (changes requested):

lib/core/src/model.rs
roeierez: @hydra-yse  actually I see they are copied from boltz client, any reason why not fix there?

ok300: I asked him to first do it here and get it working.

Then @hydra-yse this might be a good time to add it upstream and include it in our patch branch.

hydra-yse:
@ok300 About the test code comments, it's something that needs further discussion:
1. In order to check for implicit RBF, we need to check the tx ancestors and make sure they are confirmed. This could be easily done via electrum by checking the scripts history, but currently we cannot do that (see #297)
2. Same issue in regards to the `TransactionConfirmed` state, I think it's good if we actually validate against our chain explorer if the tx has been confirmed.

So just keeping it there for reference. Will add a TODO to both so they can be changed in the near future.

lib/core/src/sdk.rs
hydra-yse: The switch to pending state has been removed and handled in a separate PR (#301)

lib/core/src/sdk.rs
```diff
@@ -211,20 +207,21 @@ impl LiquidSdk {
...
+                        Ok(update) => {
                             let _ = cloned.sync().await;
-                            match cloned.persister.fetch_send_swap_by_id(&id) {
+                            let id = update.id();
+                            match cloned.persister.fetch_send_swap_by_id(id) {
                                 Ok(Some(_)) => {
-                                    match cloned.send_swap_state_handler.on_new_status(&status, &id).await {
-                                        Ok(_) => info!("Successfully handled Send Swap {id} update"),
+                                    match cloned.send_swap_state_handler.on_new_status(&update).await {
+                                        Ok(_) => info!("Succesfully handled Send Swap {id} update"),
```
dangeross, line 216: I think this was reverted in the rebase

lib/core/src/sdk.rs
```diff
@@ -211,20 +207,21 @@ impl LiquidSdk {
...
                                     }
                                 }
                                 _ => {
-                                    match cloned.persister.fetch_receive_swap(&id) {
+                                    match cloned.persister.fetch_receive_swap(id) {
                                         Ok(Some(_)) => {
-                                            match cloned.receive_swap_state_handler.on_new_status(&status, &id).await {
-                                                Ok(_) => info!("Successfully handled Receive Swap {id} update"),
+                                            match cloned.receive_swap_state_handler.on_new_status(&update).await {
+                                                Ok(_) => info!("Succesfully handled Receive Swap {id} update"),
```
dangeross, line 224: This too

dangeross (approved):

lib/core/src/receive_swap.rs
```diff
@@ -2,18 +2,26 @@ use std::{str::FromStr, sync::Arc};
...
 use crate::{
     error::PaymentError, model::PaymentState, persist::Persister, swapper::Swapper,
     wallet::OnchainWallet,
 };
 
+/// The minimum acceptable fee rate when claiming using zero-conf
+pub const LIQUID_SDK_ZERO_CONF_FEE_RATE_TESTNET: f32 = 0.1;
+pub const LIQUID_SDK_ZERO_CONF_FEE_RATE_MAINNET: f32 = 0.01;
+/// The maximum acceptable amount in satoshi when claiming using zero-conf
+pub const LIQUID_SDK_ZERO_CONF_LIMIT_SAT: u64 = 100_000;
```
roeierez, line 21: Perhaps these parameters can be as part of the config?
`zero_conf_min_feerate`
`zero_conf_max_amount`

lib/core/src/receive_swap.rs
```diff
@@ -49,46 +62,118 @@ impl ReceiveSwapStateHandler {
...
+                };
+
+                let lockup_tx = utils::deserialize_tx_hex(&transaction.hex)?;
+
+                // If the amount is greater than the zero-conf limit
+                // TODO Make limit user-definable through config
+                let receiver_amount_sat = receive_swap.receiver_amount_sat;
+                if receiver_amount_sat > LIQUID_SDK_ZERO_CONF_LIMIT_SAT {
+                    debug!("[Receive Swap {id}] Amount is too high to claim with zero-conf ({receiver_amount_sat} sat > {LIQUID_SDK_ZERO_CONF_LIMIT_SAT} sat). Waiting for confirmation...");
+                    return Ok(());
```
roeierez, line 90: This will return from the function, did you mean `Ok(())` ?
There are more variations of these bellow.

lib/core/src/receive_swap.rs
```diff
@@ -49,46 +62,118 @@ impl ReceiveSwapStateHandler {
...
+                let lower_bound_estimated_fees = lockup_tx.vsize() as f32 * min_fee_rate * 0.8;
+
+                if lower_bound_estimated_fees > tx_fees as f32 {
+                    debug!("[Receive Swap {id}] Lockup tx fees are too low: Expected at least {lower_bound_estimated_fees} sat, got {tx_fees} sat. Waiting for confirmation...");
+                    return Ok(());
+                }
+
+                debug!("[Receive Swap {id}] Lockup tx fees are within acceptable range ({tx_fees} > {lower_bound_estimated_fees} sat). Proceeding with claim.");
+
+                match receive_swap.claim_tx_id {
```
roeierez, line 122: Seems validating whether we have or not the claim tx should be at the start of this branch. It is a bit weird to validate the lockup zero conf before we verify that point.

roeierez (changes requested):

lib/core/src/model.rs
```diff
@@ -26,7 +30,13 @@ pub struct Config {
     pub network: Network,
     /// Send payment timeout. See [crate::sdk::LiquidSdk::send_payment]
     pub payment_timeout_sec: u64,
+    /// Zero-conf minimum accepted fee-rate in sat/vbyte
+    pub zero_conf_min_fee_rate: f32,
+    /// Maximum amount in satoshi to accept zero-conf payments with
+    /// Defaults to [crate::receive_swap::DEFAULT_ZERO_CONF_LIMIT_SAT]
```
ok300, line 36:
```suggestion
    /// Defaults to [crate::receive_swap::DEFAULT_ZERO_CONF_MAX_SAT]
```

lib/core/src/receive_swap.rs
```diff
@@ -49,46 +62,117 @@ impl ReceiveSwapStateHandler {
...
+                }
+
+                let boltzv2::Update::TransactionMempool { transaction, .. } = update else {
+                    return Err(anyhow!("Unexpected payload from Boltz status stream"));
+                };
+
+                let lockup_tx = utils::deserialize_tx_hex(&transaction.hex)?;
+
+                // If the amount is greater than the zero-conf limit
+                // TODO Make limit user-definable through config
```
ok300, line 93:
```suggestion
```

This TODO was addressed, so it can be removed.

lib/core/src/receive_swap.rs
```diff
@@ -49,46 +62,117 @@ impl ReceiveSwapStateHandler {
...
+                let max_amount_sat = self.config.zero_conf_max_amount_sat();
+                let receiver_amount_sat = receive_swap.receiver_amount_sat;
+                if receiver_amount_sat > max_amount_sat {
+                    debug!("[Receive Swap {id}] Amount is too high to claim with zero-conf ({receiver_amount_sat} sat > {max_amount_sat} sat). Waiting for confirmation...");
```
ok300, line 97: Maybe this should be an `info` or even a `warn`, as it's a very useful detail about the flow of the swap.

lib/core/src/receive_swap.rs
```diff
@@ -49,46 +62,117 @@ impl ReceiveSwapStateHandler {
...
+
+                debug!("[Receive Swap {id}] Amount is within valid range for zero-conf ({receiver_amount_sat} < {max_amount_sat} sat)");
+
+                // If the transaction has RBF, see https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki
+                // TODO: Check for inherent RBF by ensuring all tx ancestors are confirmed
+                let rbf_explicit = lockup_tx.input.iter().any(|input| input.sequence.is_rbf());
+                // let rbf_inherent = lockup_tx_history.height < 0;
+
+                if rbf_explicit {
+                    debug!("[Receive Swap {id}] Lockup transaction signals RBF. Waiting for confirmation...");
```
ok300, line 109:
```suggestion
                    warn!("[Receive Swap {id}] Lockup transaction signals RBF. Waiting for confirmation...");
```

lib/core/src/receive_swap.rs
```diff
@@ -49,46 +62,117 @@ impl ReceiveSwapStateHandler {
...
+
+                debug!("[Receive Swap {id}] Lockup tx does not signal RBF. Proceeding...");
+
+                // If the fees are higher than our estimated value
+                let tx_fees: u64 = lockup_tx.all_fees().values().sum();
+                let min_fee_rate = self.config.zero_conf_min_fee_rate;
+                let lower_bound_estimated_fees = lockup_tx.vsize() as f32 * min_fee_rate * 0.8;
+
+                if lower_bound_estimated_fees > tx_fees as f32 {
+                    debug!("[Receive Swap {id}] Lockup tx fees are too low: Expected at least {lower_bound_estimated_fees} sat, got {tx_fees} sat. Waiting for confirmation...");
```
ok300, line 121:
```suggestion
                    warn!("[Receive Swap {id}] Lockup tx fees are too low: Expected at least {lower_bound_estimated_fees} sat, got {tx_fees} sat. Waiting for confirmation...");
```

PR #293, opened 2024-06-05
Adjust onchain fee estimation based on destination address type

ok300:
This PR fixes several issues around onchain tx fee estimation.

We used to call this the "broadacast fee", but that is misleading because it doesn't specify the target address type. Different destination address types can result in different fees, so using the same method to estimate fees regardless of destination type is not a good idea.

This PR splits `get_broadcast_fee_estimation(amount)` into
- `estimate_onchain_tx_fee(amount, address)` for any destination `address` type
- `estimate_onchain_tx_fee_p2tr(amount)` specifically for taproot destinations
- `estimate_lockup_tx_fee(amount)` which is a utility wrapper pointing to the above

This fixes the refund tx fee estimation. We used to estimate assuming a taproot address, but we're refunding to our `next_unused_address` which is not taproot (it's currently a BIP-84 WPKH).

This fix also addresses fee estimation issues relevant for #265.


hydra-yse (approved):
LGTM

lib/core/src/sdk.rs
```diff
@@ -802,21 +802,31 @@ impl LiquidSdk {
...
+            .values()
+            .sum())
+    }
+
+    async fn estimate_lockup_tx_fee(&self, amount_sat: u64) -> Result<u64> {
+        self.estimate_onchain_tx_fee_p2tr(amount_sat).await
+    }
+
+    /// Estimate the onchain fee for sending the given amount to a taproot address
+    async fn estimate_onchain_tx_fee_p2tr(&self, amount_sat: u64) -> Result<u64> {
```
roeierez, line 820: Not sure we need this function, it is only use inside `estimate_lockup_tx_fee` perhaps inline it there?

ok300: Done: 6e705dc5ec54dbcdfca239729a5559d247fe7936

roeierez (approved):
LGTM

PR #295, opened 2024-06-05
Send swap handler

roeierez:
Here the send swaps state machine handling is extracted out to its own module with dependencies that can be mocked and allow us to write tests.
Also it removes the details from the sdk facade layer.
Once this is reviewed and merged, next in the line is the receive swaps handler.

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -174,8 +174,19 @@ impl LiquidSdk {
                 cloned.persister.clone(),
                 cloned.swapper.clone(),
             );
+            let mut swap_state_changes = send_swap_state_handler.subscribe_payment_updates();
             loop {
                 tokio::select! {
+                    payment_id = swap_state_changes.recv() => {
+                      match payment_id {
+                          Ok(payment_id) => {
+                            if let Err(e) = cloned.emit_payment_updated(Some(payment_id)).await {
+                              error!("Failed to emit payment update: {e:?}");
+                            }
+                          }
+                          Err(e) => error!("Failed to receive send swap state change: {e:?}")
+                      }
+                    }
                     update = updates_stream.recv() => match update {
                         Ok(boltzv2::Update { id, status }) => {
                             let _ = cloned.sync().await;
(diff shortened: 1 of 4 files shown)


lib/core/src/send_swap.rs
```diff
@@ -0,0 +1,420 @@
...
+        )?;
+
+        info!(
+            "Successfully broadcast non-cooperative refund for Send Swap {}, tx: {}",
+            swap.id, refund_tx_id
+        );
+        Ok(refund_tx_id)
+    }
+
+    fn validate_state_transition(
```
dangeross, line 382: I notice the `validate_state_transition` function is copied (not moved) from the sdk module. I assume you will move it for the receive swap handler. I guess it's the best way to do it even though at the moment the validation is the same for both send and receive.

lib/core/src/sdk.rs
```diff
@@ -172,14 +168,31 @@ impl LiquidSdk {
         tokio::spawn(async move {
             let mut shutdown_receiver = cloned.shutdown_receiver.clone();
             let mut updates_stream = cloned.status_stream.subscribe_swap_updates();
+            let send_swap_state_handler = SendSwapStateHandler::new(
```
dangeross, line 171: Any reason the SendSwapStateHandler is not stored in the LiquidSdk struct?

dangeross (approved):
Looks good (with some comments)

lib/core/src/send_swap.rs
roeierez: Next on my plate is to extract out the receive swap handler. If this logic needs to be common I will refactor it out to a shared component.
I just not sure atm that the state transition validation should be the same for both. For example tonedout doesn't exist in receive.

lib/core/src/sdk.rs
roeierez: In my mind it is not a statefull struct it is just a stateless handler that can be instantiate on demand leaving the LiquodSdk struct simpler.
I assume saving it on the LiquidSdk struct will make sense if it is abstracted behind a trait but this is not the case and I don't think we need it atm.
Does it make sense to you as well?

dangeross: Yeah, I thought about that because it doesn't need the trait. I'm happy

PR #296, opened 2024-06-06
Extract out receive swap handling logic

roeierez:
Extract out receive swap handling to its own module.
@dangeross also noticed you were right about the need to put the handlers as part of the LiquidSDK state as we subscribe to events on them therefore we should use a single instance for handling.

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -202,7 +202,7 @@ impl LiquidSdk {
                                 error!("Failed to emit payment update: {e:?}");
                               }
                             }
-                            Err(e) => error!("Failed to receive send swap state change: {e:?}")
+                            Err(e) => error!("Failed to receive swap state change: {e:?}")
                         }
                       }
                     }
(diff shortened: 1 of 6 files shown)


lib/core/src/sdk.rs
```diff
@@ -168,39 +184,42 @@ impl LiquidSdk {
...
+                      if let Some(payment_id) = payment_id {
+                        match payment_id {
+                            Ok(payment_id) => {
+                              if let Err(e) = cloned.emit_payment_updated(Some(payment_id)).await {
+                                error!("Failed to emit payment update: {e:?}");
+                              }
                             }
-                          }
-                          Err(e) => error!("Failed to receive send swap state change: {e:?}")
+                            Err(e) => error!("Failed to receive send swap state change: {e:?}")
```
dangeross, line 205:
```suggestion
                            Err(e) => error!("Failed to receive swap state change: {e:?}")
```

dangeross (approved):
Looks good, just a minor nit

lib/core/src/sdk.rs
roeierez: Done

PR #298, opened 2024-06-06
Send Chain Swap

dangeross:
TODO:
- [x] Add table migration
- [x] Add persistence functions
- [x] Add models
- [x] Add `prepare_pay_onchain` and `pay_onchain` functions
- [x] Add `ChainSwapStateHandler`
  - [x] Lockup funds on Swap Created
  - [x] Wait for `TransactionServerMempool` / `TransactionServerConfirmed`, cooperate on remote claim
  - [x] Handle fail / refund
- [x] Add bindings interface

Tested:
- [x] Received payment is correct
- [x] LBTC-BTC claim cooperative flow
- [x] LBTC-BTC refund cooperative flow (tested triggering a refund, coop refund was not eligible, waiting for non coop locktime)

### DIFF
--- a/lib/bindings/src/breez_liquid_sdk.udl
+++ b/lib/bindings/src/breez_liquid_sdk.udl
@@ -3,6 +3,7 @@ enum LiquidSdkError {
     "AlreadyStarted",
     "Generic",
     "NotStarted",
+    "ServiceConnectivity",
 };
 
 [Error]
@@ -29,7 +30,8 @@ enum PaymentError {
 
 dictionary Config {
     string boltz_url;
-    string electrum_url;
+    string liquid_electrum_url;
+    string bitcoin_electrum_url;
     string working_dir;
     Network network;
     u64 payment_timeout_sec;
@@ -79,6 +81,20 @@ dictionary PrepareReceiveResponse {
 dictionary ReceivePaymentResponse {
     string id;
     string invoice;
+};
+
+dictionary PreparePayOnchainRequest {
+    u64 amount_sat;
+};
+
+dictionary PreparePayOnchainResponse {
+    u64 amount_sat;
+    u64 fees_sat;
+};
+
+dictionary PayOnchainRequest {
+    string address;
+    PreparePayOnchainResponse prepare_res;
 };
 
 dictionary BackupRequest {
@@ -203,6 +219,12 @@ interface BindingLiquidSdk {
 
     [Throws=PaymentError]
     ReceivePaymentResponse receive_payment(PrepareReceiveResponse req);
+
+    [Throws=PaymentError]
+    PreparePayOnchainResponse prepare_pay_onchain(PreparePayOnchainRequest req);
+
+    [Throws=PaymentError]
+    SendPaymentResponse pay_onchain(PayOnchainRequest req);
 
     [Throws=PaymentError]
     sequence<Payment> list_payments();
--- a/lib/bindings/src/lib.rs
+++ b/lib/bindings/src/lib.rs
@@ -111,6 +111,17 @@ impl BindingLiquidSdk {
         rt().block_on(self.sdk.receive_payment(&req))
     }
 
+    pub fn prepare_pay_onchain(
+        &self,
+        req: PreparePayOnchainRequest,
+    ) -> Result<PreparePayOnchainResponse, PaymentError> {
+        rt().block_on(self.sdk.prepare_pay_onchain(&req))
+    }
+
+    pub fn pay_onchain(&self, req: PayOnchainRequest) -> Result<SendPaymentResponse, PaymentError> {
+        rt().block_on(self.sdk.pay_onchain(&req))
+    }
+
     pub fn list_payments(&self) -> Result<Vec<Payment>, PaymentError> {
         rt().block_on(self.sdk.list_payments())
     }
--- a/lib/core/src/chain_swap.rs
+++ b/lib/core/src/chain_swap.rs
@@ -1,0 +1,568 @@
+use std::{str::FromStr, sync::Arc};
+
+use anyhow::{anyhow, Result};
+use boltz_client::swaps::boltzv2;
+use boltz_client::swaps::{boltz::ChainSwapStates, boltzv2::CreateChainResponse};
+use log::{debug, error, info, warn};
+use lwk_wollet::elements::Transaction;
+use lwk_wollet::ElectrumUrl;
+use tokio::sync::{broadcast, Mutex};
+
+use crate::chain::bitcoin::{BitcoinChainService, ElectrumClient};
+use crate::chain::ChainService;
+use crate::model::PaymentState::{Complete, Created, Failed, Pending, TimedOut};
+use crate::model::{ChainSwap, Config, Direction, PaymentTxData, PaymentType};
+use crate::swapper::Swapper;
+use crate::wallet::OnchainWallet;
+use crate::{error::PaymentError, model::PaymentState, persist::Persister};
+
+pub(crate) struct ChainSwapStateHandler {
+    onchain_wallet: Arc<dyn OnchainWallet>,
+    persister: Arc<Persister>,
+    swapper: Arc<dyn Swapper>,
+    liquid_chain_service: Arc<Mutex<dyn ChainService>>,
+    bitcoin_chain_service: Arc<Mutex<dyn BitcoinChainService>>,
+    subscription_notifier: broadcast::Sender<String>,
+}
+
+impl ChainSwapStateHandler {
+    pub(crate) fn new(
+        config: Config,
+        onchain_wallet: Arc<dyn OnchainWallet>,
+        persister: Arc<Persister>,
+        swapper: Arc<dyn Swapper>,
+        liquid_chain_service: Arc<Mutex<dyn ChainService>>,
+    ) -> Result<Self> {
+        let (subscription_notifier, _) = broadcast::channel::<String>(30);
+        let bitcoin_chain_service = Arc::new(Mutex::new(ElectrumClient::new(&ElectrumUrl::new(
+            &config.bitcoin_electrum_url,
+            true,
+            true,
+        ))?));
+        Ok(Self {
+            onchain_wallet,
+            persister,
+            swapper,
+            liquid_chain_service,
+            bitcoin_chain_service,
+            subscription_notifier,
+        })
+    }
+
+    pub(crate) fn subscribe_payment_updates(&self) -> broadcast::Receiver<String> {
+        self.subscription_notifier.subscribe()
+    }
+
+    /// Handles status updates from Boltz for Chain swaps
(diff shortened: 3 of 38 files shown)


lib/core/src/chain_swap.rs
```diff
@@ -0,0 +1,382 @@
...
+                                    None,
+                                    None,
+                                    None,
+                                    Some(&refund_tx_id),
+                                )
+                                .await?;
+                            }
+
+                            // If there is a lockup tx when receiving we need to refund to a sender address
+                            // TODO: Get the sender refund address
```
ok300, line 179: I see two ways to do this:
- the SDK asks for a refund address each time an onchain swap is started
- the SDK config contains a BTC xpub, which is set on init, from which we can derive new refund addresses when needed

The xpub way is probably best, @roeierez what do you think?

ok300: If we refund back to L-BTC, then it's simpler: `sdk.next_unused_address()` should do the trick.

dangeross: The refund is on the BTC lockup side, so it should be the sender getting the refund.

> If we refund back to L-BTC, then it's simpler: `sdk.next_unused_address()` should do the trick.

Yes, the L-BTC lockup side refund is covered like this

ok300: > it should be the sender getting the refund

Then looks like this should do the trick:

```rust
let sender_refund_address = swap.get_boltz_create_response()?.lockup_details.refund_address;
```

lib/core/src/chain_swap.rs
```diff
@@ -0,0 +1,382 @@
...
+        broadcast_fees_sat: u64,
+    ) -> Result<String, PaymentError> {
+        info!(
+            "Initiating non-cooperative refund for Chain Swap {}",
+            &swap.id
+        );
+
+        let current_height = match swap.payment_type {
+            PaymentType::Receive => self.bitcoin_chain_service.tip()?.height as u32,
+            PaymentType::Send => self.onchain_wallet.tip().await.height(),
```
roeierez, line 337: Why not using the liquid chain service here?
Also I am still confused about the receive type. Do we support Receiving via BTC address in this PR?

roeierez (changes requested):
Looks good.
My main confusion is about the Receive (BTC -> L-BTC).

lib/bindings/src/lib.rs
```diff
@@ -111,6 +111,20 @@ impl BindingLiquidSdk {
         rt().block_on(self.sdk.receive_payment(&req))
     }
 
+    pub fn prepare_send_onchain_payment(
```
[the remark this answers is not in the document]
dangeross: Sure, we can rename.

I have another PR which focuses on the receive side adding the interfaces and testing.

lib/core/src/chain_swap.rs
```diff
@@ -0,0 +1,382 @@
...
+                Ok(())
+            }
+
+            // Boltz announced the server lockup tx is in the mempool or has been confirmed.
+            // If it's a zero conf swap or confirmed, proceed to cooperative claim
+            ChainSwapStates::TransactionServerMempool
+            | ChainSwapStates::TransactionServerConfirmed => {
+                match swap.claim_tx_id {
+                    None => match (swap.accept_zero_conf, swap_state) {
+                        (true, _) | (_, ChainSwapStates::TransactionServerConfirmed) => {
```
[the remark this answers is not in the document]
dangeross: It is possible that boltz could accept a zero conf swap here, if we always want to wait for confirmation then we can remove it

lib/core/src/chain_swap.rs
dangeross: When receiving the lockup side is BTC, so we have to check the locktime relative to the BTC chain

dangeross: A lot of the receive side is covered in this PR, just not the interface. There are some open questions regarding handling of the receive (BTC-LBTC) refund. It would be a refund of the lockup to the BTC sender. Maybe we can have a call on Monday to step through it?

lib/core/src/chain_swap.rs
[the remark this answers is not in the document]
roeierez: Right, I missed that. Then indeed we need the mempool condition.

lib/core/src/chain_swap.rs
roeierez: OK, I see there is another PR for receive so I will ignore that in my review.

lib/core/src/chain_swap.rs
roeierez: Sure. My question was mainly why for the BTC chain we get the tip from the bitcoin chain service (electrum that is always up to date) and for liquid chain we get it from the onchain wallet (that may not be in sync)  and not the liquid_chain_service.

dangeross: Yeah right,  good point 👍

lib/core/src/chain_swap.rs
roeierez: @dangeross @ok300 I started to review the other PR and I see the intention is to handle refund here so here are my thoughts:
I think refunding to an external BTC address should be manual (user API) and not automatic from the following reasons:
1. It allow the user defines the fee at the time of refund
2. It allows the user bump the fees for the refund tx if needed.
3. It is consistent with the breez-sdk

So what do you think about adding the following endpionts:
1. list_refundables
2. refund

And ignore refundable states in the chain_swap.rs

roeierez (changes requested):

lib/core/src/sdk.rs
```diff
@@ -222,26 +231,22 @@ impl LiquidSdk {
...
+                            match cloned.persister.fetch_swap_by_id(id) {
+                                Ok(Swap::Send(_)) => match cloned.send_swap_state_handler.on_new_status(&update).await {
+                                    Ok(_) => info!("Successfully handled Send Swap {id} update"),
+                                    Err(e) => error!("Failed to handle Send Swap {id} update: {e}")
+                                },
+                                Ok(Swap::Receive(_)) => match cloned.receive_swap_state_handler.on_new_status(&update).await {
+                                    Ok(_) => info!("Successfully handled Receive Swap {id} update"),
+                                    Err(e) => error!("Failed to handle Receive Swap {id} update: {e}")
+                                },
+                                Ok(Swap::Chain(_)) => match cloned.chain_swap_state_handler.on_new_status(&update).await {
```
roeierez, line 244: @dangeross I think we might consider implementing two handlers (chain_send, chain_receive) instead of one handler that handles both.
The attempt to consolidate them into one handler perhaps reduce the code a bit but it makes it difficult to understand as some points in the code are trying to handle both use cases (such as claim) which I think lack some separation of concerns. Bottom line by reviewing the code one always needs to consider both cases in mind and given the fact that each case is complicated as it is it makes the review/maintaining more difficult than it could have been.
If you tried that and think otherwise I would be happy to hear some arguments.

lib/core/src/sdk.rs
```diff
@@ -285,6 +300,34 @@ impl LiquidSdk {
         });
     }
 
+    async fn check_chain_swap_expiration(&self, chain_swap: &ChainSwap) -> Result<()> {
+        if chain_swap.user_lockup_tx_id.is_some() && chain_swap.refund_tx_id.is_none() {
+            match chain_swap.payment_type {
+                PaymentType::Send => {
+                    let swap_script = chain_swap.get_lockup_swap_script()?.as_liquid_script()?;
+
+                    let current_height = self.onchain_wallet.tip().await.height();
```
roeierez, line 309: Better to use liquid_chain_service?

roeierez (changes requested):

lib/core/src/sdk.rs
dangeross: Makes sense from a maintainability point of view

lib/core/src/swapper/mod.rs
```diff
@@ -176,114 +424,172 @@ impl Swapper for BoltzSwapper {
...
+                self.refund_swap_cooperative(
+                    swap.id.clone(),
+                    swap_script.as_liquid_script()?,
+                    &refund_keypair,
+                    output_address,
+                    broadcast_fees_sat,
+                )
+            }
+            PaymentType::Receive => Err(PaymentError::Generic {
+                err: format!("Cannot refund a receive Chain Swap {}", swap.id),
```
roeierez, line 466: We actually do need to refund on the receive case. The user sends btc to the script address and if boltz doesn't claim them then we need to refund the user to an address he/she is passing.
I suggested this kind of refund will be in a different API but we still need that logic on the swapper.
Regardless I think worth considering call this method refund_send_chain_swap_cooperative and let it handle only the send case.

lib/bindings/src/breez_liquid_sdk.udl
```diff
@@ -81,6 +83,21 @@ dictionary ReceivePaymentResponse {
     string invoice;
 };
 
+dictionary PrepareOnchainRequest {
```
roeierez, line 86: Why do we need the payment_type field here? Is it so we will be able to use the same API for both swap in/ swap out? I see the rational but I think it will be too confusing for users.
Also from consistency with the sdk standpoint we probably need:
1. pay_onchain
2. receive_onchain

Unless you think this is something needs to be changed?

lib/core/src/persist/migrations.rs
```diff
@@ -35,5 +35,24 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
             amount_sat INTEGER NOT NULL,
             fees_sat INTEGER NOT NULL
         ) STRICT;",
+        "CREATE TABLE IF NOT EXISTS chain_swaps (
+            id TEXT NOT NULL PRIMARY KEY,
+            payment_type INTEGER NOT NULL,
```
roeierez, line 40: The term payment_type belongs to a higher level of abstraction IMO. It may also have more variants in the future that may not fit here.
Perhaps we can just use a different enum? Direction{ Outgoing, Incoming} or  Direction(Send, Receive} ?

roeierez (changes requested):

lib/core/src/chain_swap.rs
dangeross: This is only for the receive case where we need the BTC address right? Otherwise we handle it automatically in the send case

roeierez: Yes, exactly.

roeierez (approved):
Dropped one NIT but otherwise LGTM and ready to go in.

lib/bindings/src/breez_liquid_sdk.udl
```diff
@@ -81,6 +83,20 @@ dictionary ReceivePaymentResponse {
...
```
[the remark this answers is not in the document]
dangeross: I was using the same request/response structs for both prepare_pay_onchain() and prepare_receive_onchain(), but I'll separate them

ok300 (approved):

PR #299, opened 2024-06-07
Cleanup, fix typos

ok300:


dangeross (commented):
RN codegen needs running,  otherwise looks good

lib/core/src/swapper/boltz_status_stream.rs
```diff
@@ -140,7 +140,9 @@ impl SwapperStatusStream for BoltzStatusStream {
                                                 Err(e) => warn!("WS response is invalid SwapUpdate: {e:?}"),
                                             }
                                         },
-                                        Ok(msg) => info!("Unhandled msg: {msg:?}"),
+                                        Ok(Message::Ping(_)) => debug!("Received ping"),
```
dangeross, line 143: The nice thing to do here would be to send back a Pong

ok300: Yep, tungstenite does that automatically, here it only surfaces the received ping event.

dangeross: Cool 👍

dangeross (approved):

PR #301, opened 2024-06-07
Receive: Switch payment to pending state when lockup is in the mempool

### LINKED ISSUE #227: Show Receive as pending in `list-payments` even when we don't broadcast the claimtx immediately

hydra-yse:
Closes #227.


ok300 (approved):
I added some comments for clarity and to cover some edge cases in the payments query.

LGTM.

Not an ideal solution, but good enough for now. The problem is that apps cannot identify a Payment by the `tx_id` anymore, as for Receive Swaps it can alternate between `lockup_tx_id` and `claim_tx_id`. They also cannot use the `swap_id` as Payment ID, since for non-swap payments this is not set.

We can address the payment ID topic in a future release (#302).

ok300:
> The problem is that apps cannot identify a Payment by the `tx_id` anymore

This is only the case for Receive Swaps, so as workaround they can fallback on the `swap_id`, which in that case is set.

This way they can still identify and correlate payments when they receive events.

lib/core/src/model.rs
```diff
@@ -545,6 +547,9 @@ pub struct PaymentSwapData {
 #[derive(Debug, Clone, PartialEq, Serialize)]
 pub struct Payment {
     /// The tx ID of the onchain transaction
+    ///
+    /// For Receive Payments that are [PaymentState::Pending], this may be the lockup or the claim
+    /// tx ID. For Receive Payments that are [PaymentState::Complete], this is the claim tx ID.
```
roeierez, line 552: I don't think the lockup in case of receive should be the tx_id.
It seems to me that tx_id should only identify a transaction of the user's onchain wallet.

roeierez (changes requested):

lib/core/src/receive_swap.rs
```diff
@@ -143,7 +165,13 @@ impl ReceiveSwapStateHandler {
             is_confirmed: false,
         })?;
 
-        self.update_swap_info(swap_id, Pending, Some(&claim_tx_id))
+        // The lockup tx_id was used to identify the Receive Swap until the claim tx is available
+        // We now purge the temporary lockup tx from the tx data table
+        if let Some(lockup_tx_id) = &ongoing_receive_swap.lockup_tx_id {
+            self.persister.remove_temporary_tx(lockup_tx_id)?;
```
roeierez, line 171: Is there a reason we need to remove it?

lib/core/src/persist/mod.rs
```diff
@@ -194,18 +214,32 @@ impl Persister {
...
+                fees_sat: pending_receive_swap.payer_amount_sat
+                    - pending_receive_swap.receiver_amount_sat,
+                preimage: None,
+                refund_tx_id: None,
+                refund_tx_amount_sat: None,
+                payment_type: PaymentType::Receive,
+                status: pending_receive_swap.state,
+            })
+            .collect();
+        Ok([payments, unclaimed_payments].concat())
```
roeierez, line 243: The fact we are concatenating here will prevent us from exposing paging/filtering interface for the list_payments API.
I think we should come up with a way to get the pending receives as part of the `select_payment_query`
Perhaps we can use full outer join instead?

roeierez (changes requested):

lib/core/src/persist/mod.rs
roeierez: Since this flow is getting more complicated now I suggest to add some tests to the `get_payments` function with various payment types.

roeierez (changes requested):

lib/core/src/persist/mod.rs
```diff
@@ -118,28 +118,32 @@ impl Persister {
                 ss.state,
                 rtx.amount_sat
             FROM payment_tx_data AS ptx          -- Payment tx (each tx results in a Payment)
-            LEFT JOIN receive_swaps AS rs        -- Receive Swap data
+            FULL JOIN receive_swaps AS rs        -- Receive Swap data (by claim)
```
roeierez, line 121: @hydra-yse  Will this take also the receives swap that are in Created state?

ok300 (commented):
Looks good, just a few NITs.

roeierez (approved):
LGTM. @hydra-yse let's add the persistency tests in a new PR.

PR #303, opened 2024-06-09
Abstract ChainService

roeierez:
ChainService is abstracted.
There are still specific instantiations of electrum inside Swapper & Wallet but this is due to the explicit dependency of these trait implementations in Electrum. This is OK IMO as these implementations are already behind traits.

### DIFF
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -1,12 +1,11 @@
 use anyhow::{anyhow, Result};
-use boltz_client::network::electrum::ElectrumConfig;
 use boltz_client::network::Chain;
 use boltz_client::swaps::boltzv2::{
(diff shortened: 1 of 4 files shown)


dangeross:
I was thinking about this for the Chain Swap also as we need to validate the server lockup tx, construct the claim and broadcast it. How we could abstract the interface for chain services for liquid and bitcoin. I don't have any solid answers, maybe a shared trait and placeholder/associated types

roeierez:
> I was thinking about this for the Chain Swap also as we need to validate the server lockup tx, construct the claim and broadcast it. How we could abstract the interface for chain services for liquid and bitcoin. I don't have any solid answers, maybe a shared trait and placeholder/associated types

 Since The ChainService contains general access to the blockhain it shouldn't change from different chain types.
 I think for BTX we will need another implementation of the same trait. If we need more functionality IMO we will need to extend the trait.

dangeross (approved):
LGTM

PR #306, opened 2024-06-12
`get_info`: remove `with_scan` argument

ok300:
This PR removes the argument for `get_info`.


dangeross:
I think all the bindings tests need updating and the Flutter/React Native example apps also

dangeross (approved):
LGTM

PR #308, opened 2024-06-12
Integrate `sdk-common`

### LINKED ISSUE #275: Implement Lnurl pay

ok300:
- [x] `parse_invoice`
- [x] `parse`
- [x] LNURL-pay
  - [ ] Store success action in separate table, linked to `SendSwap` by `payment_hash`
  - [ ] Update min/max range for `LnUrlPayRequestData` to also reflect swap min/max
- [x] LNURL-withdraw
- [x] LNURL-auth
  - [ ] TBD: Do we directly want to access the seed? Or better export this whole method upstream to `lwk`?

Closes #275, #276 

Depends on https://github.com/breez/breez-sdk/pull/1012


lib/core/src/lib.rs
```diff
@@ -13,3 +13,160 @@ pub(crate) mod send_swap;
...
+
+pub use sdk_common::prelude::*;
+
+// === FRB mirroring
+//
+// This section contains frb "mirroring" structs and enums.
+// These are needed by the flutter bridge in order to use structs defined in an external crate.
+// See <https://cjycode.com/flutter_rust_bridge/v1/feature/lang_external.html#types-in-other-crates>
+
+use flutter_rust_bridge::frb;
```
roeierez, line 25: Can we move these to the binding.rs same as we did with the sdk?

lib/core/src/model.rs
```diff
@@ -68,53 +68,57 @@ impl Config {
 }
 
 #[derive(Debug, Copy, Clone, PartialEq, Serialize)]
-pub enum Network {
+pub enum LiquidSdkNetwork {
```
roeierez, line 71: IMO no need to add "sdk" as we already inside the sdk. LiquidNetwork seems fine.

lib/core/src/model.rs
```diff
@@ -767,6 +708,38 @@ impl From<SwapTree> for InternalSwapTree {
     }
 }
 
+/// LNURL-related wrappers
+pub mod lnurl {
```
roeierez, line 712: Why are these structures in a different module (lnurl)? Why not just inside the models?

roeierez (changes requested):
Looks good. Dropped some comments.

roeierez:
@ok300 for the open points:
1. Success action stored in a separate table seems correct to me
2. min/max range should IMO reflect only the LNURL, otherwise the user won't know the service original result
3. for deriving I think indeed better not to expose the seed directly but can be done in a separate PR.

lib/core/src/model.rs
ok300: Done, moved them to `models.rs`: cc7a206dabf548af5d2c516d76abcb42487dae0d

lib/core/Cargo.toml
```diff
@@ -25,8 +25,9 @@ lwk_signer = "0.5.1"
 # Switch back to published version once this PR is merged and included in release: https://github.com/Blockstream/lwk/pull/34 (ETA in v0.5.2)
 lwk_wollet = { git = "https://github.com/Blockstream/lwk", rev = "ffd793d0a1b1122c9bba7de23ccb73033eded98c" }
 #lwk_wollet = "0.5.1"
-rusqlite = { version = "0.31", features = ["backup", "bundled"] }
+rusqlite = { version = "0.29", features = ["backup", "bundled"] } # Same version as used in sdk-common
```
dangeross, line 28: Why does the sdk-common crate need rusqlite? I know its the FromSql/ToSql implementations, but I think it should the impartial and up to the application to implement these

ok300: In principle I agree, but Rust doesn't allow you to implement traits for structs derived in another crate.

To simplify `sdk-common` later on, we could place `rusqlite` behind a feature flag, so callers who don't need it don't have to download it.

lib/core/src/lib.rs
ok300: Done: db821d9b4ff20c00d4aef8ea51558d011c5490fd

lib/core/Cargo.toml
roeierez: The FromSQL ToSQL are helpers to read/write structs to the db using rusqlite.
We don't necessary have to use these helpers.
I think there are two points in the code where we use them:
1. sql_row_to_payment (for read)
2. insert_payment_external_info (for write)

I think it worth serializing manually to/from string there (serde_json::to_string, serde_json::from_str) and get rid of the rusqlite depenency.

roeierez (changes requested):
Other than one sugestion for getting rid of rusqlite, LGTM.

dangeross (commented):
Looks good. I think the example Flutter and RN apps need updating with the `Network` rename

lib/core/Cargo.toml
ok300: Done in https://github.com/breez/breez-sdk/pull/1019

roeierez (approved):
LGTM

PR #310, opened 2024-06-14
Receive Chain Swaps

dangeross:
TODO:
- [x] Add models
- [x] Add `prepare_receive_onchain` and `receive_onchain` functions
  - [x] Wait for `TransactionServerMempool` / `TransactionServerConfirmed`, cooperate on remote claim
  - [x] Handle fail / add `list_refundables`, `prepare_refund` and `refund` functions
- [x] Add bindings interface

Tested:
- [x] Received payment is correct (amount - boltz fee - claim fee - server fee)
- [x] BTC-LBTC claim cooperative flow
- [x] BTC-LBTC refund cooperative flow

### DIFF
--- a/lib/bindings/src/breez_liquid_sdk.udl
+++ b/lib/bindings/src/breez_liquid_sdk.udl
@@ -312,6 +312,24 @@ dictionary PayOnchainRequest {
     PreparePayOnchainResponse prepare_res;
 };
 
+dictionary PrepareReceiveOnchainRequest {
+    u64 amount_sat;
+};
+
+dictionary PrepareReceiveOnchainResponse {
+    u64 amount_sat;
+    u64 fees_sat;
+};
+
+dictionary ReceiveOnchainRequest {
+    PrepareReceiveOnchainResponse prepare_res;
+};
+
+dictionary ReceiveOnchainResponse {
+    string address;
+    string bip21;
+};
+
 dictionary BackupRequest {
     string? backup_path = null;
 };
@@ -345,6 +363,34 @@ enum PaymentState {
     "Complete",
     "Failed",
     "TimedOut",
+    "Refundable",
+};
+
+dictionary RefundableSwap {
+    string swap_address;
+    u32 timestamp;
+    u64 amount_sat;
+};
+
+dictionary PrepareRefundRequest {
+    string swap_address;
+    string refund_address;
+    u32 sat_per_vbyte;
+};
+
+dictionary PrepareRefundResponse {
+    u32 refund_tx_vsize;
+    u64 refund_tx_fee_sat;
+};
+
+dictionary RefundRequest {
+    string swap_address;
+    string refund_address;
+    u32 sat_per_vbyte;
+};
+
+dictionary RefundResponse {
+    string refund_tx_id;
 };
 
 [Enum]
@@ -416,7 +462,22 @@ interface BindingLiquidSdk {
     SendPaymentResponse pay_onchain(PayOnchainRequest req);
 
     [Throws=PaymentError]
+    PrepareReceiveOnchainResponse prepare_receive_onchain(PrepareReceiveOnchainRequest req);
+
+    [Throws=PaymentError]
+    ReceiveOnchainResponse receive_onchain(ReceiveOnchainRequest req);
+
+    [Throws=PaymentError]
     sequence<Payment> list_payments();
+
+    [Throws=LiquidSdkError]
+    sequence<RefundableSwap> list_refundables();
+
+    [Throws=LiquidSdkError]
+    PrepareRefundResponse prepare_refund(PrepareRefundRequest req);
+
+    [Throws=PaymentError]
+    RefundResponse refund(RefundRequest req);
 
     [Throws=LiquidSdkError]
     void sync();
--- a/lib/core/Cargo.toml
+++ b/lib/core/Cargo.toml
@@ -15,7 +15,7 @@ frb = ["dep:flutter_rust_bridge"]
 anyhow = { workspace = true }
 bip39 = { version = "2.0.0", features = ["serde"] }
 #boltz-client = { git = "https://github.com/SatoshiPortal/boltz-rust", rev = "a05731cc33030ada9ae14afcafe0cded22842ba6" }
-boltz-client = { git = "https://github.com/dangeross/boltz-rust", rev = "cb2cb02d44fb81cc8ce5d8000346f52cc26b3fc1" }
+boltz-client = { git = "https://github.com/dangeross/boltz-rust", branch = "savage-chain-swaps" }
 chrono = "0.4"
 env_logger = "0.11"
 flutter_rust_bridge = { version = "=2.0.0", features = ["chrono"], optional = true }
@@ -23,7 +23,8 @@ log = { workspace = true }
 lwk_common = "0.5.1"
 lwk_signer = "0.5.1"
 # Switch back to published version once this PR is merged and included in release: https://github.com/Blockstream/lwk/pull/34 (ETA in v0.5.2)
-lwk_wollet = { git = "https://github.com/Blockstream/lwk", rev = "ffd793d0a1b1122c9bba7de23ccb73033eded98c" }
+#lwk_wollet = { git = "https://github.com/Blockstream/lwk", rev = "ffd793d0a1b1122c9bba7de23ccb73033eded98c" }
+lwk_wollet = { git = "https://github.com/dangeross/lwk", branch = "savage-try-headers-subscribe" }
 #lwk_wollet = "0.5.1"
 rusqlite = { version = "0.31", features = ["backup", "bundled"] }
 rusqlite_migration = "1.0"
--- a/lib/core/src/chain_swap.rs
+++ b/lib/core/src/chain_swap.rs
@@ -5,13 +5,13 @@ use boltz_client::swaps::boltzv2;
 use boltz_client::swaps::{boltz::ChainSwapStates, boltzv2::CreateChainResponse};
 use log::{debug, error, info, warn};
 use lwk_wollet::elements::Transaction;
-use lwk_wollet::ElectrumUrl;
 use tokio::sync::{broadcast, Mutex};
 
-use crate::chain::bitcoin::{BitcoinChainService, ElectrumClient};
+use crate::chain::bitcoin::BitcoinChainService;
 use crate::chain::liquid::LiquidChainService;
-use crate::model::PaymentState::{Complete, Created, Failed, Pending, TimedOut};
-use crate::model::{ChainSwap, Config, Direction, PaymentTxData, PaymentType};
+use crate::error::{LiquidSdkError, LiquidSdkResult};
+use crate::model::PaymentState::{Complete, Created, Failed, Pending, Refundable, TimedOut};
+use crate::model::{ChainSwap, Direction, PaymentTxData, PaymentType};
 use crate::swapper::Swapper;
 use crate::wallet::OnchainWallet;
 use crate::{error::PaymentError, model::PaymentState, persist::Persister};
@@ -27,18 +27,13 @@ pub(crate) struct ChainSwapStateHandler {
 
 impl ChainSwapStateHandler {
     pub(crate) fn new(
-        config: Config,
         onchain_wallet: Arc<dyn OnchainWallet>,
         persister: Arc<Persister>,
         swapper: Arc<dyn Swapper>,
         liquid_chain_service: Arc<Mutex<dyn LiquidChainService>>,
+        bitcoin_chain_service: Arc<Mutex<dyn BitcoinChainService>>,
     ) -> Result<Self> {
         let (subscription_notifier, _) = broadcast::channel::<String>(30);
-        let bitcoin_chain_service = Arc::new(Mutex::new(ElectrumClient::new(&ElectrumUrl::new(
-            &config.bitcoin_electrum_url,
-            true,
-            true,
-        ))?));
         Ok(Self {
             onchain_wallet,
             persister,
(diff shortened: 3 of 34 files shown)


lib/core/src/chain_swap.rs
```diff
@@ -380,7 +380,15 @@ impl ChainSwapStateHandler {
 
     async fn claim(&self, chain_swap: &ChainSwap) -> Result<(), PaymentError> {
```
ok300, line 381: Maybe `claim_or_refund` is a better name? Seems to cover both cases, depending on direction.

lib/core/src/chain_swap.rs
```diff
@@ -380,7 +380,15 @@ impl ChainSwapStateHandler {
...
+        let refund_address = match chain_swap.direction {
+            Direction::Incoming => {
+                chain_swap
+                    .get_boltz_create_response()?
+                    .lockup_details
+                    .lockup_address
+            }
+            Direction::Outgoing => self.onchain_wallet.next_unused_address().await?.to_string(),
+        };
         let claim_tx_id = self.swapper.claim_chain_swap(chain_swap, refund_address)?;
```
ok300, line 392: `claim_or_refund_chain_swap`?

lib/core/Cargo.toml
```diff
@@ -15,7 +15,7 @@ frb = ["dep:flutter_rust_bridge"]
 anyhow = { workspace = true }
 bip39 = { version = "2.0.0", features = ["serde"] }
 #boltz-client = { git = "https://github.com/SatoshiPortal/boltz-rust", rev = "a05731cc33030ada9ae14afcafe0cded22842ba6" }
-boltz-client = { git = "https://github.com/dangeross/boltz-rust", rev = "cb2cb02d44fb81cc8ce5d8000346f52cc26b3fc1" }
+boltz-client = { git = "https://github.com/dangeross/boltz-rust", rev = "0b6874c12332206b9821734561d4a140f6c0cfbb" }
```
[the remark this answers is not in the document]
dangeross: Done

lib/core/src/sdk.rs
```diff
@@ -301,28 +305,40 @@ impl LiquidSdk {
...
+
+                    info!("Checking Chain Swap {} expiration: locktime_from_height = {locktime_from_height:?},  swap_script.locktime = {:?}", chain_swap.id, swap_script.locktime);
+                    if swap_script.locktime.is_implied_by(locktime_from_height) {
+                        let id: &String = &chain_swap.id;
+                        info!("Chain Swap {} user lockup tx was broadcast. Setting the swap to refundable.", id);
+                        self.chain_swap_state_handler
+                            .update_swap_info(id, Refundable, None, None, None, None)
+                            .await?;
+                    }
+                }
```
roeierez, line 329: This step for refunding incoming chain swaps is a bit tricky since there are some edge cases:
1. We can refund several times (even though there was an historical refund_id) for example when a user has sent funds to the swap address after it has completed .
2. Also we enter here only for pending swaps and we need to check also historical swaps (although we can do that not at the same frequencey if we want to optimize performance).

In general I think we should mark refundable if the swap address has positive balance (can lwk wallet give us this information?) and the swap has expired.

lib/core/src/sdk.rs
```diff
@@ -301,28 +305,40 @@ impl LiquidSdk {
...
+                        LockTime::from_height(current_height).map_err(|e| {
+                            PaymentError::Generic {
+                                err: format!(
+                                    "Error getting locktime from height {current_height:?}: {e}",
+                                ),
+                            }
+                        })?;
+
+                    info!("Checking Chain Swap {} expiration: locktime_from_height = {locktime_from_height:?},  swap_script.locktime = {:?}", chain_swap.id, swap_script.locktime);
+                    if swap_script.locktime.is_implied_by(locktime_from_height) {
```
roeierez, line 322: What are we checking here? Not sure is is correct as we need to check that the current tip minus the height of the lockup tx confirmation block is less than the timeout in blocks.

roeierez (changes requested):
Looks good. Did initial review, will need to do another one tomorrow.

lib/core/src/chain_swap.rs
```diff
@@ -270,7 +270,16 @@ impl ChainSwapStateHandler {
...
+        let refund_address = match chain_swap.payment_type {
+            PaymentType::Receive => {
+                // TODO: sender refund address?
+                chain_swap
+                    .get_boltz_create_response()?
+                    .lockup_details
+                    .lockup_address
+            }
+            PaymentType::Send => self.onchain_wallet.next_unused_address().await?.to_string(),
+        };
```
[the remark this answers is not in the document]
dangeross: Yes.

This particular code is getting a refund address to form a refund tx that is only used to provide a partial signature. This is then sent to boltz with the preimage and the claim tx to get a partial signature.

lib/core/src/sdk.rs
dangeross: > In general I think we should mark refundable if the swap address has positive balance (can lwk wallet give us this information?) and the swap has expired.

In this incoming case the swap address is a BTC address. During the swap we rely on the boltz status to tell us if there is a tx going to this address.

lib/core/src/sdk.rs
dangeross: This is doing essentially the same as `utils::is_locktime_expired` but for bitcoin.

```
   pub fn is_implied_by(&self, other: LockTime) -> bool {
        use LockTime::*;

        match (*self, other) {
            (Blocks(this), Blocks(other)) => this.value() <= other.value(),
            (Time(this), Time(other)) => this.value() <= other.value(),
            _ => false, // Not the same units.
        }
    }
```

lib/core/src/chain_swap.rs
ok300, line 1: Maybe `chain_swap.rs` could be split into `receive_onchain_swap.rs` / `send_onchain_swap.rs`?

lib/core/src/model.rs
```diff
@@ -643,6 +706,12 @@ pub enum PaymentState {
     /// This covers the case when the swap state is still Created and the swap fails to reach the
     /// Pending state in time. The TimedOut state indicates the lockup tx should never be broadcast.
     TimedOut = 4,
+
+    /// ## Incoming Chain Swaps
+    ///
+    /// This covers the case when the swap failed for any reason and there is a user lockup tx.
+    /// The swap in this case has to be manually refunded with a provided Bitcoin address
+    Refundable = 5,
```
ok300, line 714: There is `refund_outgoing_swap`, so the description of `Refundable` state should mention what it means in the "Outgoing Chain Swaps" case as well.

ok300 (commented):
Looks good. I added some questions and comments for clarity.

I still have to test refund for Receive Chain Swaps, but all other tests I ran went well.

lib/core/src/sdk.rs
ok300: Then maybe we can replace `utils::is_locktime_expired` with `is_implied_by` in the other places, too?

lib/core/src/chain_swap.rs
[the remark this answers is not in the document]
dangeross: Only the claim is happening here, the refund_address is only used to construct a tx using the lockup script and refund keypair to get a partial signature. Using this boltz then provides a partial signature of the claim tx

lib/core/src/chain_swap.rs
```diff
@@ -380,7 +380,15 @@ impl ChainSwapStateHandler {
 
     async fn claim(&self, chain_swap: &ChainSwap) -> Result<(), PaymentError> {
         debug!("Initiating claim for Chain Swap {}", &chain_swap.id);
-        let refund_address = self.onchain_wallet.next_unused_address().await?.to_string();
+        let refund_address = match chain_swap.direction {
+            Direction::Incoming => {
+                chain_swap
+                    .get_boltz_create_response()?
+                    .lockup_details
+                    .lockup_address
```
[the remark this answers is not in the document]
dangeross: See above, it's only used to get a partial signature of the lockup script/refund keypair

lib/core/src/sdk.rs
dangeross: Seems the elements LockTime doesn't have this function

lib/core/src/model.rs
dangeross: The outgoing chain swaps are automatically refunded back to the L-BTC wallet. If its cooperative it can be done straight away. I guess if it can only be done non-cooperatively, we can mark it Refundable until the locktime expires. Thoughts @roeierez?

lib/core/src/sdk.rs
roeierez: I see, the lock height is absolute, I missed that.

lib/core/src/model.rs
roeierez: I don't think so, we are using Refundable to mark these that the user needs to refund manually. If the process is automatic then marking these addresses as refundables would be confusing.
In most cases the refund in these cases is immediate (cooperative) and the payment is still pending untill refund confirms.
I think it makes sense to add a different status, perhaps PENDING_REFUND ? We can use these status immediately after we get a status from the swapper that the payment failed and we know we need to refund.

roeierez: @dangeross @ok300 This PR is already big enough so I suggest to leave current send_swaps as pending and not introducing another status (PENDING_REFUND). This means currently these will be left as PENDING untill they are refunded automatically. I seems reasonable for now IMO and we can discuss that later.

lib/bindings/src/breez_liquid_sdk.udl
```diff
@@ -345,6 +363,34 @@ enum PaymentState {
     "Complete",
     "Failed",
     "TimedOut",
+    "Refundable",
+};
+
+dictionary SwapRefundInfo {
```
roeierez, line 369: NIT `RefundableSwap` ?

lib/bindings/src/breez_liquid_sdk.udl
```diff
@@ -345,6 +363,34 @@ enum PaymentState {
...
+    string id;
```
roeierez, line 370: Perhaps the swap btc address instead of id? I think it is more intuitive for the user.

lib/core/src/chain_swap.rs
```diff
@@ -408,58 +416,132 @@ impl ChainSwapStateHandler {
...
+        sat_per_vbyte: u32,
+    ) -> LiquidSdkResult<(u32, u64)> {
+        let swap =
+            self.persister
+                .fetch_chain_swap_by_id(swap_id)?
+                .ok_or(LiquidSdkError::Generic {
+                    err: format!("Swap {} not found", swap_id),
+                })?;
+        match swap.refund_tx_id {
+            Some(refund_tx_id) => Err(LiquidSdkError::Generic {
```
roeierez, line 432: We should allow the user refund multiple times. It is useful for example to bump fees.

lib/core/src/chain_swap.rs
```diff
@@ -408,58 +416,132 @@ impl ChainSwapStateHandler {
...
+        sat_per_vbyte: u32,
+    ) -> Result<String, PaymentError> {
+        let swap =
+            self.persister
+                .fetch_chain_swap_by_id(swap_id)?
+                .ok_or(PaymentError::Generic {
+                    err: format!("Swap {} not found", swap_id),
+                })?;
+        match swap.refund_tx_id {
+            Some(refund_tx_id) => Err(PaymentError::Generic {
```
roeierez, line 458: Same here, multiple refunds.

lib/core/src/chain_swap.rs
```diff
@@ -408,58 +416,132 @@ impl ChainSwapStateHandler {
...
+                err: format!(
+                    "Refund tx for Chain Swap {} was already broadcast: txid {refund_tx_id}",
+                    swap.id
+                ),
+            }),
+            None => {
+                let output_address = self.onchain_wallet.next_unused_address().await?.to_string();
+                let (_, broadcast_fees_sat) =
+                    self.swapper
+                        .prepare_chain_swap_refund(swap, &output_address, 0.1)?;
```
roeierez, line 516: Can we use the lowball fee rate here?

roeierez (changes requested):
looks very good.
I dropped some comments and I think we are missing the verification of the server lockup (finding on the blockchain and verify the amounts, rbf, etc...) as we do with the liquid swaps. I think we can do another PR for that.

lib/core/src/chain_swap.rs
```diff
@@ -139,14 +134,19 @@ impl ChainSwapStateHandler {
...
+                                self.update_swap_info(id, Refundable, None, None, None, None)
+                                    .await?;
+                            }
+                            (None, ChainSwapStates::TransactionLockupFailed) => {
+                                info!("Chain Swap {id} user lockup tx was broadcast but lockup has failed. Setting the swap to refundable.");
+                                self.update_swap_info(id, Refundable, None, None, None, None)
+                                    .await?;
+                            }
+                            (None, _) => {
+                                info!("Chain Swap {id} user lockup tx was never broadcast. Resolving payment as failed.");
```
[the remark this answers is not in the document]
dangeross: We handle the SwapExpired status, but I didn't test this particular case. I'll test that

lib/core/src/chain_swap.rs
dangeross: I have a follow up PR #322 which removes this check and monitors the lockup addresses

lib/core/src/chain_swap.rs
dangeross: True, I'll update

lib/core/src/model.rs
ok300: Sounds good, then I agree we can leave this as it is, namely `Refundable` only used for receive onchain swaps.

lib/core/src/chain_swap.rs
ok300: Also handled in #322, as the one above.

lib/core/src/chain_swap.rs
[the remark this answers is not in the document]
ok300: All clear now. I added #328 to help bypass the `refund_address` topic entirely.

lib/bindings/src/breez_liquid_sdk.udl
ok300: IMO the `id` is useful to associate this with the `Payment` from `list_payments`.

So I would keep `id`, maybe for clarity also rename it to `swap_id`, as that's the `Payment` field it would match.

roeierez: @ok300 but we don't have id on Payment. We do have swap_id though.

roeierez (approved):
LGTM

ok300 (approved):
Only open NIT is https://github.com/breez/breez-liquid-sdk/pull/310#discussion_r1651021886

Otherwise LGTM.

lib/bindings/src/breez_liquid_sdk.udl
ok300: Yes, the suggestion was to add `swap_id` to the `RefundableSwap` / `SwapRefundInfo`, in addition to the `swap_address`.

Without it, people cannot cross-reference the refundable info with the payment info from `list_payments`.

PR #311, opened 2024-06-15
Fix (re)connect swap subscription

### LINKED ISSUE #173: WS update stream: make loop resilient against network disconnects, continue listening on re-connect
If we disconnect the network, `socket.read()` never returns anymore. This remains the case even after re-connection.

This means the update stream loop remains stuck, unable to process further subscriptions or status updates.

This happens even for a very short disconnect (1-2s).

dangeross:
This PR fixes a issue where swaps are not being subscribed to on (re)connect. 

In `on_stream_reconnect()` we are adding swaps to track via `track_swap_id()` which uses the `subscription_notifier` to add swap subscriptions. But because the `on_stream_reconnect()` is called before we subscribe to the `subscription_notifier` broadcasts, the initial subscriptions are lost as there is no receiver.

Fixes #173

### DIFF
--- a/lib/core/src/swapper/boltz_status_stream.rs
+++ b/lib/core/src/swapper/boltz_status_stream.rs
@@ -82,12 +82,12 @@ impl SwapperStatusStream for BoltzStatusStream {
                 debug!("Start of ws stream loop");
                 match self.connect().await {
                     Ok(mut ws_stream) => {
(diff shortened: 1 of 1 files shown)


roeierez (approved):
LGTM

ok300 (approved):

PR #312, opened 2024-06-18
Add persistence and swap state tests

hydra-yse:


lib/core/src/persist/mod.rs
```diff
@@ -224,3 +224,87 @@ impl Persister {
...
+
+    pub(crate) struct TempPersister {
+        pub(crate) persister: Persister,
+        #[allow(dead_code)]
+        pub(crate) temp_dir: TempDir,
+    }
+
+    pub(crate) fn new_persister() -> Result<TempPersister> {
+        let temp_dir = TempDir::new("liquid-sdk")?;
+        let persister = Persister::new(
```
roeierez, line 244: Can we use in memory persister for testing?

lib/core/src/persist/mod.rs
```diff
@@ -224,3 +224,87 @@ impl Persister {
...
+
+#[cfg(test)]
+pub(crate) mod test_utils {
+    use anyhow::{anyhow, Result};
+    use bip39::rand::{self, distributions::Alphanumeric, Rng};
+    use tempdir::TempDir;
+
+    use super::{Network, PaymentTxData, PaymentType, Persister};
+
+    pub(crate) struct TempPersister {
```
roeierez, line 236: Why do we need this TempPersister wrapper? Can't we use the Persister itself?

lib/core/src/send_swap.rs
```diff
@@ -431,3 +434,111 @@ impl SendSwapStateHandler {
...
+            (TimedOut, Pending),
+            (Complete, Complete),
+            (Failed, Complete),
+            (TimedOut, Complete),
+            (Complete, TimedOut),
+            (Failed, TimedOut),
+            (Complete, Failed),
+        ];
+
+        for (first_state, second_state) in invalid_combinations.iter() {
```
roeierez, line 532: This is very nice, it would be great also to test the `on_new_status` function of both send_swap and receive_swap

lib/core/src/receive_swap.rs
```diff
@@ -260,12 +260,116 @@ impl ReceiveSwapStateHandler {
...
 
+            (Complete, Failed) => Err(PaymentError::Generic {
+                err: format!("Cannot transition from {from_state:?} to Failed state"),
+            }),
             (_, Failed) => Ok(()),
         }
     }
 }
+
+pub(crate) mod test_utils {
```
roeierez, line 276: As far as I see the test_utils contains mocks for the various services.
What do you think about centralizing these on one file?

roeierez (commented):
@hydra-yse looks very good!
dropped some comments and looking forward to merge it.
We also need the CI to run the tests.

lib/core/src/persist/mod.rs
hydra-yse: Unfortunately I don't think this is possible, since we can't override the Persister's `get_connection` method (which is used internally by all other methods).

lib/core/src/persist/mod.rs
hydra-yse: The TempPersister exists because we create a `TempDir`, which gets deleted automatically once the tests are over. We need to store this object in memory until we are done with the test, then the directory will be deleted automatically once it goes out of scope.

lib/core/src/receive_swap.rs
hydra-yse: Agreed :+1:

lib/core/src/persist/mod.rs
roeierez: Why not? We can save a flag whether the persister is in memory or not (inside the persister) and change the get_connection accordingly. Or you see an issue with that?

lib/core/src/persist/mod.rs
roeierez: It is another good argument to use in-memory persister.

lib/core/src/persist/mod.rs
hydra-yse: I see, didn't think about making it a struct flag, rather an argument to `get_connection`. Will do!

lib/core/src/receive_swap.rs
```diff
@@ -260,12 +260,87 @@ impl ReceiveSwapStateHandler {
...
+            storage.insert_receive_swap(&receive_swap)?;
+
+            assert!(receive_swap_state_handler
+                .update_swap_info(&receive_swap.id, *second_state, None, None)
+                .await
+                .is_ok());
+        }
+
+        // Test invalid combinations of states
+        let invalid_combinations = [
```
ok300, line 318: Why not threat the invalid combinations as `all_possible_combinations - valid_combinations`?

E.g. I'd generate all possible combinations, then iterate:
- if it's in the valid list, perform the test expected for valid ones
- if not, perform the test expected for invalid ones

lib/core/src/send_swap.rs
```diff
@@ -431,3 +434,75 @@ impl SendSwapStateHandler {
...
+            storage.insert_send_swap(&send_swap)?;
+
+            assert!(send_swap_state_handler
+                .update_swap_info(&send_swap.id, *second_state, None, None, None)
+                .await
+                .is_ok());
+        }
+
+        // Test invalid combinations of states
+        let invalid_combinations = [
```
ok300, line 480: Same here.

lib/core/src/send_swap.rs
hydra-yse: Will be addressed in a separate PR, see #313.

ok300 (approved):

roeierez (approved):
LGTM

PR #315, opened 2024-06-19
Workaround subscriptions not persisting with reconnects

### LINKED ISSUE #314: Chain Service de-syncs from tip
The electrum client keeps sync with the latest headers by subscribing to header events. The lwk_wollet uses this to get the tip when needed. But if the underlying stream is not used regularly the stream connection is dropped by the remote. Because of the way lwk_wollet is initialised it only subscribes to header events once, if the connection is dropped the tip de-syncs as no new header events come in (not being re-subscribed on reconnect)

https://github.com/bitcoindevkit/rust-electrum-client/issues/124

dangeross:
This PR forks the upstream lwk_wollet to add the same logic as in the bitcoin module. If no header is popped then try to subscribe to headers in case the connection was dropped. It might be that the client has reconnected and subscriptions don't persist across connections. Calling `client.block_headers_pop_raw()` doesn't reestablish a connection. Calling `client.ping()` won't help because the successful retry (when reestablishing a connection) will prevent us knowing about the reconnect and not able to re-subscribe to header notifications.

Fixes #314

### DIFF
--- a/lib/core/src/chain/bitcoin.rs
+++ b/lib/core/src/chain/bitcoin.rs
@@ -9,7 +9,7 @@ use lwk_wollet::{
         consensus::{deserialize, serialize},
         BlockHash, Script, Transaction, Txid,
     },
-    ElectrumUrl, History,
+    ElectrumOptions, ElectrumUrl, Error, History,
 };
 
 type Height = u32;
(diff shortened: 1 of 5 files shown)


lib/core/src/chain/bitcoin.rs
```diff
@@ -60,9 +65,21 @@ impl BitcoinChainService for ElectrumClient {
...
 
-        if let Some(popped_header) = maybe_popped_header {
-            let tip: HeaderNotification = popped_header.try_into()?;
-            self.tip = tip;
+        match maybe_popped_header {
+            Some(popped_header) => {
+                let tip: HeaderNotification = popped_header.try_into()?;
+                self.tip = tip;
+            }
+            None => {
```
roeierez, line 73: Does this mean it will resubscribe anytime we will call tip and there isn't any new block? If this is so then I am concern with unnecessary network overhead whenever we call tip.

dangeross: Understandable. I can dig into https://github.com/bitcoindevkit/rust-electrum-client to see if it's fixable there. It seems `block_headers_pop_raw()` does not check the stream state, so jit ust tries to pop the header from an array. Currently there needs to be some call to network see if the stream is closed and then re-establish the connection (then re-subscribe to the header notifications)

roeierez (approved):

PR #320, opened 2024-06-20
Hybrid chain service

roeierez:
This PR reimplement the liquid chain service as a hybrid service. Still using electrum apart from the broadcast and get_script_history which uses the lowball node.
Also added verification for the swapper reported tx id and content before moving to the claim.

### DIFF
--- a/lib/core/src/chain/liquid.rs
+++ b/lib/core/src/chain/liquid.rs
@@ -85,7 +85,7 @@ impl LiquidChainService for HybridLiquidChainService {
             .to_hex();
         let url = format!("{}/scripthash/{}/txs", LIQUID_ESPLORA_URL, script_hash);
         // TODO must handle paging -> https://github.com/blockstream/esplora/blob/master/API.md#addresses
-        let response = get_with_retry(&url, 2).await?;
+        let response = get_with_retry(&url, 3).await?;
         let json: Vec<EsploraTx> = response.json().await?;
 
         let history: Vec<History> = json.into_iter().map(Into::into).collect();
--- a/lib/core/src/chain/liquid.rs
+++ b/lib/core/src/chain/liquid.rs
@@ -1,0 +1,128 @@
+use anyhow::{anyhow, Result};
+use async_trait::async_trait;
+use boltz_client::ToHex;
+use log::info;
+use lwk_wollet::{
+    elements::{pset::serialize::Serialize, BlockHash, Script, Transaction, Txid},
+    hashes::{sha256, Hash},
+    BlockchainBackend, ElectrumClient, ElectrumUrl, History,
+};
+use reqwest::Response;
+use serde::Deserialize;
+use std::str::FromStr;
+
+use crate::model::Config;
+
+const LIQUID_ESPLORA_URL: &str = "https://lq1.breez.technology/liquid/api";
+
+#[async_trait]
+pub trait LiquidChainService: Send + Sync {
+    /// Get the blockchain latest block
+    async fn tip(&mut self) -> Result<u32>;
+
+    /// Broadcast a transaction
+    async fn broadcast(&self, tx: &Transaction, swap_id: Option<&str>) -> Result<Txid>;
+
+    /// Get a list of transactions
+    async fn get_transactions(&self, txids: &[Txid]) -> Result<Vec<Transaction>>;
+
+    /// Get the transactions involved in a list of scripts including lowball
+    async fn get_script_history(&self, scripts: &Script) -> Result<Vec<History>>;
+}
+
+#[derive(Deserialize)]
+struct EsploraTx {
+    txid: Txid,
+    status: Status,
+}
+
+// TODO some of this fields may be Option in unconfirmed
+
+#[derive(Deserialize)]
+struct Status {
+    block_height: Option<i32>,
+    block_hash: Option<BlockHash>,
+}
+
+pub(crate) struct HybridLiquidChainService {
+    electrum_client: ElectrumClient,
+}
+
+impl HybridLiquidChainService {
+    pub(crate) fn new(config: Config) -> Result<Self> {
+        let electrum_client =
+            ElectrumClient::new(&ElectrumUrl::new(&config.liquid_electrum_url, true, true))?;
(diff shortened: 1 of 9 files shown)


lib/core/src/receive_swap.rs
```diff
@@ -270,6 +301,78 @@ impl ReceiveSwapStateHandler {
             (_, Failed) => Ok(()),
         }
     }
+
+    async fn verify_lockup_tx(
```
ok300, line 305: Does this only check if the lockup was broadcasted, or if it's also confirmed?

It's used in both cases in this PR, when handling:
- `RevSwapStates::TransactionMempool`
- `RevSwapStates::TransactionConfirmed`

Maybe it could return `Result<bool>` to cover both cases, based on `lockup_tx_history.is_none()` / `.is_some()`?

lib/core/src/receive_swap.rs
```diff
@@ -270,6 +301,78 @@ impl ReceiveSwapStateHandler {
...
+        receive_swap: &ReceiveSwap,
+        swap_update_tx: &SwapUpdateTxDetails,
+    ) -> Result<()> {
+        // looking for lockup script history to verify lockup was broadcasted
+        let script_history = self.lockup_script_history(receive_swap).await?;
+        let lockup_tx_history = script_history
+            .iter()
+            .find(|h| h.txid.to_hex().eq(&swap_update_tx.id));
+
+        if lockup_tx_history.is_none() {
```
ok300, line 316: Might be better to replace this with a
```rust
match lockup_tx_history {
    None => {

    }
    Some(history) => {

    }
}
```

to make various cases simpler to handle, and to get rid of the `unwrap` below.

ok300 (commented):
Looks good, found just a few NITs.

lib/core/src/receive_swap.rs
dangeross: The `lockup_tx_history` can be confirmed and unconfirmed. `lockup_tx_history.height > 0` will be a confirmed tx.

lib/core/src/chain/liquid.rs
```diff
@@ -0,0 +1,128 @@
...
+    async fn get_script_history(&self, scripts: &Script) -> Result<Vec<History>>;
+}
+
+#[derive(Deserialize)]
+struct EsploraTx {
+    txid: Txid,
+    status: Status,
+}
+
+// TODO some of this fields may be Option in unconfirmed
```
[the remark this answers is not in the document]
roeierez: Removed

lib/core/src/chain/liquid.rs
```diff
@@ -0,0 +1,128 @@
...
+
+#[async_trait]
+impl LiquidChainService for HybridLiquidChainService {
+    async fn tip(&mut self) -> Result<u32> {
+        Ok(self.electrum_client.tip()?.height)
+    }
+
+    async fn broadcast(&self, tx: &Transaction, swap_id: Option<&str>) -> Result<Txid> {
+        let tx_bytes = tx.serialize();
+        info!("tx: {}", tx_bytes.to_hex());
```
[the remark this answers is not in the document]
roeierez: Updated

lib/core/src/model.rs
```diff
@@ -18,6 +18,8 @@ use crate::receive_swap::{
 };
 use crate::utils;
 
+pub const LOWBALL_FEE_RATE: f32 = 0.01;
```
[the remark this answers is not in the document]
roeierez: Changed

lib/core/src/receive_swap.rs
roeierez: I added verification for the confirmed case to actually verify the tx is confirmed as the swapper reports.

lib/core/src/receive_swap.rs
roeierez: fixed

lib/core/src/chain/liquid.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    async fn get_script_history(&self, script: &Script) -> Result<Vec<History>> {
+        let script = lwk_wollet::elements::bitcoin::Script::from_bytes(script.as_bytes());
+        let script_hash = sha256::Hash::hash(script.as_bytes())
+            .to_byte_array()
+            .to_hex();
+        let url = format!("{}/scripthash/{}/txs", LIQUID_ESPLORA_URL, script_hash);
+        // TODO must handle paging -> https://github.com/blockstream/esplora/blob/master/API.md#addresses
+        let response = get_with_retry(&url, 2).await?;
```
dangeross, line 88: Is 2 attempts enough here? It could potentially fail the swap right if this fails when checking in TransactionConfirned

roeierez: The swap won't fail, the status stays as it is (pending) and next time we will get the status from boltz we will try again.
Regardless, I agree we can retry more here. I am thinking to change to 3, what do you think?

dangeross: True, on the next reconnect. Sounds good

dangeross (approved):
LGTM

lib/core/src/receive_swap.rs
```diff
@@ -270,6 +307,87 @@ impl ReceiveSwapStateHandler {
...
+
+        match lockup_tx_history {
+            Some(history) => {
+                info!("swapper lockup found, verifying transaction content...");
+
+                let lockup_tx = utils::deserialize_tx_hex(&swap_update_tx.hex)?;
+                if !lockup_tx
+                    .txid()
+                    .to_hex()
+                    .eq(&lockup_tx_history.unwrap().txid.to_hex())
```
ok300, line 331:
```suggestion
                    .eq(&history.txid.to_hex())
```

We already have `history`, so no need to `unwrap()`.

ok300 (commented):
One final NIT, otherwise LGTM.

lib/core/src/receive_swap.rs
roeierez: fixed

ok300 (approved):

PR #322, opened 2024-06-21
Monitor chain swap addresses

### LINKED ISSUE #324: Add PENDING_REFUND state
Add a state for when a swap fails and was either automatically triggered for a refund or manually initiated by the user.

dangeross:
This PR adds monitoring of the lockup addresses of incoming chain swaps, to check the state of any unspent funds for the swap lockup address after the swap is finished. It will periodically:
- For every swap with a pending refund tx, check if the lockup address has any unspent outputs. If there are non, sets the swap to Failed/Complete.
- For every swap that is expired, check if the lockup address has any confirmed unspent outputs for a month (4320 blocks). If there are some, sets the swap to Refundable.

In addition the `prepare_refund` / `refund` interface is updated to allow multiple refunds of a swap (after swap the locktime).

The RefundPending state is added to help differentiate between a pending succeeding swap and a failed swap being refunded.

Fixes #324

### DIFF
--- a/lib/core/src/chain/bitcoin.rs
+++ b/lib/core/src/chain/bitcoin.rs
@@ -1,10 +1,9 @@
 use std::collections::HashMap;
 
 use anyhow::Result;
-use electrum_client::{Client, ElectrumApi, HeaderNotification};
+use electrum_client::{Client, ElectrumApi, HeaderNotification, ListUnspentRes};
 use lwk_wollet::{
     bitcoin::{
-        self,
         block::Header,
         consensus::{deserialize, serialize},
         BlockHash, Script, Transaction, Txid,
@@ -13,6 +12,28 @@ use lwk_wollet::{
 };
 
 type Height = u32;
+
+#[allow(dead_code)]
+pub struct Unspent {
+    /// Confirmation height of the transaction that created this output.
+    pub height: usize,
+    /// Txid of the transaction
+    pub tx_hash: Txid,
+    /// Index of the output in the transaction.
+    pub tx_pos: usize,
+    /// Value of the output.
+    pub value: u64,
+}
+impl From<ListUnspentRes> for Unspent {
+    fn from(value: ListUnspentRes) -> Self {
+        Unspent {
+            height: value.height,
+            tx_hash: value.tx_hash,
+            tx_pos: value.tx_pos,
+            value: value.value,
+        }
+    }
+}
 
 /// Trait implemented by types that can fetch data from a blockchain data source.
 #[allow(dead_code)]
@@ -37,6 +58,9 @@ pub trait BitcoinChainService: Send + Sync {
 
     /// Get the transactions involved in a list of scripts
     fn get_scripts_history(&self, scripts: &[&Script]) -> Result<Vec<Vec<History>>>;
+
+    /// Get a list of unspend outputs
+    fn script_list_unspent(&self, script: &Script) -> Result<Vec<Unspent>>;
 }
 
 pub(crate) struct ElectrumClient {
@@ -91,13 +115,8 @@ impl BitcoinChainService for ElectrumClient {
     }
 
     fn get_transactions(&self, txids: &[Txid]) -> Result<Vec<Transaction>> {
-        let txids: Vec<bitcoin::Txid> = txids
-            .iter()
-            .map(|t| bitcoin::Txid::from_raw_hash(t.to_raw_hash()))
-            .collect();
-
         let mut result = vec![];
-        for tx in self.client.batch_transaction_get_raw(&txids)? {
+        for tx in self.client.batch_transaction_get_raw(txids)? {
             let tx: Transaction = deserialize(&tx)?;
             result.push(tx);
         }
@@ -118,16 +137,20 @@ impl BitcoinChainService for ElectrumClient {
     }
 
     fn get_scripts_history(&self, scripts: &[&Script]) -> Result<Vec<Vec<History>>> {
-        let scripts: Vec<&bitcoin::Script> = scripts
-            .iter()
-            .map(|t| bitcoin::Script::from_bytes(t.as_bytes()))
-            .collect();
-
         Ok(self
             .client
-            .batch_script_get_history(&scripts)?
+            .batch_script_get_history(scripts)?
             .into_iter()
             .map(|e| e.into_iter().map(Into::into).collect())
+            .collect())
+    }
+
+    fn script_list_unspent(&self, script: &Script) -> Result<Vec<Unspent>> {
+        Ok(self
+            .client
+            .script_list_unspent(script)?
+            .into_iter()
+            .map(Into::into)
             .collect())
     }
 }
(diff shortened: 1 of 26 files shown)


lib/core/src/chain/bitcoin.rs
```diff
@@ -101,16 +120,20 @@ impl BitcoinChainService for ElectrumClient {
...
             .into_iter()
             .map(|e| e.into_iter().map(Into::into).collect())
             .collect())
     }
+
+    fn script_list_unspent(&self, script: &Script) -> Result<Vec<Unspent>> {
+        Ok(self
+            .client
+            .script_list_unspent(script)?
+            .into_iter()
```
ok300, line 135:
```suggestion
            .into_iter()
            .filter(|utxo| script.dust_value().lt(&bitcoin::amount::Amount::from_sat(utxo.value)))
```

Outputs below the dust limit should be ignored when deciding if a swap is refundable or not.

roeierez: Not sure about this @ok300. Also there can be several utxos which their aggregated amount is above dust.

ok300: Agree. As discussed, we can leave this out.

ok300 (approved):

lib/core/src/sdk.rs
```diff
@@ -286,22 +289,123 @@ impl LiquidSdk {
...
+                    _ = shutdown_receiver.changed() => {
+                        info!("Received shutdown signal, exiting chain swap transactions loop");
+                        return;
+                    }
+                }
+            }
+        });
+    }
+
+    async fn check_chain_swap_refund_transactions(&self) -> Result<()> {
```
roeierez, line 334: I wonder whether we really need this function/tracking.
If the goal is to set the final state of the receive chain swap (Failed/Complete) then I have the following in my mind:
1. The Complete status is set after our claim (liquid) is confirmed and the place we do for lightning swaps is here: https://github.com/breez/breez-liquid-sdk/blob/main/lib/core/src/sdk.rs#L1193 perhaps it is also the place to verify the claim for chain swaps?
2. The Failed status indeed needs to be set when the refund tx is confirmed so perhaps we can wire it to the `check_expired_chain_swap_lockup_transactions` that already checks the expired swaps? maybe set to Failed if unspent_sat=0?

dangeross:
Waiting on #310 to merge

roeierez (changes requested):

lib/core/src/sdk.rs
dangeross: There are two functions for the tracking in `check_chain_swap_refund_transactions()`:
1. In the case of a refund during the swap, mark the swap Failed once the lockup is spent. This can be cooperative refunds also, not just expired/non-cooperative. This is only for incoming chain swaps as the outgoing refund is [tracked here](https://github.com/breez/breez-liquid-sdk/blob/a9d75a06727599f813c4dd029150feb3ee03c966/lib/core/src/sdk.rs#L1325-L1331).
2. In the case extra funds sent to the lookup address (outside of the swap flow) and refunds are manually triggered (set back to Pending), mark successful/unsuccessful swap back to Complete/Failed once the funds are spent.

> 1. The Complete status is set after our claim (liquid) is confirmed and the place we do for lightning swaps is here: https://github.com/breez/breez-liquid-sdk/blob/main/lib/core/src/sdk.rs#L1193 perhaps it is also the place to verify the claim for chain swaps?

We could instead move the confirmation (set to Complete) of the incoming chain swap liquid claim here. We also don't monitor the confirmation of the outgoing chain swap bitcoin claim. Maybe we should?

> 2. The Failed status indeed needs to be set when the refund tx is confirmed so perhaps we can wire it to the `check_expired_chain_swap_lockup_transactions` that already checks the expired swaps? maybe set to Failed if unspent_sat=0?

We can but in this case it would only be the non-cooperative / expired refunds.

I main reason for the separation into two functions is to make sure refunds are always handled if the refund is pending.

lib/core/src/chain_swap.rs
```diff
@@ -62,6 +90,87 @@ impl ChainSwapStateHandler {
...
+            .sum())
+    }
+
+    async fn rescan_chain_swap(&self, swap: &ChainSwap, current_height: u32) -> Result<()> {
+        let monitoring_block_height =
+            swap.timeout_block_height + CHAIN_SWAP_MONTIORING_PERIOD_BITCOIN_BLOCKS;
+        let is_swap_expired = current_height > swap.timeout_block_height;
+        let is_monitoring_expired = current_height > monitoring_block_height;
+
+        match swap.state {
```
roeierez, line 137: Just as an attempt to simplify the match statement here.
Can we replace the match with an if of one condition that checks if the swap  is: 
`!is_monitoring_expired && expired || swap.state == Refundable`

If the given condition is true then update the new state (only if defer from the current state) according to the confirmed unspent amount.
positive => Refundable
zero => if claim then Complete else Failed

lib/core/src/chain_swap.rs
```diff
@@ -62,6 +90,87 @@ impl ChainSwapStateHandler {
         }
     }
 
+    pub(crate) async fn rescan_chain_swaps(&self) -> Result<()> {
```
roeierez, line 93: NIT: `rescan_incoming_chain_swaps` ?

lib/core/src/chain_swap.rs
```diff
@@ -62,6 +90,87 @@ impl ChainSwapStateHandler {
...
+            .lock()
+            .await
+            .script_list_unspent(script_pubkey.as_script())?
+            .iter()
+            .filter(|u| u.height > 0)
+            .map(|u| u.value)
+            .sum())
+    }
+
+    async fn rescan_chain_swap(&self, swap: &ChainSwap, current_height: u32) -> Result<()> {
```
roeierez, line 131: NIT: `rescan_incoming_chain_swap` ?

roeierez (changes requested):
Looks good. Dropped some NITs.

dangeross:
Apologies, I've included in the PR adding the RefundPending state because it made sense to use it when monitoring the swap

lib/core/src/chain_swap.rs
```diff
@@ -104,46 +106,56 @@ impl ChainSwapStateHandler {
...
+            let confirmed_unspent_sat: u64 = self
+                .bitcoin_chain_service
+                .lock()
+                .await
+                .script_list_unspent(script_pubkey.as_script())?
+                .iter()
+                .filter(|u| u.height > 0)
+                .map(|u| u.value)
+                .sum();
+            if confirmed_unspent_sat > 0 && swap.state != Refundable {
```
roeierez, line 141: If the swap is in RefundPending state here (and the refund is not confirmed yet) won't this change its state to Refundable ?

roeierez: OK, reading the docs of script_list_unspent it seems it takes into account mempool spends so in fact if the refund is broadcasted (RefundPending state) then we should get zero unspent.
So it seems it will move RefundPending to Failed/Complete ? I think we are not taking into account the mempool spends here.

dangeross: True, fixed

roeierez (approved):
LGTM

PR #329, opened 2024-06-24
Use swapper proxy to extract referral ID

### LINKED ISSUE #135: Create a proxy server for boltz swapper
This is so we can add the referrer_id in the API so boltz knows the request is from breez and rev share with us.

ok300:
Closes #135

### DIFF
--- a/lib/bindings/src/breez_liquid_sdk.udl
+++ b/lib/bindings/src/breez_liquid_sdk.udl
@@ -245,6 +245,7 @@ enum PaymentError {
 
 dictionary Config {
     string boltz_url;
+    string? referral_id;
     string liquid_electrum_url;
     string bitcoin_electrum_url;
     string working_dir;
--- a/lib/bindings/src/breez_liquid_sdk.udl
+++ b/lib/bindings/src/breez_liquid_sdk.udl
@@ -245,7 +245,6 @@ enum PaymentError {
 
 dictionary Config {
     string boltz_url;
-    string? referral_id;
     string liquid_electrum_url;
     string bitcoin_electrum_url;
     string working_dir;
--- a/lib/bindings/src/breez_liquid_sdk.udl
+++ b/lib/bindings/src/breez_liquid_sdk.udl
@@ -244,7 +244,6 @@ enum PaymentError {
 };
 
 dictionary Config {
-    string boltz_url;
     string liquid_electrum_url;
     string bitcoin_electrum_url;
     string working_dir;
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -57,42 +57,45 @@ pub struct LiquidSdk {
 impl LiquidSdk {
     pub async fn connect(req: ConnectRequest) -> Result<Arc<LiquidSdk>> {
         // If we can fetch a proxy swapper URL, extract referral ID and dynamically set it in the config
-        let referral_id = match &req.config.network {
-            LiquidNetwork::Testnet => None,
+        let (boltz_api_base_url, referral_id) = match &req.config.network {
+            LiquidNetwork::Testnet => (None, None),
             LiquidNetwork::Mainnet => {
                 match BreezServer::new("https://bs1.breez.technology:443".into(), None) {
                     Ok(breez_server) => match breez_server.fetch_boltz_swapper_urls().await {
                         Ok(boltz_swapper_urls) => match boltz_swapper_urls.first().cloned() {
                             Some(first_url) => Url::parse(first_url.as_str())
                                 .map(|url| match url.query() {
-                                    None => None,
+                                    None => (None, None),
                                     Some(query) => {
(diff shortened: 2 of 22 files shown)


lib/bindings/src/breez_liquid_sdk.udl
```diff
@@ -245,6 +245,7 @@ enum PaymentError {
 
 dictionary Config {
     string boltz_url;
+    string? referral_id;
```
roeierez, line 248: We don't want to expose that in the config. 
Also let's remove the boltz_url from the config since now we get it from the remote server.

lib/core/src/sdk.rs
```diff
@@ -56,7 +56,50 @@ pub struct LiquidSdk {
...
+                            Some(first_url) => Url::parse(first_url.as_str())
+                                .map(|url| match url.query() {
+                                    None => (None, None),
+                                    Some(query) => {
+                                        let boltz_api_base_url = url
+                                            .domain()
+                                            .map(|domain| format!("https://{domain}/v2"));
+                                        let parts: Vec<String> =
+                                            query.split('=').map(Into::into).collect();
+                                        let referral_id = parts.get(1).cloned();
```
roeierez, line 75: I think it is better to save the query as is and append it to the create requests, not need to parse it for the specific param.

lib/core/src/sdk.rs
```diff
@@ -56,7 +56,50 @@ pub struct LiquidSdk {
...
+                                        (boltz_api_base_url, referral_id)
+                                    }
+                                })
+                                .unwrap_or_default(),
```
roeierez, line 79: Can we expose this block of code to its own function?

roeierez (changes requested):

lib/core/src/sdk.rs
```diff
@@ -56,7 +56,50 @@ pub struct LiquidSdk {
 
 impl LiquidSdk {
     pub async fn connect(req: ConnectRequest) -> Result<Arc<LiquidSdk>> {
-        let config = req.config;
+        // If we can fetch a proxy swapper URL, extract referral ID and dynamically set it in the config
+        let (boltz_api_base_url, referral_id) = match &req.config.network {
```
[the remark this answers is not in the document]
roeierez: Also perhaps we can persist the last swapper url so in case we get error from the service we can use the persisted value.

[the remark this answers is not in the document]
ok300: > I think it is better to pass the url directly to the swapper instance and let it parse the query instead of doing it here.

Done: 56e0942ee0737362cc10055468f91f24155cc95b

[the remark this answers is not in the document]
ok300: > Also perhaps we can persist the last swapper url so in case we get error from the service we can use the persisted value.

Done: 31a037cc9c7fcfc8eeed2c07b2f6c3a92d5a07fd

lib/core/src/sdk.rs
```diff
@@ -56,14 +56,21 @@ pub struct LiquidSdk {
 
 impl LiquidSdk {
     pub async fn connect(req: ConnectRequest) -> Result<Arc<LiquidSdk>> {
-        let config = req.config;
-        let sdk = LiquidSdk::new(config, req.mnemonic)?;
+        let breez_server = BreezServer::new("https://bs1.breez.technology:443".into(), None)?;
+        let swapper_urls = breez_server.fetch_boltz_swapper_urls().await?;
```
roeierez, line 60: Perhaps silent handle errors here? I am thinking in case the user is offline this will return error on restart.

lib/core/src/swapper/mod.rs
```diff
@@ -147,15 +148,38 @@ pub trait Swapper: Send + Sync {
...
+                            let referral_id = parts.get(1).cloned();
+                            (api_base_url, referral_id)
+                        }
+                    })
+                    .unwrap_or_default(),
+                None => (None, None),
+            },
+        };
+
+        let boltz_url = boltz_api_base_url.unwrap_or(config.boltz_url.clone());
```
roeierez, line 178: I think we should remove the boltz_url from the config as we take it from the breez server and the user won't understand why it is not in use. It is actually sort of fallback so I suggest to have it as constant inside the swapper and not expose it in the configuration.

roeierez (changes requested):
Looks good. Two last comments.

lib/core/src/swapper/mod.rs
ok300: Done: d7634421ca7b7fdcf4d2437331b1e15f3de9bab8

lib/core/src/sdk.rs
ok300: Done: 7bcc63b90a3406535863c9f3b62bf646bdfc46d4

roeierez (approved):
LGTM

PR #330, opened 2024-06-25
Verify chain swap server lockup transaction

### LINKED ISSUE #325: Add chain swap server lockup verification

dangeross:
This PR:

- Moves the liquid transaction verification into the chain service and adds bitcoin transaction verification
- For chain swap statuses `transaction.server.mempool` & `transaction.server.confirmed`: 
  - Verify the transaction exists in the script history
  - Verify not RBF if only a mempool status
  - Verify the amount is at least the minimum amount of the swap

Fixes #325

### DIFF
--- a/lib/core/src/chain/bitcoin.rs
+++ b/lib/core/src/chain/bitcoin.rs
@@ -72,17 +72,17 @@ impl ElectrumClient {
         info!("Fetching script history for {}", script_hash);
         let mut script_history = vec![];
 
-        let mut retry = 1;
-        while retry < retries {
+        let mut retry = 0;
+        while retry <= retries {
             script_history = self.get_script_history(script)?;
             match script_history.is_empty() {
                 true => {
+                    retry += 1;
                     info!(
                         "Script history for {} got zero transactions, retrying in {} seconds...",
                         script_hash, retry
                     );
                     thread::sleep(Duration::from_secs(retry));
-                    retry += 1;
                 }
                 false => break,
             }
--- a/lib/core/src/chain/bitcoin.rs
+++ b/lib/core/src/chain/bitcoin.rs
@@ -1,20 +1,22 @@
-use std::collections::HashMap;
+use std::{thread, time::Duration};
 
-use anyhow::Result;
+use anyhow::{anyhow, Result};
+use async_trait::async_trait;
+use boltz_client::{Address, ToHex};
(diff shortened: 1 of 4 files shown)


roeierez:
This looks pretty good to me. I will wait for the rebase and getting out of draft.

roeierez (approved):
LGTM

lib/core/src/chain/bitcoin.rs
```diff
@@ -58,8 +60,38 @@ impl ElectrumClient {
...
+        retries: u64,
+    ) -> Result<Vec<History>> {
+        let script_hash = sha256::Hash::hash(script.as_bytes())
+            .to_byte_array()
+            .to_hex();
+        info!("Fetching script history for {}", script_hash);
+        let mut script_history = vec![];
+
+        let mut retry = 1;
+        while retry < retries {
```
ok300, line 76: Its easy to get into an endless loop if `retries` is 0.

Maybe go for
```rust
let mut retry = 0;
while retry < retries { ... }
```

lib/core/src/chain/liquid.rs
```diff
@@ -56,6 +70,35 @@ impl HybridLiquidChainService {
...
+        script: &Script,
+        retries: u64,
+    ) -> Result<Vec<History>> {
+        let script_hash = sha256::Hash::hash(script.as_bytes())
+            .to_byte_array()
+            .to_hex();
+        info!("Fetching script history for {}", script_hash);
+        let mut script_history = vec![];
+
+        let mut retry = 1;
```
ok300, line 85: Same issue with retries.

ok300 (commented):
One or two NITs, otherwise LGTM.

lib/core/src/chain/bitcoin.rs
ok300: My mistake, its not an endless loop, its an edge case that would cause it to not fetch the history at all.

lib/core/src/chain/liquid.rs
ok300, line 85: Same retry issue here.

lib/core/src/chain/bitcoin.rs
dangeross: You're right, fixed

ok300 (approved):

PR #331, opened 2024-06-25
Add `FiatAPI` methods to Liquid SDK

### LINKED ISSUE #262: Add support for Fiat Currency & exchange rates

ok300:
Closes #262

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -111,7 +111,7 @@ impl LiquidSdk {
             bitcoin_chain_service.clone(),
         )?;
 
-        let breez_server = BreezServer::new("https://bs1.breez.technology:443".into(), None)?;
+        let breez_server = BreezServer::new(PRODUCTION_BREEZSERVER_URL.into(), None)?;
 
         let sdk = Arc::new(LiquidSdk {
             config: config.clone(),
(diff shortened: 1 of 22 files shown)


roeierez (approved):
LGTM

lib/core/src/sdk.rs
```diff
@@ -109,6 +110,8 @@ impl LiquidSdk {
             bitcoin_chain_service.clone(),
         )?;
 
+        let breez_server = BreezServer::new("https://bs1.breez.technology:443".into(), None)?;
```
erdemyerebasmaz, line 113: Looking at the implementation on Breez SDK, should Breez server URL be part of `Config`? 
If applicable here, Production & Staging(Mainnet & Testnet) could use different URLs as part of default config.

erdemyerebasmaz (approved):
Looks good. Have a single question regarding Breez server url.

lib/core/src/sdk.rs
ok300: Good question.

The main difference I see is Breez SDK has Production / Staging configs, whereas Liquid SDK has Mainnet / Testnet configs.

So at the moment I don't see a place for the Staging breez server URL, unless we add e.g. Mainnet Staging.

@roeierez what do you think?

roeierez: staging is no longer exist in breez_sdk (we should actually remove it) soI think we don't need it at the config.
We can add a  `new_mainnet` function in BreezServer sdk-common implementation to return an instance already populated with the mainnet URL so we don't have to pass it from the lqiuid sdk.

ok300: The Breez Server URLs are defined in static variables in Breez SDK.

I moved them to `sdk-common` so they can easily be referenced here: https://github.com/breez/breez-sdk/pull/1023 . I can then reference `PRODUCTION_BREEZSERVER_URL` here.

Would this be ok as a solution?

roeierez: IMO yes.

ok300: Done: 71647e33753248169479d3b731a8ec703fb5613e

PR #332, opened 2024-06-25
Use lowball fees only on lockup and cooperative refund

roeierez:
This fixes a regression introduced in HybridChainService PR where the non cooperative refund also used the lowball fee rate.
Also fixes the broadcast and script_history of chain service on testnet.

### DIFF
--- a/lib/core/src/chain/liquid.rs
+++ b/lib/core/src/chain/liquid.rs
@@ -11,7 +11,7 @@ use reqwest::Response;
 use serde::Deserialize;
 use std::str::FromStr;
 
-use crate::model::Config;
+use crate::model::{Config, LiquidNetwork};
 
 const LIQUID_ESPLORA_URL: &str = "https://lq1.breez.technology/liquid/api";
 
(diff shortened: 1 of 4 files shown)


lib/core/src/chain/liquid.rs
```diff
@@ -61,35 +65,49 @@ impl LiquidChainService for HybridLiquidChainService {
...
+                let url = format!("{}/scripthash/{}/txs", LIQUID_ESPLORA_URL, script_hash);
+                // TODO must handle paging -> https://github.com/blockstream/esplora/blob/master/API.md#addresses
+                let response = get_with_retry(&url, 3).await?;
+                let json: Vec<EsploraTx> = response.json().await?;
+
+                let history: Vec<History> = json.into_iter().map(Into::into).collect();
+                Ok(history)
+            }
+            LiquidNetwork::Testnet => {
+                let mut history_vec = self.electrum_client.get_scripts_history(&[script])?;
```
dangeross, line 106: There should be a `electrum_client.script_get_history(script)` to get a single script history

roeierez: We are using ElectrumClient from lwk which wraps the electrum_client.Client which doesn't expose the inner client.
Do you think it worth saving also the Client on the struct only for this call?

dangeross: No, I just thought it would be a bit cleaner

ok300 (approved):

dangeross (approved):
LGTM

PR #341, opened 2024-06-27
Receive onchain: add min/max amount to prepare response

ok300:
This brings the Liquid SDK in alignment with the Breez SDK, where the swap info contains the min/max that the payer can send onchain to the SDK.

This also helps make the docs clearer and the UX more user-friendly.

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -1208,8 +1208,8 @@ impl LiquidSdk {
         Ok(PrepareReceiveOnchainResponse {
             amount_sat,
             fees_sat: pair.fees.boltz(amount_sat) + claim_fees_sat + server_fees_sat,
-            min_payer_amount_sat: pair.limits.maximal,
-            max_payer_amount_sat: pair.limits.minimal,
+            min_payer_amount_sat: pair.limits.minimal,
+            max_payer_amount_sat: pair.limits.maximal,
         })
     }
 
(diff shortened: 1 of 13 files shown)


lib/core/src/sdk.rs
```diff
@@ -1208,6 +1208,8 @@ impl LiquidSdk {
         Ok(PrepareReceiveOnchainResponse {
             amount_sat,
             fees_sat: pair.fees.boltz(amount_sat) + claim_fees_sat + server_fees_sat,
+            min_payer_amount_sat: pair.limits.maximal,
+            max_payer_amount_sat: pair.limits.minimal,
```
dangeross, line 1212: These seem to be the wrong way around. Also in the chain swap, the min has to be at least the `amount_sat`

ok300: > These seem to be the wrong way around.

Fixed: 0095d4f140b6b5d76a763f88b8813a41b623fa3f

> Also in the chain swap, the min has to be at least the `amount_sat`

That brings a good point. If we have `prepare_receive_onchain(payer_amount)`, adding min/max in the response goes against the specified amount. So in that sense, maybe we shouldn't have min/max in the receive onchain response at all. Would you agree? @roeierez what do you think?

dangeross: I think for consistency it makes sense to add this to a `receive_onchain_limits` method. The SDK user needs to know beforehand what are the limits of receiving via an onchain tx. Receive amount needs to be above the minimum limit (and agreed amount with the swapper) so I think it makes sense to find out what the minimum is first.

roeierez: Perhaps we can have one method for onchain limits that will return 4 values? Min/max for both send/receive?

ok300: I opened https://github.com/breez/breez-liquid-sdk/pull/343 to address this.

Will abandon this PR, as the min/max limits shouldn't be in the Prepare Receive Onchain response.

ok300:
Closing this in favor of #343

PR #343, opened 2024-06-28
Add `fetch_onchain_limits` service method

ok300:
This lets callers find out the limits before starting the swap.

Addresses https://github.com/breez/breez-liquid-sdk/pull/341#discussion_r1658685922


lib/core/src/model.rs
```diff
@@ -186,21 +185,19 @@ pub struct ReceivePaymentResponse {
...
-            min_payer_amount_sat: pair_limits.minimal,
-            max_payer_amount_sat_zero_conf: pair_limits.maximal_zero_conf,
-        }
-    }
+    /// Minimum swap amount for a Send Swap to be valid
+    pub send_min_payer_amount_sat: u64,
+    /// Maximum swap amount for a Send Swap to be valid
+    pub send_max_payer_amount_sat: u64,
+    /// Maximum swap amount which the swapper will accept for zero-conf Send Swaps
+    pub send_max_payer_amount_sat_zero_conf: u64,
```
dangeross, line 193: For send it will be receiver_amount_sat

dangeross: I would prefer to shorten the field names here, but i understand the reasons.

lib/core/src/sdk.rs
```diff
@@ -875,15 +875,29 @@ impl LiquidSdk {
...
             .get_chain_pairs(Direction::Outgoing)?
             .ok_or(PaymentError::PairsNotFound)
-            .map(|pair| pair.limits)
-            .map(Into::into)
+            .map(|pair| pair.limits)?;
+        let receive_limits = self
+            .swapper
+            .get_chain_pairs(Direction::Incoming)?
+            .ok_or(PaymentError::PairsNotFound)
+            .map(|pair| pair.limits)?;
```
dangeross, line 891: This will be two network requests, maybe the swapper function needs changing to return both directions

lib/core/src/model.rs
ok300: > For send it will be receiver_amount_sat

AFAIK the limits refer to "swap amount", which is the amount being swapped.

I think its that's neither `payer_amount` nor `receiver_amount`:

```
// For Send Onchain
payer_amount_sat = swap_amount + lockup_tx_fees

// For Receive Onchain
receiver_amount_sat = swap_amount - service_fee - claim_tx_fees
```

Is this correct?

If so, we can't really return payer/receiver limits, as that depends on the feerate used. Or am I missing something?

roeierez: About the long field names perhaps we can have a nested struct:
```
struct Limits {
  min_sats: u64
  max_sats: u64
  max_zero_conf_sats: u64
}
```

The we can use it in the outer struct:

```
struct OnchainPaymentLimitsResponse {
 send: Limits
 receive: Limits
}
```

dangeross: Looking at the boltz-backend, the pair limits validate the user_lock_amount. 

So for receiving its easy: payer_amount_sat = user_lock_amount

For sending we set the server_lock_amount (receiver_amount_sat + claim_fees). 
The user_lock_amount can be calculated from that: 
```
user_lock_amount = ((server_lock_amount + server_fee) / (1 - fee_percent)).ceil()
```

So we could for the sending limits recalculate to get the equivalent server_lock_amount limits

```
minimal = (limit.minimal + claim_fees - boltz_fee(limit.minimal + claim_fees) - server_fee).floor()
maximal = (limit.maximal - boltz_fee(limit.maximal) - server_fee - claim_fees).floor()
```

lib/core/src/sdk.rs
ok300: Done: f6477695730baa1bd47f8995a03ce882f80f69e4

lib/core/src/model.rs
ok300: As discussed, we stick to the swap amount for both Send and Receive.

I updated (and shortened) the field names, and updated the field docs: 7095b34b813bf933e7518a4beb92171d6eb8316a

lib/core/src/model.rs
```diff
@@ -183,6 +183,23 @@ pub struct ReceivePaymentResponse {
...
 }
 
+#[derive(Debug, Serialize)]
+pub struct OnchainPaymentLimitsResponse {
+    /// Minimum swap amount for a Send Payment to be valid
+    pub send_min_amount_sat: u64,
+    /// Maximum swap amount for a Send Payment to be valid
+    pub send_max_amount_sat: u64,
+    /// Maximum swap amount which the swapper will accept for zero-conf Send Payments
+    pub send_max_amount_sat_zero_conf: u64,
```
dangeross, line 193: I think it reads better like this, also with the unit at the end
```suggestion
    pub send_max_zero_conf_amount_sat: u64,
```

lib/core/src/model.rs
```diff
@@ -183,6 +183,23 @@ pub struct ReceivePaymentResponse {
...
+
+    /// Minimum swap amount for a Receive Payment to be valid
+    pub receive_min_amount_sat: u64,
+    /// Maximum swap amount for a Receive Payment to be valid
+    pub receive_max_amount_sat: u64,
+    /// Maximum swap amount which the swapper will accept for zero-conf Receive Payments
+    pub receive_max_amount_sat_zero_conf: u64,
```
dangeross, line 200:
```suggestion
    pub receive_max_zero_conf_amount_sat: u64,
```

dangeross (commented):
Looks good, just a nit about the field names

lib/core/src/model.rs
roeierez: @ok300 don't you think it is better to reuse one struct:

```
struct Limits {
  min_sats: u64
  max_sats: u64
  max_zero_conf_sats: u64
}
```

In both send and receive as I suggested above?

ok300: Done: 1c620b5185206632daa7dc9de7dea3741cc9e733

lib/core/src/model.rs
ok300: Done

lib/core/src/model.rs
ok300: Done

roeierez (approved):
LGTM

dangeross (approved):
LGTM

PR #346, opened 2024-06-28
Tests: Add unit tests for primary SDK methods

hydra-yse:
The following PR aims at defining unit tests for the locally scoped SDK methods, specifically those addressing swap and chain payment updates. 

TODO:
- [x] Send Swap Flow
- [X] Receive Swap Flow


lib/core/src/sdk.rs
```diff
@@ -1674,3 +1674,306 @@ impl ReconnectHandler for SwapperReconnectHandler {
...
+            create_mock_tx,
+            persist::{new_persister, new_receive_swap, new_send_swap},
+            sdk::new_liquid_sdk,
+            status_stream::MockStatusStream,
+            swapper::MockSwapper,
+        },
+    };
+    use paste::paste;
+
+    macro_rules! trigger_swap_update {
```
roeierez, line 1703: What is the benefit here for macro instead of a function?

lib/core/src/test_utils/wallet.rs
```diff
@@ -0,0 +1,10 @@
+#![cfg(test)]
+
+use crate::{model::Config, wallet::LiquidOnchainWallet};
+use anyhow::Result;
+
+use super::TEST_MNEMONIC;
+
+pub(crate) fn new_onchain_wallet(config: &Config) -> Result<LiquidOnchainWallet> {
```
roeierez, line 8: I think we should mock both the onchain_wallet and the chain_service.
From looking at the tests it shouldn't too much of a headache since they are not really in use atm.
Network effect in tests makes them very hard to maintain.

roeierez (changes requested):
Looks good!

lib/core/src/sdk.rs
hydra-yse: We can't use a function because the return type differs between each kind. It's possible but would be more verbose to implement and we would have to pattern match the `Swap` result within the tests

lib/core/src/test_utils/wallet.rs
hydra-yse: Based on what we also said during the call, I agree. It doesn't cost much to mock the build_tx methods and adds less overhead to the simulation

lib/core/src/receive_swap.rs
```diff
@@ -320,7 +322,7 @@ impl ReceiveSwapStateHandler {
         verify_confirmation: bool,
     ) -> Result<()> {
         // Looking for lockup script history to verify lockup was broadcasted
-        let script = receive_swap.get_swap_script()?;
+        let script = receive_swap.get_swap_script().unwrap();
```
dangeross, line 325: Is there a reason to change to an unwrap? It will panic if it can't

lib/core/src/send_swap.rs
```diff
@@ -74,8 +74,8 @@ impl SendSwapStateHandler {
             Ok(SubSwapStates::InvoiceSet) => {
                 match (swap.state, swap.lockup_tx_id.clone()) {
                     (PaymentState::Created, None) | (PaymentState::TimedOut, None) => {
-                        let create_response = swap.get_boltz_create_response()?;
-                        let lockup_tx = self.lockup_funds(id, &create_response).await?;
+                        let create_response = swap.get_boltz_create_response().unwrap();
+                        let lockup_tx = self.lockup_funds(id, &create_response).await.unwrap();
```
dangeross, line 78: Same with the `unwrap()`, is there a reason?

hydra-yse: Good catch Ross, I used those unwraps for debugging, thanks

dangeross (approved):
LGTM

roeierez (approved):
LGTM

PR #349, opened 2024-07-01
RN fixes

dangeross:
This PR fixes the network type used in the example app and the param when calling setLogger


erdemyerebasmaz (approved):
LGTM

PR #351, opened 2024-07-01
Improve MRH amount validation

ok300:
Instead of a specific BIP21 amount, we allow any amount that is at or below the initially expected amount.


erdemyerebasmaz (approved):

ok300:
Abandoning this in favor of the Boltz fix, which they already deployed.

PR #357, opened 2024-07-03
Attempt to load Breez Liquid library via `ExternalLibrary.process()` if t fails to load via `open()`

erdemyerebasmaz:
`ExternalLibrary.process()` resolves any symbol in a library currently loaded with global visibility.


dangeross (approved):
LGTM

PR #360, opened 2024-07-04
Send: Fix refund `InsufficientFunds` error when calculating broadcast fee

hydra-yse:
This PR adds the method `estimate_refund_fees` so we can calculate the refund transaction without having to build it through lwk.

**Note:** There are some comments I left in regards to some unanswered questions, let me know what you think.

### DIFF
--- a/lib/core/src/send_swap.rs
+++ b/lib/core/src/send_swap.rs
@@ -204,7 +204,7 @@ impl SendSwapStateHandler {
         let lockup_tx = self
             .onchain_wallet
             .build_tx(
-                self.config.lowball_fee_rate(),
+                self.config.lowball_fee_rate_msat_per_vbyte(),
                 &create_response.address,
                 create_response.expected_amount,
             )
--- a/lib/core/src/send_swap.rs
+++ b/lib/core/src/send_swap.rs
@@ -362,7 +362,7 @@ impl SendSwapStateHandler {
         let output_address = self.onchain_wallet.next_unused_address().await?.to_string();
 
         let cooperative_refund_tx_fees_sat =
-            utils::estimate_refund_fees(swap, &self.config, &output_address, false)?;
+            utils::estimate_refund_fees(swap, &self.config, &output_address, true)?;
         let refund_res = self.swapper.refund_send_swap_cooperative(
             swap,
             &output_address,
--- a/lib/core/src/send_swap.rs
+++ b/lib/core/src/send_swap.rs
@@ -1,3 +1,4 @@
+use std::time::Duration;
 use std::{str::FromStr, sync::Arc};
 
 use anyhow::{anyhow, Result};
@@ -24,6 +25,10 @@ use crate::{
     model::{PaymentState, PaymentTxData, PaymentType},
     persist::Persister,
 };
+
+pub(crate) const MAX_REFUND_ATTEMPTS: u8 = 6;
+pub(crate) const REFUND_REATTEMPT_DELAY_SECS: u64 = 10;
+
 #[derive(Clone)]
 pub(crate) struct SendSwapStateHandler {
     config: Config,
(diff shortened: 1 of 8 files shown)


hydra-yse:
Why is the `LOWBALL_FEE_RATE_SAT_PER_VBYTE` multiplied by 1000 here?

https://github.com/breez/breez-liquid-sdk/blob/abe5e385330f2060602be3a829202dc4278a7762/lib/core/src/model.rs#L72-L77

lib/core/src/send_swap.rs
```diff
@@ -359,25 +359,24 @@ impl SendSwapStateHandler {
...
 
-        let fee = self
-            .onchain_wallet
-            .build_tx(None, &output_address, amount_sat)
-            .await?
-            .all_fees()
-            .values()
-            .sum();
+        let cooperative_refund_tx_fees_sat =
+            utils::estimate_refund_fees(swap, &self.config, &output_address, false)?;
```
hydra-yse, line 365: Currently, setting the `is_lowball` check to false seems to be the only option that works on mainnet (thus defaulting to 0.1 sat/vbyte). Are we sure we can broadcast lowball through boltz when attempting refunds (using 0.01 causes a min_relay_fee not met error, as it should be >129 but is actually 14).

dangeross:
> Why is the `LOWBALL_FEE_RATE_SAT_PER_VBYTE` multiplied by 1000 here?

@hydra-yse It's because the LWK wallet when building the tx divides the fee rate by 1000

lib/core/src/send_swap.rs
roeierez: @hydra-yse are you sure it fails for the cooperative case with lowball fees?  I am asking because the cooperative case seems to be using boltz api which supports lowball for the cooperative claim.

roeierez (changes requested):

lib/core/src/utils.rs
```diff
@@ -49,3 +52,38 @@ pub(crate) fn deserialize_tx_hex(tx_hex: &str) -> Result<Transaction> {
...
+    );
+    let swap_tx = boltz_client::LBtcSwapTxV2::new_refund(
+        swap_script,
+        &output_address.to_string(),
+        &electrum_config,
+        config.liquid_electrum_url.clone(),
+        swap.id.clone(),
+    )?;
+    let dummy_fees = Amount::from_sat(100);
+    let dummy_tx = swap_tx.sign_refund(&swap.get_refund_keypair()?, dummy_fees, None)?;
```
dangeross, line 78: Maybe there is a difference in the tx vsize in a cooperative vs non-cooperative refund that we should take into account...

I guess it's actually not easy to do without doing any of the cooperative part, I see in boltz-rust there is a TODO to hardcode the cooperative spend size
https://github.com/SatoshiPortal/boltz-rust/blob/6ba4f48bebfe5bc05ce65fa5f44c1cfd9c42ffce/src/swaps/liquid.rs#L1183-L1193

lib/core/src/utils.rs
```diff
@@ -49,3 +52,38 @@ pub(crate) fn deserialize_tx_hex(tx_hex: &str) -> Result<Transaction> {
...
+
+    let fee_rate = if is_lowball {
+        config
+            .lowball_fee_rate()
```
dangeross, line 82: Yeah the `* 1000` should be removed so the absolute fee is calculated correctly

lib/core/src/send_swap.rs
dangeross: I see there are some conditions to the lowball in the boltz-backend
> // Only allow lowball when Reverse Swaps are being claimed or Submarine Swaps being refunded
   // and no Submarine Swap lockup happens
   // That prevents transactions that we could accept 0-conf for from being broadcast through our API
    const isSwapRelated = swapsFunded.length === 0 && swapsSpent.length > 0;

roeierez: @dangeross do you understand how could be a refundable submarine swap without lockup? isn't the point that the refund tx spends the lockup?

dangeross: True, the comment doesn't really make sense. From looking at it it allows lowball on the lockup, no lowball on the refund
https://github.com/BoltzExchange/boltz-backend/blob/5d2cab089771cfb96be23349cfc3432c7b174e25/lib/service/Service.ts#L707-L722

roeierez: Looking further at the code here: https://github.com/BoltzExchange/boltz-backend/blob/5d2cab089771cfb96be23349cfc3432c7b174e25/lib/service/TransactionFetcher.ts#L184

It seems that boltz means by the comment that they don't want to allow broadcasting lockups via their API.
It seems that they demand that the broadcasted transaction outputs are not related to any lockup script.
Which seems meet our criteria for the refund as they also validates it spends a current lockup script.

ok300: @hydra-yse I see that `util::estimate_refund_fees` is hardcoding `is_cooperative = None` when singing the refund tx. Could it be that's why it fails to broadcast with the lowball feerate? Have you tried setting the `is_cooperative` arg of `sign_refund` to `Some` if `is_lowball` is true?

lib/core/src/utils.rs
ok300: No, we should instead make it obvious what unit we're using (e.g. rename `lowball_fee_rate` -> `lowball_fee_rate_msat_per_vbyte`).

LWK initializes the feerate to `100.0`: https://github.com/Blockstream/lwk/blob/ef7b0058439d496754f1d025f76c711baed8dc19/lwk_wollet/src/tx_builder.rs#L79 which is the 0.1 s/vB.

This means when we give a feerate arg to LWK, e.g. when building the LWK transactions, the our `LOWBALL_FEE_RATE_SAT_PER_VBYTE = 0.1` needs to be multiplied by 1000, which it is at the moment, to match the different unit expected by LWK.

ok300: We could even have two feerate methods, for use internally (in s/vB) and for use with LWK (ms/vB).

ok300 (approved):
LGTM.

I think you have to also re-generate the flutter bridge files.

hydra-yse:
The refund seems to work properly, but I seem to be getting this error from time to time (restarting the application makes it succeed): 
```[2024-07-04 11:56:52.002 ERROR breez_liquid_sdk::sdk:268] Failed to handle Send Swap 5xZwtqyMgNoR update: Generic error: Could not contact servers: "Bad URL: failed to parse URL: EmptyHost: empty host"```
Do you think this is an error on Boltz's side? Or something related to my connectivity? Can you reproduce this (simply by attempting a refund)?

**Comment by ok300** [comment]
I haven't seen it, but looking at the code I see it comes when converting a `boltz_client::error::Error::HTTP` to one of our error types. In the boltz crate, that error is used for ureq errors, which is their HTTP client.

If that's not related to this PR, maybe open a separate issue to track it?

**Comment by hydra-yse** [comment]
> If that's not related to this PR, maybe open a separate issue to track it?

It is related to this PR as the error is introduced while executing `estimate_refund_fees`, though I'm not sure why it seems to automatically fix itself on app restart.

**Comment by ok300** [comment]
Can you add more logs from just before this happened, so we can better pin-point which operation is causing the failure?

**Comment by hydra-yse** [comment]
@ok300 @roeierez @dangeross I've found the root cause of the error; it's due to the fact that the lockup tx is not confirmed yet and thus we cannot retrieve the utxos to build the refund transaction. Do you suggest we add a loop with a max attempts timeout to avoid this issue within the `on_new_status` thread, or should we keep things as-is and just wait for the user to relaunch the application?

**Comment by ok300** [comment]
The delay and retry count make sense to me, as it's checking every 10 seconds and waiting up to 1 min.

File: lib/core/src/send_swap.rs [hunk 4]
```diff
@@ -150,25 +155,45 @@ impl SendSwapStateHandler {
...
-                                None,
-                                None,
-                                Some(&refund_tx_id),
-                            )
-                            .await?;
+                            None => {
+                                warn!("Send Swap {id} is in an unrecoverable state: {swap_state:?}, and lockup tx has been broadcast. Attempting refund.");
+
+                                let mut refund_attempts = 0;
+                                while refund_attempts < MAX_REFUND_ATTEMPTS {
```
roeierez, line 167: Sorry for responding late but won't this block the task that handles swap updates?

PR #367, opened 2024-07-05
Fix transition from `RefundPending` to `Failed`

ok300:
This PR includes `RefundPending` state in the list_pending filter. This is done for swaps that can be refunded, like Chain Swaps and Send Swaps.

Without this, refunded payments never transition away from `RefundPending` to `Failed` (refunded), because they're not considered a pending swap when handling the final transition.


dangeross (approved):
LGTM

PR #369, opened 2024-07-08
List payments filtering and paging

### LINKED ISSUE #358: Return payment list in reverse chronological order
Payment list items are currently returned from oldest to newest.

dangeross:
This PR add payment filtering by payment type, from/to timestamp and paging. It also sorts payments by reverse chronological order

Fixes #358 #375

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -5,6 +5,7 @@ mod migrations;
 pub(crate) mod receive;
 pub(crate) mod send;
 
+use std::collections::HashSet;
 use std::{fs::create_dir_all, path::PathBuf, str::FromStr};
 
 use anyhow::{anyhow, Result};
@@ -124,7 +125,12 @@ impl Persister {
         .concat())
     }
 
-    fn select_payment_query(&self, where_clause: Option<&str>) -> String {
+    fn select_payment_query(
+        &self,
+        where_clause: Option<&str>,
+        offset: Option<u32>,
+        limit: Option<u32>,
+    ) -> String {
         format!(
             "
             SELECT
@@ -174,8 +180,13 @@ impl Persister {
             AND                                  -- Filter out refund txs from Chain Swaps
                 ptx.tx_id NOT IN (SELECT refund_tx_id FROM chain_swaps WHERE refund_tx_id NOT NULL)
             AND {}
+            ORDER BY ptx.timestamp DESC
+            LIMIT {}
+            OFFSET {}
             ",
-            where_clause.unwrap_or("true")
+            where_clause.unwrap_or("true"),
+            limit.unwrap_or(u32::MAX),
+            offset.unwrap_or(0),
         )
     }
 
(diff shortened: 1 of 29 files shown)


erdemyerebasmaz (approved):
LGTM

ok300:
Doesn't this diverge from how Breez SDK does it?

AFAIK `list_payments` there shows them from oldest to newest.

erdemyerebasmaz:
> Doesn't this diverge from how Breez SDK does it?
> 
> AFAIK `list_payments` there shows them from oldest to newest.

They are sorted in reverse chronological order when retrieved from DB: [transactions.rs](https://github.com/breez/breez-sdk/blob/8563ff8372dbd22cbef5cb4d98603a4f4a6f4b64/libs/sdk-core/src/persist/transactions.rs#L305)

ok300:
You're right, my mistake.

ok300 (approved):

lib/core/src/sdk.rs
```diff
@@ -1475,12 +1476,13 @@ impl LiquidSdk {
         Ok(())
     }
 
-    /// Lists the SDK payments. The payments are determined based on onchain transactions and swaps.
+    /// Lists the SDK payments in reverse chronological order.
```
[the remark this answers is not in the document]
ok300: IMO it would make sense to add this.

ok300 (approved):
One NIT, otherwise tested and looks good.

lib/core/src/persist/mod.rs
```diff
@@ -304,12 +322,57 @@ impl Persister {
...
 
+fn filter_to_where_clause(
+    type_filters: Option<Vec<PaymentType>>,
+    from_timestamp: Option<i64>,
+    to_timestamp: Option<i64>,
+) -> String {
+    let mut where_clause: Vec<String> = Vec::new();
+
+    if let Some(t) = from_timestamp {
+        where_clause.push(format!("ptx.timestamp >= {t}"));
```
roeierez, line 333: I think for pending receive swap we don't have a tx. How. would this filter behave on this case? Perhaps it should fallback to the swap.created_at ?

roeierez (changes requested):
Looks good! Commented about one edge case

lib/core/src/persist/mod.rs
dangeross: I think at this point there will be no entry in the payment_tx_data table, so there's no payment to filter. Is that right @ok300?

ok300: For pending receives, there will be a pseudo-tx created: https://github.com/breez/breez-sdk-liquid/blob/da43f7ff78498b3210015a2fa1075c14190b88c9/lib/core/src/receive_swap.rs#L262-L271

To cover this edge-case, it would be enough to set the `timestamp` to `Some(utils::now())` in that snippet (it's now `None`).

ok300: Alternatively yes, the `receive_swap` table should have a `Created` entry, so the swap creation timestamp could be retrieved from there.

dangeross: I find it kind of messy because there are 3 swap tables joined in the query, so we have to check the created_at of each table. Setting the pseudo PaymentTxData timestamp to the ongoing_receive_swap.created_at (or now) has the same effect

roeierez: I think it makes sense to replace the None with now and then we don't need to fallback in the query.
Also setting to 'now' seems correct as this is the best time we know the transaction was broadcasted. 
What do you think @dangeross @ok300 ?

dangeross: Done [fbac4a2](https://github.com/breez/breez-sdk-liquid/pull/369/commits/fbac4a2c5f23ad70d4ae969df3ef67a09f74bba5)

roeierez: Looks good @dangeross but I think we should also change the `from_pending_swap` that puts the swap.created date on the payment because it is inconsistent with the new timestamp on the payment tx data.
Should we remove that method completely and use the `from_tx_data` instead now that we have the timestamp?

dangeross: I think the `from_pending_swap` is still needed to include swaps in the payments list that do not have an accompanying payment_tx_data entry (pseudo or otherwise).

roeierez: Aren't these the pending receives that you just added the timestamp for?

dangeross: Yeah right, looking at the SQL query and `sql_row_to_payment()` it will only ever return rows with payment_tx_data so this `match (tx, swap.clone())` is not needed
```
        match (tx, swap.clone()) {
            (None, None) => Err(maybe_tx_tx_id.err().unwrap()),
            (None, Some(swap)) => Ok(Payment::from_pending_swap(swap, payment_type)),
            (Some(tx), None) => Ok(Payment::from_tx_data(tx, None)),
            (Some(tx), Some(swap)) => Ok(Payment::from_tx_data(tx, Some(swap))),
        }
```
Does that sound right?

roeierez: That's what I thought. @ok300 can you confirm?

dangeross: I missed it the first time, there is a FULL JOIN between payment_tx_data and receive_swap, then LEFT JOINs on send_swaps and chain_swaps. So it will list receive_swaps without payment_tx_data and the match is valid.

I will add a coalesce to the timestamp filter to validate the first non null timestamp
```
coalesce(ptx.timestamp, rs.created_at) >= {t}
```

roeierez: I started with that and then got confused when we added the timestamp. Yes, I guess this is the right thing now.

roeierez (approved):

PR #388, opened 2024-07-11
Map `InsufficientFunds` LWK error to correct `PaymentError`

ok300:
An `InsufficientFunds` error was so far wrapped in a `PaymentError.generic(err: Lwk error: InsufficientFunds)`.

This PR maps it directly to the identically-named `PaymentError` value.


ok300:
Superseded by #386

PR #391, opened 2024-07-12
Allow custom `pay_onchain` claim fees

### LINKED ISSUE #390: Allow custom send onchain claim fees
Sending onchain should allow the user to set a custom BTC claim fee rate

- Add optional param in the request to set the fee rate
- Add method to get recommended BTC fees

dangeross:
This PR:
- adds an optional `sat_per_vbyte` param to `PreparePayOnchainRequest` to calculate the claim tx fees. 
- adds `recommended_fees()` method to find the optimal fee rate for the Bitcoin claim and refund.

Fixes #390

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -7,6 +7,7 @@ use async_trait::async_trait;
 use boltz_client::{swaps::boltzv2::*, util::secrets::Preimage, Bolt11Invoice};
 use boltz_client::{LockTime, ToHex};
 use chain::liquid::{HybridLiquidChainService, LiquidChainService};
+use chain_swap::ESTIMATED_BTC_CLAIM_TX_VSIZE;
 use futures_util::stream::select_all;
 use futures_util::StreamExt;
 use log::{debug, error, info};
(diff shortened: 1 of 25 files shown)


lib/core/src/sdk.rs
```diff
@@ -927,23 +932,23 @@ impl LiquidSdk {
 
         let receiver_amount_sat = req.prepare_res.receiver_amount_sat;
         let pair = self.validate_chain_pairs(Direction::Outgoing, receiver_amount_sat)?;
-        let claim_fees_sat = pair.fees.claim_estimate();
+        let claim_fees_sat = req.prepare_res.claim_fees_sat;
```
ok300, line 935: How will Boltz become aware of this custom BTC claim feerate? AFAIK they rely on the `pair_hash` arg in the `create_chain_swap` call below to calculate the claim fee.

dangeross: Boltz are using the pair hash and the current fee rates calculate a claim fee estimate. This is only an estimate for the user, as the user broadcasts the claim. So when creating the swap we are setting the serverLockAmount to `amount + custom claim fee`, then when creating the claim tx we can use any fee to broadcast the claim. We use the `custom claim fee`, leaving the `amount` as output

ok300 (approved):

roeierez (approved):
LGTM

PR #392, opened 2024-07-14
Test: Complete Chain Swap Tests

hydra-yse:
Here's a list of the remaining state tracking tests that have to be implemented:
- [X] Common (`TransactionMempool`, `TransactionConfirmed`, Failure States)
- [X] Outgoing (`TransactionServerMempool`, `TransactionServerConfirmed`, `Created`)
- [x] Incoming (`TransactionServerMempool`, `TransactionServerConfirmed`)


dangeross (commented):
Looking good

lib/core/src/sdk.rs
```diff
@@ -1728,17 +1753,21 @@ mod tests {
         ) => {{
             let swap = match $type {
                 "chain" => {
-                    let swap = new_chain_swap($direction.unwrap(), None);
+                    let swap = new_chain_swap(
+                        $args.direction.unwrap_or(Direction::Outgoing),
+                        $args.initial_payment_state,
+                        $args.accepts_zero_conf.unwrap_or(false),
```
ok300, line 1759: If `direction` and `accepts_zero_conf` default to these values here when not set, do they need to be optional at all? Why not make them non-optional and add a `impl Default for NewSwapArgs` with these 2 defaults?

lib/core/src/sdk.rs
```diff
@@ -1969,39 +2007,105 @@ mod tests {
                     ChainSwapStates::TransactionMempool,
                     ChainSwapStates::TransactionConfirmed,
                 ] {
-                    let mock_tx = create_mock_tx(sdk.onchain_wallet.clone()).await.unwrap();
-                    let mock_lockup_tx_id = mock_tx.txid().to_string();
-
+                    let mock_tx = TEST_LIQUID_TX.clone();
+                    let mock_tx_id = mock_tx.txid().to_string();
```
dangeross, line 2011: Should we test this the same way (but opposite) as testing TransactionServerMempool/TransactionServerConfirmed? So:
- Direction::Incoming => TEST_BITCOIN_TX
- Direction::Outgoing => TEST_LIQUID_TX

lib/core/src/sdk.rs
```diff
@@ -1950,7 +1988,7 @@ mod tests {
                 for status in &unrecoverable_states {
                     let persisted_swap = trigger_swap_update!(
                         "chain",
-                        Some(direction),
+                        NewSwapArgs::default().set_direction(direction),
```
roeierez, line 1991: Bellow ( on line 1999) we compare to either failed or refundable. I think it s better to split then according to the swap type and compare to a specific value.

roeierez (approved):
Small NIT, other than that LGTM!

ok300 (approved):

dangeross (approved):
LGTM

PR #398, opened 2024-07-16
Buy Bitcoin

### LINKED ISSUE #396: Add `buy_bitcoin` integration

dangeross:
This PR adds `prepare_buy_bitcoin()` and `buy_bitcoin()` SDK methods to initiate a purchase of Bitcoin via a Chain Swap. In the moonpay provider we generate a URL that locks the amount to purchase to the same value as defined in the `PrepareBuyBitcoinRequest`.

Builds on PR https://github.com/breez/breez-sdk/pull/1046 to use the common MoonpayProvider.

Fixes #396


lib/core/src/fiat/moonpay.rs
ok300, line 1: I see some duplication with Breez SDK.

Wouldn't it make sense to move the Breez SDK `moonpay.rs` to `sdk-common` and have it re-used here?

roeierez: It seems that if instead of passing SwapInfo/ChainSwap the provide will receive the address and amount we can make it reusable.

dangeross: It's not a simple move because it constructs the URL differently and with a different Swap. But yeah, I can extract the swap type, add optional params to the config and move it to sdk-common

cli/src/commands.rs
```diff
@@ -54,6 +54,13 @@ pub(crate) enum Command {
         /// Amount the payer will send, in satoshi
         payer_amount_sat: u64,
     },
+    /// Generates an URL to buy bitcoin from a 3rd party provider
+    BuyBitcoin {
+        provider: BuyBitcoinProvider,
```
ok300, line 59: Does the provider have to be specified? (i.e. can't we default to Moonpay for now?)

And if so, what's the correct way to specify it in the CLI?

ok300 (commented):
Do we check if the Moonpay min/max limits fit within the Receive Onchain limits from Boltz?

lib/core/src/buy.rs
```diff
@@ -0,0 +1,65 @@
...
+impl BuyBitcoinApi for BuyBitcoinService {
+    async fn buy_bitcoin(
+        &self,
+        provider: BuyBitcoinProvider,
+        chain_swap: &ChainSwap,
+        redirect_url: Option<String>,
+    ) -> Result<String> {
+        if self.config.network != LiquidNetwork::Mainnet {
+            return Err(anyhow!("Can only buy bitcoin on Mainnet"));
+        }
```
ok300, line 48: Looks like Moonpay supports Testnet too: https://dev.moonpay.com/docs/sandbox-testing

> Once you've hit the "Pay" button, all Cryptocurrency withdrawals are executed on test networks:
>...
> Testnet3 for Bitcoin

Would that be easy to integrate?

cli/src/commands.rs
dangeross: I wanted to keep the same approach as the Greenlight SDK, with the addition of providing the amount to buy. In the CLI we first call `prepare_buy_bitcoin()` to validate the amount limits and the amount is locked in the request to Moonpay.
```
breez-liquid-cli> buy-bitcoin moonpay 100000
```

lib/core/src/buy.rs
dangeross: I'll add that

roeierez (approved):
LGTM

lib/core/src/buy.rs
dangeross: @ok300 It's currently not possible to sign the Moonpay url using the staging breez server / sandbox secret. So I'll make an issue to support that

ok300: Sounds good, thanks.

ok300 (approved):

PR #399, opened 2024-07-17
Restore: Associate swap tx IDs from onchain data

### LINKED ISSUE #373: Associate claim, refund, lockup using onchain data
Currently for ongoing swaps we get the lockup/claim tx from boltz statuses and the refund we set once we broadcast it.
This doesn't allow us to restore these values when the user restores a wallet.
One way to solve this is that on restore we will scan the chain and make the right associations.
For every swap:
1. Get the lockup tx
2. Fetch the tx that spends the lockup and mark it as either claim or refund.

ok300:
This PR recovers the various swap tx IDs from
- initial immutable DB data
- onchain data (LWK)
- a single Electrum batch call to fetch script histories

Open points
- [x] Send swaps
- [x] Receive swaps
- [x] Chain swaps

Fixes #373


lib/core/src/sdk.rs
```diff
@@ -1431,6 +1440,118 @@ impl LiquidSdk {
             }
         }
 
+        // TODO Ensure this happens only on first sync
```
roeierez, line 1443: Perhaps it is better for now to put all logic of recovering chain data in its own function (recover_swaps_chain_data ?). I am thinking that because it might need to be reusable code and we still not sure how we should incorporate it.

lib/core/src/sdk.rs
```diff
@@ -1431,6 +1440,118 @@ impl LiquidSdk {
...
+        let possible_claim_or_refund_txs_by_input: HashMap<OutPoint, Txid> = tx_map
+            .iter()
+            .filter(|(_, tx)| tx.balance.values().sum::<i64>() > 0)
+            .filter(|(_, tx)| tx.tx.output.len() == 2) // Refund tx + fee
+            .filter(|(_, tx)| tx.tx.input.len() == 1)
+            .filter_map(|(_, wallet_tx)| {
+                wallet_tx
+                    .tx
+                    .input
+                    .first()
```
roeierez, line 1491: I agree it will work but if on future versions we will find a way to aggregate multiple refunds in one transaction (to save fees) it will break. So perhaps just map all inputs?

roeierez (commented):
Looks like the right direction to me.

ok300:
@roeierez I rewrote the PR to use the script histories fetched from Electrum, as discussed.

This also addressed all your points above.

So input data is now:
- immutable DB data (simulated, I fetch it from the current DB)
- onchain data (our onchain txs from LWK)
- script histories from 1 Electrum call (txs associated with every swap script)

lib/core/src/restore_onchain.rs
```diff
@@ -0,0 +1,334 @@
...
+            // TODO Persist updated swaps
+            // TODO     Send updates with found txids
+            // TODO     How to set tx IDs without having to also set state?
+            // self.send_swap_state_handler
+            //     .update_swap_info(&swap_id, Failed, None, None, Some(&refund_tx_id_str))
+            //     .await?;
+
+            Ok(())
+        }
+        pub(crate) async fn get_onchain_data(
```
roeierez, line 154: The whole process contains lots of steps and since there are multiple types of data mapping structures  and lots of intermediate steps it makes it hard to follow the code.

In my mind the process could be simplify by following these 4 steps:
1. Fetch wallet transactions, map them by tx id.
2. Fetch two kind of swap items from db (receive, send), map them by script (e.g  Map<script, (swap_id, type)>
3. Fetch script history for all swaps together
4. For each script history:
- Find lockup and claim and add the info to the swap items.

The code roughly does that but I think it can be simplified.
For example we have:
`receive_db: HashMap<String, (CreateReverseResponse, LBtcSwapScript, Script)>`
and 
```
let mut receive_composite_data: HashMap<
                String,
                (CreateReverseResponse, LBtcSwapScript, Script, &Vec<History>)
```
Perhaps it is better and simpler to create a struct with the composite fields ( CreateReverseResponse, LBtcSwapScript, Script, &Vec<History> )
and fill them.
Also instead of &Vec<History> can we populate in that struct directly the lockup & claim ? It will save another processing (mapping) later.

lib/core/src/restore_onchain.rs
```diff
@@ -0,0 +1,334 @@
...
+
+    use anyhow::Result;
+    use boltz_client::boltz::CreateReverseResponse;
+    use boltz_client::LBtcSwapScript;
+    use log::{error, info};
+    use lwk_wollet::elements::Script;
+
+    use crate::sdk::LiquidSdk;
+
+    pub(crate) struct ImmutableDb {
```
roeierez, line 14: NIT: Wouldn't use immuntable as it doesn't hint anytning about the data it contains. Perhaps just SwapsList ?

lib/core/src/restore_onchain.rs
```diff
@@ -0,0 +1,334 @@
...
+        pub(crate) send_db: HashMap<String, (LBtcSwapScript, Script)>,
+        pub(crate) receive_db: HashMap<String, (CreateReverseResponse, LBtcSwapScript, Script)>,
+    }
+
+    impl LiquidSdk {
+        pub(crate) async fn get_immutable_db_data(&self) -> Result<ImmutableDb> {
```
roeierez, line 20: NIT: get_swaps_list?

roeierez (changes requested):

lib/core/src/chain/liquid.rs
```diff
@@ -31,9 +31,12 @@ pub trait LiquidChainService: Send + Sync {
     /// Get a list of transactions
     async fn get_transactions(&self, txids: &[Txid]) -> Result<Vec<Transaction>>;
 
-    /// Get the transactions involved in a list of scripts including lowball
+    /// Get the transactions involved in a script, including lowball transactions
     async fn get_script_history(&self, scripts: &Script) -> Result<Vec<History>>;
 
+    /// Get the transactions involved in a list of scripts, including lowball transactions.
+    async fn get_scripts_history_electrum(&self, scripts: &[&Script]) -> Result<Vec<Vec<History>>>;
```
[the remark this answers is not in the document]
ok300: Done: 5eb753846f4a66ee93f4c44fd1b3d78f4ba72c19

lib/core/src/restore_onchain.rs
```diff
@@ -0,0 +1,334 @@
+/// Methods to simulate the immutable DB data available from real-time sync
+// TODO Remove once real-time sync is integrated
+pub(crate) mod immutable {
```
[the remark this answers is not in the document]
ok300: Done: 73dfa71e84e7b0f0dd5c08198ffa78ccf52c38e2

lib/core/src/restore_onchain.rs
```diff
@@ -0,0 +1,334 @@
...
+            Ok(ImmutableDb {
+                send_db: send_swap_immutable_db,
+                receive_db: receive_swap_immutable_db,
+            })
+        }
+    }
+}
+
+/// Methods to restore the swap tx IDs from the onchain data
+pub(crate) mod onchain {
```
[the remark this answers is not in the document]
ok300: Done, flattened it in `restore.rs`: 73dfa71e84e7b0f0dd5c08198ffa78ccf52c38e2

lib/core/src/restore.rs
```diff
@@ -0,0 +1,820 @@
...
+                .lbtc_lockup_script_history
+                .iter()
+                .find(|tx_id| tx_map.incoming_tx_map.contains_key::<Txid>(tx_id));
+            if let Some(refund_tx_id) = maybe_refund_tx_id {
+                lbtc_refund_tx_map.insert(swap_id.clone(), *refund_tx_id);
+            }
+
+            match history.btc_claim_script_history.len() {
+                2 => {
+                    // TODO How to tell the claim tx apart from the lockup tx? Is the order in which they're received from Electrum reliable?
```
roeierez, line 232: We might need to fetch these transactions in order to do it reliably.

lib/core/src/restore.rs
```diff
@@ -0,0 +1,820 @@
...
+            lbtc_server_lockup_claim_tx_ids,
+            btc_user_lockup_tx_ids,
+            btc_refund_tx_ids,
+        })
+    }
+}
+
+/// Methods to simulate the immutable DB data available from real-time sync
+// TODO Remove once real-time sync is integrated
+pub(crate) mod immutable {
```
roeierez, line 335: NIT: I don't know if we need another module here. Also immutable is too generic in this context IMO.

lib/core/src/restore.rs
```diff
@@ -0,0 +1,820 @@
...
+
+    type BtcScript = lwk_wollet::bitcoin::ScriptBuf;
+    type LBtcScript = lwk_wollet::elements::Script;
+
+    pub(crate) type SendSwapHistory = Vec<Txid>;
+    pub(crate) type ReceiveSwapHistory = Vec<Txid>;
+
+    #[allow(dead_code)]
+    #[derive(Clone)]
+    struct SendSwapImmutableData {
```
roeierez, line 354: NIT: also here Immutable seems too generic. Basically all we have here is the swap_script (event he script is derive from it). So perhapas we can rename to SendSwapScriptsData ?

lib/core/src/restore.rs
```diff
@@ -0,0 +1,820 @@
...
+pub(crate) struct RecoveredOnchainData {
+    send: RecoveredOnchainDataSend,
+    receive: RecoveredOnchainDataReceive,
+    chain_send: RecoveredOnchainDataChainSend,
+    chain_receive: RecoveredOnchainDataChainReceive,
+}
+
+impl LiquidSdk {
+    pub(crate) async fn recover_from_onchain(&self, tx_map: TxMap) -> Result<()> {
+        let immutable_db = self.get_swaps_list().await?;
```
roeierez, line 67: NIT: instead of immutable_db perhaps just `swaps_scripts_list`? or `swaps_list`?

roeierez: Also it is better to call this function out of the recover_from_onchain since I can see later that we will need to invoke it on a list that comes from realtime sync (and doesn't exist in the db).

lib/core/src/restore.rs
```diff
@@ -0,0 +1,820 @@
...
+
+        let _recovered = self.get_onchain_data(tx_map, immutable_db).await?;
+
+        // TODO Persist updated swaps
+        // TODO     Send updates with found txids
+        // TODO     How to set tx IDs without having to also set state?
```
roeierez, line 73: Why not setting the state? Does anything prevent us from calculating the state based on the onchain data?

roeierez (changes requested):
Looks a lot better. Added some comments.

lib/core/src/restore.rs
```diff
@@ -0,0 +1,820 @@
...
+        info!(
+            "[Recover Send] Found {} lockup txs from onchain data",
+            lockup_tx_map.len()
+        );
+        info!(
+            "[Recover Send] Found {} refund txs from onchain data",
+            refund_tx_map.len()
+        );
+
+        Ok(RecoveredOnchainDataSend {
```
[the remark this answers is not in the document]
ok300: As discussed, it makes more sense to have the preimage in the immutable data. It can only be retrieved for some states (non-cooperative claim tx), in other cases its only known from interactions with Boltz. So for consistency, it makes more sense to store it in the immutable data, when available (e.g. after claim is sucessful).

lib/core/src/restore.rs
```diff
@@ -0,0 +1,820 @@
...
+    async fn recover_receive_swap_tx_ids(
+        &self,
+        tx_map: &TxMap,
+        receive_histories_by_swap_id: HashMap<String, ReceiveSwapHistory>,
+    ) -> Result<RecoveredOnchainDataReceive> {
+        let mut lockup_claim_tx_ids_map: HashMap<String, (Txid, Txid)> = HashMap::new();
+
+        for (swap_id, history) in receive_histories_by_swap_id {
+            match history.len() {
+                2 => {
```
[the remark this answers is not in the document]
ok300: I switch to that logic, ignoring everything that is above 2 txs (address re-use).

lib/core/src/restore.rs
ok300: Done, added a call to fetch the BTC txs so we can look at the vouts / vins.

lib/core/src/restore.rs
ok300: As discussed, the state is now derived as a "partial state", based on the immutable data and the onchain data.

lib/core/src/restore_onchain.rs
ok300: Done, added custom types and simplified the various types used.

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1164 @@
...
+    fn get_partial_state(&self) -> PaymentState {
+        match (&self.lockup_tx_id, &self.claim_tx_id) {
+            (Some(_), Some(claim_tx_id)) => match claim_tx_id.confirmed {
+                true => PaymentState::Complete,
+                false => PaymentState::Pending,
+            },
+            (Some(_), None) => PaymentState::Pending,
+            // TODO How to distinguish between Failed and Created (if in both cases, no lockup or claim tx present)
+            //   See https://docs.boltz.exchange/v/api/lifecycle#reverse-submarine-swaps
+            _ => PaymentState::Failed,
```
roeierez: If no lockup or claim present then it means the payer didn't start paying (swapper didn't broadcast the lockup) from our perspective it is always Created (even on normal flow without restore).

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1164 @@
...
+                Some(_) => PaymentState::Complete,
+                None => match &self.lbtc_refund_tx_id {
+                    Some(tx) => match tx.confirmed {
+                        true => PaymentState::Failed,
+                        false => PaymentState::RefundPending,
+                    },
+                    None => PaymentState::Created,
+                },
+            },
+            None => PaymentState::Created,
```
roeierez, line 105: Same as with send swap this should always be Timeout as on send we automatically attempt to pay and never leave the swap in Created state.
We also ensure the Created ones are converted to Timeout on startup: https://github.com/breez/breez-sdk-liquid/blob/main/lib/core/src/sdk.rs#L184

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1164 @@
...
+pub(crate) struct RecoveredOnchainDataChainSend {
+    lbtc_user_lockup_tx_id: Option<HistoryTxId>,
+    lbtc_refund_tx_id: Option<HistoryTxId>,
+    btc_server_lockup_tx_id: Option<HistoryTxId>,
+    btc_claim_tx_id: Option<HistoryTxId>,
+}
+impl PartialSwapState for RecoveredOnchainDataChainSend {
+    fn get_partial_state(&self) -> PaymentState {
+        // TODO How to detect TimedOut state?
+        //     TimedOut: This covers the case when the swap state is still Created and the swap fails to reach the
```
roeierez, line 92: See comment bellow

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1164 @@
...
+        //     Pending state in time. The TimedOut state indicates the lockup tx should never be broadcast.
+        match &self.lbtc_user_lockup_tx_id {
+            Some(_) => match &self.btc_claim_tx_id {
+                Some(_) => PaymentState::Complete,
```
roeierez, line 96: Shouldn't we check if the btc_claim_tx is confirmed or not? If it is still not confirmed should the payment be still in Pending state?

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1164 @@
...
+pub(crate) struct RecoveredOnchainDataChainReceive {
+    lbtc_server_lockup_tx_id: Option<HistoryTxId>,
+    lbtc_server_claim_tx_id: Option<HistoryTxId>,
+    btc_user_lockup_tx_id: Option<HistoryTxId>,
+    btc_refund_tx_id: Option<HistoryTxId>,
+}
+impl PartialSwapState for RecoveredOnchainDataChainReceive {
+    fn get_partial_state(&self) -> PaymentState {
+        match &self.btc_user_lockup_tx_id {
+            Some(_) => match &self.lbtc_server_claim_tx_id {
```
roeierez, line 119: Do we care about the server claim? As far as I understand it is the last step in the swap which we don't care about. The swap completes when we claim If I get it right.

roeierez (changes requested):
Went over the state restoration, now will go over the other parts.

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1172 @@
...
+    /// For each swap, recovers data from chain services.
+    ///
+    /// The returned data include txs and the partial swap state. This is a partial state, because
+    /// certain swap states cannot be determined based on initial and onchain data (e.g. [PaymentState::TimedOut])
+    ///
+    /// The caller is expected to merge this data with any other data available, then persist the
+    /// reconstructed swap.
+    pub(crate) async fn recover_from_onchain(&self, tx_map: TxMap) -> Result<RecoveredOnchainData> {
+        // Immutable DB, as fetched from endpoint
+        let imm_db = self.get_swaps_list().await?;
```
roeierez, line 151: Ideally this should be given as parameter to this function so we can use it on swaps that are synced in from outside.

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1172 @@
...
+                    let exp_state = expected.state;
+                    let rec_state = recovered.get_partial_state();
+                    info!("state: {exp_state:?} / {rec_state:?}");
+                }
+                (Some(_), None) => error!("No recovered data for Send Swap {send_swap_id}"),
+                (None, Some(_)) => error!("No expected data for Send Swap {send_swap_id}"),
+                (None, None) => error!("Unknown Send Swap {send_swap_id}"),
+            }
+            info!("[Restore Send] Validated {send_swap_id}");
+        }
```
roeierez, line 186: What is the purpose of this validation step? If everything matches what we already have then why fetch on chain data from the first place? The idea is that we won't have these swaps in the db so we will have nothing to compare to.
Sames feedback goes for the validation step of other swap types.

roeierez (changes requested):

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1172 @@
...
+    /// Reconstruct Receive Swap tx IDs from the onchain data and the immutable DB data
+    async fn recover_receive_swap_tx_ids(
+        &self,
+        tx_map: &TxMap,
+        receive_histories_by_swap_id: HashMap<String, ReceiveSwapHistory>,
+    ) -> Result<HashMap<String, RecoveredOnchainDataReceive>> {
+        let mut res: HashMap<String, RecoveredOnchainDataReceive> = HashMap::new();
+        for (swap_id, history) in receive_histories_by_swap_id {
+            let (lockup_tx_id, claim_tx_id) = match history.len() {
+                2 => {
```
roeierez, line 408: We should consider here <=2 as there might be only lockup but no claim.

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1172 @@
...
+                    let first = history[0].clone();
+                    let second = history[1].clone();
+
+                    // If a history tx is a known incoming tx, it's the claim tx
+                    match tx_map.incoming_tx_map.contains_key::<Txid>(&first.txid) {
```
roeierez, line 413: We are missing here a case of refund to the swapper. Case can be that swapper initiated a lockup (first tx), we didn't claim and then swapper refunded itself (in that case the tx doesn't exist on our map).

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1172 @@
...
+        &self,
+        tx_map: &TxMap,
+        chain_receive_histories_by_swap_id: HashMap<String, ReceiveChainSwapHistory>,
+        receive_chain_swap_immutable_db_by_swap_id: &HashMap<String, ReceiveChainSwapImmutableData>,
+    ) -> Result<HashMap<String, RecoveredOnchainDataChainReceive>> {
+        let mut res: HashMap<String, RecoveredOnchainDataChainReceive> = HashMap::new();
+        for (swap_id, history) in chain_receive_histories_by_swap_id {
+            info!("[Recover Chain Receive] Checking swap {swap_id}");
+
+            let (lbtc_server_lockup_tx_id, lbtc_server_claim_tx_id) = match history
```
roeierez, line 525: It is the server lockup and the user claim, isn't it? Shouldn't we rename `lbtc_server_claim_tx_id` to `lbtc_user_claim_tx_id`

lib/core/src/restore.rs
roeierez: After reading all the code I now see it is not the server claim but is our claim.
So perhaps rename to `lbtc_user_claim_tx_id` ?

roeierez (changes requested):

lib/core/src/restore.rs
ok300: No lockup or claim tx can mean both Created (no payment happened yet) or Failed (`invoice.expired`, `swap.expired`, or even `transaction.failed`).

We default to Created now, but I left this comment to point out that we lose this Boltz-specific status information.

roeierez: I see. Wasn't aware that we set it to Failed in these cases. I am not sure BTW we need to set it to Failed in case of receiving but this is another story out of the scope of this issue.

lib/core/src/restore.rs
ok300: Done: 07759523cb6f57ecee90eaf72a4d9dee591a21ab

lib/core/src/restore.rs
ok300: Done: 07759523cb6f57ecee90eaf72a4d9dee591a21ab

lib/core/src/restore.rs
ok300: I followed the existing state handling logic, where we mark it as Complete as soon as we broadcast the claim:

https://github.com/breez/breez-sdk-liquid/blob/dc0e347e2b1604e8187f30eedc10486323522b91/lib/core/src/chain_swap.rs#L580-L603

lib/core/src/restore.rs
ok300: > After reading all the code I now see it is not the server claim but is our claim. So perhaps rename to `lbtc_user_claim_tx_id` ?

I checked with @dangeross and this is the server claim as per Boltz nomenclature. For them, the initiator (not creator) of the swap is the "user", and txs associated with those funds are user txs. In this case, that's the BTC lockup tx, BTC claim tx and even BTC refund tx.

Conversely, the funds locked by the swapper ("server") are associated with server txs. In this case, that's the LBTC lockup tx and LBTC claim tx (the tx you referred to).

ok300: > Do we care about the server claim? As far as I understand it is the last step in the swap which we don't care about. The swap completes when we claim If I get it right.

The server claim is our claim, that's the only reason we check this.

lib/core/src/restore.rs
roeierez: I see, indeed it is consistent. Question is do we have an issue with chain swap then? @dangeross shouldn't we make the chain swap completed only after our claim tx i s confirmed?

dangeross: It is currently marking the swap complete for both directions when the claim is broadcast. 

Perhaps for the incoming swap we can let the sync mark it complete. For the outgoing swap we are currently not monitoring the BTC address for confirmed transactions (but we could). What would you suggest @roeierez?

lib/core/src/restore.rs
ok300: As discussed in the group chat, this is correct as it is, so marking this as resolved.

lib/core/src/restore.rs
ok300: Done: c7575afbf96dcf34f1cc40026c7b21241a31c865

lib/core/src/restore.rs
roeierez: @dangeross  For send I think we should especially because it is BTC mainnet where transactions may not confirm or go out of the mempool.
Also users see the swap as completed but still don't see the BTC on the destination address.
For receive, you are right the sync can do that same way we do for receive_swap.
But let's not decouple it with this PR. Once we fix it in a following PR we will fix here as well.

lib/core/src/restore.rs
ok300: Done: 31ebefe037be9d20133b314a2167b04732b00d2f

lib/core/src/restore.rs
ok300: As discussed elsewhere, they are both "server txs", as they both act on the swapper-locked funds.

lib/core/src/restore.rs
ok300: Done: da39009268bde34d26b81a505873e8f028acaf0b

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1164 @@
...
+                .partition(|(_, tx)| tx.balance.values().sum::<i64>() < 0);
+
+        Self {
+            outgoing_tx_map,
+            incoming_tx_map,
+        }
+    }
+}
+
+trait PartialSwapState {
```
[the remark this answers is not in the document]
ok300: > Why do we need a trait here compare to only a function for each swap?

I did it that way so the function `get_partial_state` can have the docs in one place, applicable to all instances. IMO it's important to point out why its a partial state and what it means.

> Also IMO it would be more clear to call the function `restore_state()`

Restore implies it's storing it somewhere, which it doesn't. I renamed it to `derive_partial_state`:  9566f7b87390b824b1e2015221b22156eb598054

lib/core/src/restore.rs
ok300: Done: da39009268bde34d26b81a505873e8f028acaf0b

lib/core/src/restore.rs
ok300: I originally started with that, but later realized these structs are inputs from the Immutable DB, representing the immutable data for the swaps. My understanding is that immutable data will be read into such structs, bringing everything that's available (create response, keypairs, etc). Restore is then supposed to use them as inputs.

From this struct, it happens I'm only using the script data for restoring. I named it `SendSwapImmutableData` in the expectation it will include the rest of the immutable data as well.

lib/core/src/restore.rs
ok300: This module (marked as `TODO Remove once real-time sync is integrated`) is a placeholder for the immutable DB interface.

It contains ways to "simulate" the immutable data (now deriving it from actual DB data) until the interface is available. I placed these helper methods in this module, so it's easy to remove them when the interface is ready.

lib/core/src/restore.rs
ok300: These validations were a way for me to check if the correct tx IDs and states are restored for various swap types.

I removed them: a0c62b802338a258fe48784fffa05b84e402ae3d

I will add real integration tests in a follow-up.

lib/core/src/restore.rs
```diff
@@ -0,0 +1,1055 @@
...
+            // The exact type of the second is found in the next step.
+            let (btc_user_lockup_tx_id, btc_second_tx_id) = match history
+                .btc_lockup_script_history
+                .len()
+            {
+                // Only lockup tx available
+                1 => (Some(history.btc_lockup_script_history[0].clone()), None),
+
+                // Both txs available (lockup + claim, or lockup + refund)
+                // Any tx above the first two, we ignore, as that is address re-use which is not supported
```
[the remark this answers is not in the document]
ok300: Which lines do you mean? In L436-462 it's already checking for `n >= 2` txs and ignores the ones after the 2nd.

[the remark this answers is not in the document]
dangeross: Ah yeah right, I missed that sorry

lib/core/src/restore.rs
ok300: The chain swap change was done in https://github.com/breez/breez-sdk-liquid/pull/450

The corresponding change to this PR was done in 97d62c5792b7b36da2e99f2926264116ccd59157

dangeross (approved):

dangeross:
Just needs conflicts resolving

PR #400, opened 2024-07-18
Persistence: consolidate state filter

ok300:
This PR extracts `get_where_clause_state_in` and re-uses it to build `WHERE` clauses that filter by state.


dangeross (approved):
LGTM

PR #401, opened 2024-07-18
Remove `sync` call when handling Boltz update

ok300:
A `sync` call can take 2-3s.

IMO we don't need a `sync` when handling Boltz updates. For relevant new txs, we already add pseudo-txs.


roeierez (approved):
LGMT! I agree that if we ever need to sync is should be placed under the specific update that needs it and not as a generic mechanism for all updates.

PR #407, opened 2024-07-23
Send & Chain: Localize refund flow to swap handlers

hydra-yse:
This PR aims at fixing the refund flow loop, since it used to block the status stream thread while waiting for lockups to be confirmed. The logic has been moved to the swap handlers, so they are now independent from the swapper (currently only used for broadcasting). The logic for constructing the refund tx (only Liquid) has also been moved to the SDK side due to problems with lowball, and the requirement to fetch from our Esplora. 

The current improvements make cooperative refunds happen almost instantaneously, and with no reliance on the swapper. In case of no Boltz status update, they will be refunded non-cooperatively on locktime expiry (via `LiquidSdk::track_pending_swaps`).


lib/core/src/sdk.rs
```diff
@@ -344,67 +344,117 @@ impl LiquidSdk {
...
+    /// For outgoing chain swaps, we check whether the status is `RefundPending` and
+    /// the locktime has expired. If so, we refund non-cooperatively; else we refund cooperatively
+    ///
+    /// ## Incoming Chain Swap
+    /// For incoming chain swaps, we check whether the status is `Pending` and the locktime has expired.
+    /// If so, we set the swap status as `Refundable`
+    async fn try_chain_swap_refund(&self, chain_swap: &ChainSwap) -> Result<()> {
+        if chain_swap.user_lockup_tx_id.is_none() || chain_swap.refund_tx_id.is_some() {
+            return Err(anyhow::anyhow!(
+                "Lockup_tx not present or refund_tx already broadcast"
```
dangeross, line 358: I think this will produce errors in the logs while the refund tx is unconfirmed. This could be for a while for a Bitcoin tx specially if a low fee rate is selected

dangeross (commented):
I think by not explicitly checking the swap expiration and first relying on the swap state you are assuming Boltz is always being compliant (or functional). If swap a state is not changed to RefundPending (for outgoing chain/send swaps) in the stream updates, then there is never an attempt to refund

roeierez:
@hydra-yse IMO the coopertive refund which should in theory always succeed should be executed immeditely as soon as we understand we need it so having it as part of the status stream as before makes sense to me.
I do agree that making the non cooperative refund only in one place (the watcher task) is a good idea since anyway we will execute it only on exiration and we are not relying on boltz statuses for that but only on chain information.

lib/core/src/chain_swap.rs
```diff
@@ -465,18 +478,9 @@ impl ChainSwapStateHandler {
...
-                                self.update_swap_info(
-                                    id,
-                                    RefundPending,
-                                    None,
-                                    None,
-                                    None,
-                                    Some(&refund_tx_id),
-                                )
-                                .await?;
+                                warn!("Chain Swap {id} user lockup tx has been broadcast.");
```
roeierez, line 481: I think we should try here cooperative refund which should work for 99% only if fails signal to the non cooperative refund. Also cooperative refund should happen immediately for example when trying to pay a not online peer, the swapper will fail and then we need to refund immediately. Let's also make sure we test that as part of this PR.
Also why use lock and im memory slice? Isn't it better to just mark these in the db as failed to refund coopeartively and fetch them using a db query when needed?

roeierez:
The code as is is not clear and hard to understand (not because of this PR) so I certainly understand the motivation to do some cleanup here.
IMO a clean approach would be:
1. Create a new module (file) refund.rs
2. In there we will have two structures (SendSwapRefunder, ChainSwapRefunder) 
3. Each of them has two functions:
* Refund(swaps: []swap) - will attempt first cooperatively and if fail and timeout is applicable will try non cooperatively.
* RefundExpired() - which will query the db for those that are refundables and use the former method to refund.

Now our two cases can use the above functinoality:
1. From the status stream when need to refund just call `Refund (swaps)`.
2. From a background thread that runs periodically and only calls `RefundExpired()`.

What do you think @hydra-yse ?

roeierez (changes requested):

hydra-yse:
There are still some checks/tests to do, so I will be leaving it as draft for now, though I would appreciate a review on the interface changes :smile: . @roeierez I'll clarify later today why I think creating a separate `refund.rs` sounds good but has some practical issues, and why I decided to opt for this approach instead.

hydra-yse:
@dangeross I've also addressed Pending swap tracking in 9113ce1, so that even if Boltz doesn't return the right state we can still try non-cooperatively after the expiry. Just have to propagate the changes to chain swaps.

lib/core/src/persist/chain.rs
```diff
@@ -199,6 +199,15 @@ impl Persister {
         )
     }
 
+    pub(crate) fn list_pending_refund_chain_swaps(&self) -> Result<Vec<ChainSwap>> {
+        let con: Connection = self.get_connection()?;
+        let where_clause = vec![
+            format!("state = {}", PaymentState::RefundPending as i8),
+            "refund_tx_id IS NULL".to_string(),
```
roeierez, line 206: For incoming chain swaps we still need to return refundable as long as the refund_tx is not confirmed.
I don't know if the refund_tx_id is there also when not confirmed yet.

roeierez (changes requested):
It looks a lot better IMO. I also like the fact that every module (chainswap, send_swap) takes care of its own refunds.
I still need to do another pass.

lib/core/src/chain_swap.rs
```diff
@@ -534,22 +541,26 @@ impl ChainSwapStateHandler {
...
-                                self.update_swap_info(
-                                    id,
-                                    RefundPending,
-                                    None,
-                                    None,
-                                    None,
-                                    Some(&refund_tx_id),
-                                )
-                                .await?;
+                        match swap.user_lockup_tx_id {
```
hydra-yse, line 544: Just a comment for @dangeross, I've reverted the script history check here to avoid re-fetching. It occurs anyway when reconstructing the refund transaction (`utils::new_lbtc_refund_tx`) so there shouldn't be any problems there.

roeierez:
@hydra-yse is this ready for another review? If so can you please rebase the conflicts?

lib/core/Cargo.toml
```diff
@@ -14,7 +14,7 @@ frb = ["dep:flutter_rust_bridge"]
 [dependencies]
 anyhow = { workspace = true }
 bip39 = "2.0.0"
-boltz-client = { git = "https://github.com/SatoshiPortal/boltz-rust", rev = "c140193dab075093e1cfdcc1dd608be8e828d1ef" }
+boltz-client = { git = "https://github.com/hydra-yse/boltz-rust", branch = "yse-breez-latest" }
```
ok300, line 17: Is there a PR to propagate our fixes upstream, so we can that repo directly and avoid using a custom branch for the Boltz client?

lib/core/src/utils.rs
```diff
@@ -86,56 +86,74 @@ pub(crate) async fn derive_fee_rate_sats_per_kvb(
...
+
+    let genesis_hash = BlockHash::from_str(match config.network {
+        LiquidNetwork::Mainnet => {
+            "1466275836220db2944ca059a3a10ef6fd2ea684b0688d2c379296888a206003"
+        }
+        LiquidNetwork::Testnet => {
+            "67d5eb1aee63c6c2058a088985503ff0626fd3f7f8022bdc74fab36a359164db"
+        }
+    })
+    .expect("Expecting valid block hash");
```
ok300, line 123: Would avoid `.expect()` as it causes a `panic`. Would throw and propagate an error instead.

lib/core/Cargo.toml
hydra-yse: Just created it: https://github.com/SatoshiPortal/boltz-rust/pull/71

ok300: Nice! It already got merged, can you update the PR dependency so it uses the latest upstream commit?

ok300: Nevermind, I see there are more changes from the boltz client that would make this PR even bigger.

I created a separate PR to handle this: https://github.com/breez/breez-sdk-liquid/pull/486

lib/core/src/refund/chain.rs
```diff
@@ -0,0 +1,273 @@
...
+            self.boltz_url().to_string(),
+            swap.id.clone(),
+        )?;
+        let broadcast_fees_sat = common::estimate_btc_refund_fees_sat(
+            &refund_tx,
+            &refund_keypair,
+            &Preimage::from_str(&swap.preimage)?,
+            fee_rate_msats_per_vbyte as f64 / 1000.0,
+        )?;
+        let signed_tx = refund_tx.sign_refund(&refund_keypair, broadcast_fees_sat, cooperative)?;
```
roeierez, line 113: The feedback here is also for other refund handlers:
The attempt to put the refund logic outside the swapper enforces us to expose the swapper internals (bolt client, bitcoin electrum config, etc...) and it makes the abstraction useless.
Since the cooperative refund needs the swapper interaction it makes sense to have it inside the swapper IMO.
As for the non cooperative refund I have not objection of having it outside the swapper (if it doesn't expose the internals) but if the cooperative will move there perhaps it would be more simple to have the non cooperative there as well.
The only thing IMO that should be outside the swapper is the broadcast itself.

roeierez (changes requested):
The handlers themselves looks good IMO. I added feedback about the refund handlers which IMO needs some restructure.

lib/core/src/refund/chain.rs
```diff
@@ -0,0 +1,273 @@
...
+};
+
+use super::{
+    common::{self, get_lbtc_fee_rate_sats_per_vbyte},
+    RefundHandler,
+};
+
+pub(crate) struct ChainRefundHandler<SwapperType: ?Sized> {
+    config: Config,
+    swapper: Option<Arc<SwapperType>>,
```
ok300, line 27: Why is this an `Option` if it's always set when initializing this in `new`?

dangeross (commented):
Review still in progress...

lib/core/src/chain_swap.rs
```diff
@@ -56,18 +60,19 @@ impl ChainSwapStateHandler {
...
         })
     }
 
     pub(crate) async fn start(self: Arc<Self>, mut shutdown: watch::Receiver<()>) {
         let cloned = self.clone();
         tokio::spawn(async move {
-            let mut interval = tokio::time::interval(Duration::from_secs(60 * 10));
-            interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
+            let mut rescan_interval = tokio::time::interval(Duration::from_secs(60 * 10));
+            rescan_interval.set_missed_tick_behavior(MissedTickBehavior::Burst);
```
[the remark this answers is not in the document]
hydra-yse: You're right, will revert back. The Burst was necessary because of a previous attempt at isolating the logic within the SwapHandler.

lib/core/src/refund/chain.rs
hydra-yse: The idea is it won't be always set, since the refunder shouldn't always interact with a swapper. We can do this by adding an implementation of ChainRefundHandler<()>, which would then set the swapper field to `None`.

lib/core/src/refund/chain.rs
```diff
@@ -0,0 +1,273 @@
...
+            .unwrap_or(match self.config.network {
+                LiquidNetwork::Mainnet => BOLTZ_MAINNET_URL_V2,
+                LiquidNetwork::Testnet => BOLTZ_TESTNET_URL_V2,
+            })
+    }
+
+    fn bitcoin_electrum_config(&self) -> ElectrumConfig {
+        self.swapper
+            .as_ref()
+            .map(|swapper| swapper.bitcoin_electrum_config().clone())
```
ok300, line 60: If this is to be used in a refund handler without a swapper, then maybe it makes sense to de-couple `bitcoin_electrum_config()` from the swapper and perhaps expose it as a function of the main `Config` instead.

Otherwise we have the initialization code duplicated: once in the swapper, once here (as fallback if no swapper present). IMO this could be consolidated.

lib/core/src/refund/chain.rs
```diff
@@ -0,0 +1,273 @@
...
+    ) -> Self {
+        Self {
+            config,
+            swapper: Some(swapper),
+            liquid_chain_service,
+            bitcoin_chain_service,
+        }
+    }
+
+    fn boltz_url(&self) -> &str {
```
ok300, line 47: Given [your answer above](https://github.com/breez/breez-sdk-liquid/pull/407/files#r1758877086), `boltz_url` should IMO either return the swapper value or throw an error.

For `boltz_url`, the swapper includes logic that builds a special `boltz_url` with the referral ID as GET argument. See `BoltzSwapper::new` and #329. By building a separate `boltz_url` here (outside of the swapper) and bypassing the swapper logic, that `boltz_url` will not have the referral ID.

It also makes sense intuitively: if no swapper (Boltz API) is used for this instance of the refund handler, then there is no need for a `boltz_url`.

lib/core/src/refund/chain.rs
```diff
@@ -0,0 +1,314 @@
...
+    ) -> anyhow::Result<u64> {
+        let refund_tx_size = refund_tx.size(refund_keypair, preimage).map_err(|e| {
+            anyhow::anyhow!(
+                "Could not calculate refund tx size when estimating broadcast fees: {e:?}"
+            )
+        })?;
+        Ok((refund_tx_size as f64 * fee_rate_sats_per_vbyte).ceil() as u64)
+    }
+
+    async fn new_btc_refund_tx(
```
ok300, line 61: Are we sure we want to build the refund txs ourselves?

I'm asking because, as discussed in the previous call, I'm preparing a PR for boltz-client where the created refund tx would consider all utxos, instead of just one as it does now. With this change, I'd have to duplicate those fixes here.

On a more general note: if we duplicate the refund tx building logic, we will always be behind `boltz_client` and their fixes / changes. We will have to regularly copy-paste their fixes into our version of the tx building.

Either way is fine with me, but not duplicating this logic seems more elegant and easier for us to maintain in the long term.

lib/core/src/chain_swap.rs
```diff
@@ -681,148 +723,298 @@ impl ChainSwapStateHandler {
...
+        for swap in pending_swaps {
+            if swap.refund_tx_id.is_some() {
+                continue;
+            }
+
+            match swap.direction {
+                // Track refunds
+                Direction::Outgoing => {
+                    let refund_tx_id_result: Result<String, PaymentError> = match swap.state {
+                        Pending => self.refund_outgoing_swap(&swap, false).await,
```
roeierez, line 962: Seems like we refund any pending outgoing? without checking expiry?

lib/core/src/chain_swap.rs
```diff
@@ -681,148 +723,298 @@ impl ChainSwapStateHandler {
...
-            Ok(res) => Ok(res),
-            Err(e) => {
-                warn!("Cooperative refund failed: {:?}", e);
-                let current_height = self.bitcoin_chain_service.lock().await.tip()?.height as u32;
-                self.swapper.refund_chain_swap_non_cooperative(
-                    &swap,
-                    broadcast_fees_sat,
-                    output_address,
-                    current_height,
+        ensure_sdk!(
```
roeierez, line 793: Shouldn't we also ensure the swap is refundable? to prevent broadcasting an invalid transaction?

roeierez (changes requested):
Looks good!

lib/core/src/chain_swap.rs
hydra-yse: The expiry check happens within `refund_outgoing_swap`. The flow is as follows:

- Pending state: We refund non-cooperatively, which will probably fail the expiry check and return (unless it's been there past the expiry, then it will refund)

- Refund Pending: We first refund cooperatively, then non-cooperatively. In the second case we check for expiry.

lib/core/src/chain_swap.rs
hydra-yse: Agreed

lib/core/src/chain_swap.rs
roeierez: While the check inside refund_outgoing_swap is in its place IMO, I think it is a bit misleading to call refund_outgoing_swap for every pending (for RefundPending it seems OK).
I think we should check for expiry before calling the non cooperative refund on pending (or fetch only the expired ones from the db).

roeierez: Also note that the refund_outgoing_swap throws error in case it is not expired so we will get unnecessary errors in the logs.

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -306,6 +306,7 @@ dictionary Config {
     u64 payment_timeout_sec;
     u32 zero_conf_min_fee_rate_msat;
     u64? zero_conf_max_amount_sat;
+    string? breez_api_key;
```
ok300, line 309: Is this really optional? Seems to me broadcasting certain txs won't work if the user won't explicitly set an API key.

Maybe `connect` should fail if no `Config::breez_api_key` is set?

hydra-yse: Btw, will make a commit to #489 as that's where I'm tracking those changes

ok300 (approved):

roeierez (approved):
LGTM

PR #409, opened 2024-07-23
Update boltz-rust and add receive payment description

### LINKED ISSUE #194: Switch from own `boltz-client` fork to original library
We're now using a fork of `boltz-client` because we needed a few fixes to test our flows:

https://github.com/breez/breez-liquid-sdk/blob/c67f4ac2d7694c299a8bb1196df09ebd8c53cbaa/lib/core/Cargo.toml#L14

Let's open PRs to push our fixes to upstream and switch to using the original `boltz-client` directly.

Using the upstream crate directly will benefit us too, as it's getting regular fixes and updates.

### LINKED ISSUE #403: Support custom invoice description on Receive
It appears the Boltz API supports setting a `description` field: https://api.boltz.exchange/swagger#/Reverse/post_swap_reverse

dangeross:
This PR updates the boltz-rust to the upstream latest and by doing so includes the option to add a payment description.

Fixes #403, fixes #194


ok300:
> This PR updates the boltz-rust to the upstream latest

Does this mean #194 is fixed as well?

dangeross:
> Does this mean #194 is fixed as well?

@ok300 yeah, fixes in the branch we're already added by i5hi

ok300 (approved):
Looks good.

@hydra-yse are your `boltz-rust` fixes also now merged upstream? If so, this indeed closes #194 too.

roeierez:
@dangeross looks good. Do you think it makes sense now to add "description" field to the payment struct and extract it from the bolt11 (same as we do in the greenlight sdk) ?

dangeross:
> @dangeross looks good. Do you think it makes sense now to add "description" field to the payment struct and extract it from the bolt11 (same as we do in the greenlight sdk) ?

@roeierez I can add that for the lightning swaps. The chain swaps don't have a description AFAIK, so we can make it an optional field

roeierez:
> > @dangeross looks good. Do you think it makes sense now to add "description" field to the payment struct and extract it from the bolt11 (same as we do in the greenlight sdk) ?
> 
> @roeierez I can add that for the lightning swaps. The chain swaps don't have a description AFAIK, so we can make it an optional field

Let's do it the same way we handle it in greenlight sdk. I think we add a constant text there "Bitcoin Transfer"

roeierez (approved):

PR #410, opened 2024-07-23
Integrate webhooks

dangeross:
This PR:
- Adds `register_webhook` and `unregister_webhook` methods
- Includes the webhook in the create swap requests

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -13,7 +13,6 @@ use crate::model::*;
 use crate::{get_invoice_description, utils};
 use anyhow::{anyhow, Result};
 use migrations::current_migrations;
-use openssl::sha::Sha256;
 use rusqlite::{params, Connection, OptionalExtension, Row};
 use rusqlite_migration::{Migrations, M};
 
@@ -436,13 +435,6 @@ fn filter_to_where_clause(
     }
 
     where_clause.join(" and ")
-}
-
-fn hash_str(s: &str) -> String {
-    let mut hasher = Sha256::new();
-    hasher.update(s.as_bytes());
-
-    hex::encode(hasher.finish())
 }
 
 #[cfg(test)]
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -13,6 +13,7 @@ use crate::model::*;
 use crate::{get_invoice_description, utils};
 use anyhow::{anyhow, Result};
 use migrations::current_migrations;
+use openssl::sha::Sha256;
 use rusqlite::{params, Connection, OptionalExtension, Row};
 use rusqlite_migration::{Migrations, M};
 
@@ -435,6 +436,13 @@ fn filter_to_where_clause(
     }
 
     where_clause.join(" and ")
+}
+
+fn hash_str(s: &str) -> String {
+    let mut hasher = Sha256::new();
+    hasher.update(s.as_bytes());
+
+    hex::encode(hasher.finish())
 }
 
 #[cfg(test)]
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -986,7 +986,7 @@ impl LiquidSdk {
                 };
                 let webhook = self.persister.get_webhook_url()?.map(|url| Webhook {
                     url,
-                    hash_swap_id: true,
+                    hash_swap_id: Some(true),
                 });
                 let create_response = self.swapper.create_send_swap(CreateSubmarineRequest {
                     from: "L-BTC".to_string(),
@@ -1192,7 +1192,7 @@ impl LiquidSdk {
         };
         let webhook = self.persister.get_webhook_url()?.map(|url| Webhook {
             url,
-            hash_swap_id: true,
+            hash_swap_id: Some(true),
         });
         let create_response = self.swapper.create_chain_swap(CreateChainRequest {
             from: "L-BTC".to_string(),
@@ -1461,7 +1461,7 @@ impl LiquidSdk {
 
         let webhook = self.persister.get_webhook_url()?.map(|url| Webhook {
             url,
-            hash_swap_id: true,
+            hash_swap_id: Some(true),
         });
         let v2_req = CreateReverseRequest {
             invoice_amount: payer_amount_sat as u32, // TODO update our model
@@ -1570,7 +1570,7 @@ impl LiquidSdk {
         };
         let webhook = self.persister.get_webhook_url()?.map(|url| Webhook {
             url,
-            hash_swap_id: true,
+            hash_swap_id: Some(true),
         });
         let create_response = self.swapper.create_chain_swap(CreateChainRequest {
             from: "BTC".to_string(),
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -984,19 +984,16 @@ impl LiquidSdk {
                     compressed: true,
                     inner: keypair.public_key(),
                 };
-                let webhook = self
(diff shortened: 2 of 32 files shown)


lib/core/src/sdk.rs
```diff
@@ -830,13 +830,28 @@ impl LiquidSdk {
...
+                    .persister
+                    .get_webhook_url()?
+                    .map(|url| SubmarineWebhook {
+                        url,
+                        hash_swap_id: Some(true),
+                        status: Some(vec![
+                            SubSwapStates::InvoiceFailedToPay,
+                            SubSwapStates::SwapExpired,
+                            SubSwapStates::TransactionClaimed,
+                            SubSwapStates::TransactionClaimPending,
```
roeierez, line 843: The pending is needed so the user app wakes up and cooperate with the swapper claim, that's understood.
But the TransactionClaimed will then invoke another webhook call, do we need it? I am concerned that in most cases the user will get an unnecessary notification.
In the case of non-cooperative claim by the swapper (which should be very rare) perhaps it is acceptable that when the user opens the app the refund will be broadcasted?

lib/core/src/sdk.rs
```diff
@@ -1213,6 +1240,17 @@ impl LiquidSdk {
         let mrh_addr_hash = sha256::Hash::hash(mrh_addr_str.as_bytes());
         let mrh_addr_hash_sig = keypair.sign_schnorr(mrh_addr_hash.into());
 
+        let webhook = self.persister.get_webhook_url()?.map(|url| ReverseWebhook {
+            url,
+            hash_swap_id: Some(true),
+            status: Some(vec![
+                RevSwapStates::InvoiceExpired,
```
roeierez, line 1247: Why do we need InvoiceExpired notification? Do we have any action item here?

lib/core/src/sdk.rs
```diff
@@ -1213,6 +1240,17 @@ impl LiquidSdk {
...
+                RevSwapStates::SwapExpired,
```
roeierez, line 1248: Same here. If the Swap expired do we have anything to do?

lib/core/src/sdk.rs
```diff
@@ -1213,6 +1240,17 @@ impl LiquidSdk {
...
+                RevSwapStates::TransactionConfirmed,
+                RevSwapStates::TransactionFailed,
```
roeierez, line 1250: Is it the swapper lockup that failed? If so I think we don't need that update as well?

lib/core/src/sdk.rs
```diff
@@ -1213,6 +1240,17 @@ impl LiquidSdk {
...
+                RevSwapStates::TransactionMempool,
```
roeierez, line 1251: We get both mempool and confirmed notification which means we will get two updates for the same payment.
Perhaps registering for only one of them based on whether we use zero conf or not for this swap?

lib/core/src/sdk.rs
```diff
@@ -1031,6 +1046,17 @@ impl LiquidSdk {
             compressed: true,
             inner: refund_keypair.public_key(),
         };
+        let webhook = self.persister.get_webhook_url()?.map(|url| ChainWebhook {
+            url,
+            hash_swap_id: Some(true),
+            status: Some(vec![
+                ChainSwapStates::SwapExpired,
```
roeierez, line 1053: Do we get SwapExpired only if we broadcasted the lockup or even for swaps that we didn't start yet?
We need to make sure the user is not spammed.

roeierez (changes requested):
Looks good!
My comments are mainly to ensure the UX doesn't spam the user with messages and we only get the necessary notifications in order to complete the swap.

lib/core/src/sdk.rs
dangeross: Yeah, TransactionClaimed would normally be the final status so could be left until the app is opened. Handling of this status validates the preimage and sets the swap to complete.

lib/core/src/sdk.rs
dangeross: I think for the reverse swap all these failure statuses can be removed. The swap will be set to failed when the app is opened.

lib/core/src/sdk.rs
dangeross: Will remove

lib/core/src/sdk.rs
dangeross: Will remove

lib/core/src/sdk.rs
dangeross: Sounds good. I know the TransactionConfirmed is skipped if it's 0-conf, but with two notifications for non 0-conf it would be spammy.

lib/core/src/sdk.rs
dangeross: We also get if the lockup fails. I think it's enough to handle TransactionLockupFailed and TransactionLockupFailed

lib/core/src/sdk.rs
roeierez: I agree

lib/core/src/sdk.rs
```diff
@@ -1031,6 +1045,16 @@ impl LiquidSdk {
             compressed: true,
             inner: refund_keypair.public_key(),
         };
+        let webhook = self.persister.get_webhook_url()?.map(|url| ChainWebhook {
+            url,
+            hash_swap_id: Some(true),
+            status: Some(vec![
+                ChainSwapStates::TransactionConfirmed,
```
roeierez, line 1052: As far as I understand TransactionConfirmed is sent when the user lockup is confirmed. I don't think we need to do anything at this state as we are waiting for the server transaction to confirm. Perhaps we need to omit it too?

lib/core/src/sdk.rs
```diff
@@ -1350,6 +1388,16 @@ impl LiquidSdk {
...
```
roeierez, line 1395: If this is the user lockup then again perhaps we don't need such notification?

roeierez (changes requested):

lib/core/src/sdk.rs
dangeross: True, this is our lockup confirmation and is already recorded

lib/core/src/sdk.rs
dangeross: Nothing is done but record the Bitcoin user_lockup_tx_id, we would need to add an alternative way to record this

roeierez: perhaps we need to fetch it onchain if the state is advanced and we missed it?

roeierez: Now thinking even without the notification feature we should take care of the case where we missed the TransactionConfirmed event as use can be offline while this happens. so if we need the user lockup tx id then we need to fetch it when is needed.

roeierez (approved):
LGTM

lib/core/src/persist/mod.rs
```diff
@@ -395,6 +396,13 @@ fn filter_to_where_clause(
     where_clause.join(" and ")
 }
 
+fn hash_str(s: &str) -> String {
+    let mut hasher = Sha256::new();
```
ok300, line 400: I would not add a dependency on openssl, as it's generally bloated and will make it harder if we want to switch from it in the future.

You can use the sha256 impl from the bitcoin crate instead, like we do for building the MRH:

```rust
let hash_bytes = sha256::Hash::hash(swap.id.as_bytes());
let hash_str = hash_bytes.to_byte_array().to_hex();
```

ok300 (approved):
Looks good, just one NIT.

PR #413, opened 2024-07-24
SDK: Add per-wallet directory structure for persister and lwk

### LINKED ISSUE #412: Use data dir per seed
In order to prevent colissions between two different wallets, it is better to create the persistent storage under another directory baed on the seed (we can hash the seed).
So the sqlite will be located under: 
<data_dir>/<seed hash>

Ther are some nasty issues when android doesn't delete the application folder on uninstall on some configurations which prevents to that colissions

hydra-yse:
Closes #412

### DIFF
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -1,11 +1,16 @@
 use std::path::PathBuf;
 
 use anyhow::{anyhow, Result};
-use boltz_client::network::Chain;
-use boltz_client::swaps::boltz::{
-    CreateChainResponse, CreateReverseResponse, CreateSubmarineResponse, Leaf, Side, SwapTree,
+
+use boltz_client::{
+    network::Chain,
+    swaps::boltz::{
+        CreateChainResponse, CreateReverseResponse, CreateSubmarineResponse, Leaf, Side, SwapTree,
+    },
+    ToHex,
 };
 use boltz_client::{BtcSwapScript, BtcSwapTx, Keypair, LBtcSwapScript, LBtcSwapTx};
+use lwk_signer::SwSigner;
 use lwk_wollet::ElementsNetwork;
 use rusqlite::types::{FromSql, FromSqlError, FromSqlResult, ToSqlOutput, ValueRef};
 use rusqlite::ToSql;
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -71,16 +71,9 @@ impl Config {
         }
     }
 
-    pub(crate) fn get_wallet_working_dir(
-        &self,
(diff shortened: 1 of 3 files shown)


roeierez (approved):

dangeross (approved):

lib/core/src/model.rs
```diff
@@ -69,6 +72,23 @@ impl Config {
         }
     }
 
+    pub(crate) fn get_wallet_working_dir(
+        &self,
+        config: &Config,
+        mnemonic: &str,
```
ok300, line 78: Why not pass `signer.xpub().public_key` as the arg here instead of `mnemonic`?

It would limit the number of places `mnemonic` is propagated to, yet would still give us the PK we want.

lib/core/src/wallet.rs
```diff
@@ -71,7 +74,7 @@ impl LiquidOnchainWallet {
         })
     }
 
-    fn get_descriptor(
+    pub(crate) fn get_descriptor(
```
ok300, line 77: I don't think this needs to be exposed.

ok300:
I have a few questions:
- Should `sdk.log` be also stored in the per-wallet directory? It's now stored just above them, being a common log for all wallets.
- How about backward compatibility? If someone used the SDK before this and has a DB file and LWK cache in the old path, it won't be recognized after this commit. This means swap details are lost. Shouldn't we try to migrate (move) any existing data to the new path on startup?

lib/core/src/model.rs
ok300: Wouldn't it be better to use `signer.fingerprint()` instead of the xpub PK? The fingerprint is the first 4 bytes of the HASH160 of the PK.

It's shorter (4 bytes / 8 hex chars) compared to the PK (33 bytes / 66 chars).

roeierez:
> I have a few questions:
> 
> * Should `sdk.log` be also stored in the per-wallet directory? It's now stored just above them, being a common log for all wallets.
> * How about backward compatibility? If someone used the SDK before this and has a DB file and LWK cache in the old path, it won't be recognized after this commit. This means swap details are lost. Shouldn't we try to migrate (move) any existing data to the new path on startup?

@ok300  Unless you have any idea I don't see how we upgrade and ensure the data is compatible. The whole movitvation for this PR is to prevent mixture of data. In this case the user would get the "retore" experience, e.g onchain data without the swap related info for old transactions.
That's why it is important to release this early.

ok300 (approved):
Looks good!

@roeierez you're right, it's a bit tricky with data migration. We can skip this and add a release note instead, if necessary.

PR #414, opened 2024-07-25
Send/Receive: Add BIP21 support

### LINKED ISSUE #405: Support send/receive via bip21
The idea is to allow sending/receiving L-BTC directly when possible.

hydra-yse:
Closes #405.

Currently supported:
- [X] Send
- [x] Receive

**Note:** Switching to [yse-bip21-liquid](/breez/breez-sdk/tree/yse-bip21-liquid) until breez/breez-sdk#1058 is merged.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -19,7 +19,6 @@ use serde_json::to_string_pretty;
 
 #[derive(Parser, Debug, Clone, PartialEq)]
 pub(crate) enum Command {
-    /// TODO Add BIP21 Support
     /// Send lbtc and receive btc lightning through a swap
     SendPayment {
         /// Invoice which has to be paid
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -19,10 +19,21 @@ use serde_json::to_string_pretty;
 
 #[derive(Parser, Debug, Clone, PartialEq)]
 pub(crate) enum Command {
+    /// TODO Add BIP21 Support
     /// Send lbtc and receive btc lightning through a swap
     SendPayment {
         /// Invoice which has to be paid
-        bolt11: String,
+        #[arg(long)]
+        bolt11: Option<String>,
+
+        /// Either BIP21 URI or Liquid address we intend to pay to
+        #[arg(long)]
+        address: Option<String>,
+
+        /// The amount in satoshi to pay, in case of a direct Liquid address
+        /// or amount-less BIP21
+        #[arg(short, long)]
+        amount_sat: Option<u64>,
 
         /// Delay for the send, in seconds
         #[arg(short, long)]
@@ -46,17 +57,18 @@ pub(crate) enum Command {
     },
     /// Receive lbtc and send btc through a swap
     ReceivePayment {
+        /// The method to use when receiving. Either "lightning", "bitcoin" or "liquid"
+        #[arg(short = 'm', long = "method")]
+        payment_method: Option<PaymentMethod>,
+
         /// Amount the payer will send, in satoshi
-        payer_amount_sat: u64,
+        /// If not specified, it will generate a BIP21 URI/Liquid address with no amount
+        #[arg(short, long)]
+        payer_amount_sat: Option<u64>,
 
         /// Optional description for the invoice
         #[clap(short = 'd', long = "description")]
         description: Option<String>,
-    },
-    /// Receive lbtc and send btc onchain through a swap
-    ReceiveOnchainPayment {
-        /// Amount the payer will send, in satoshi
-        payer_amount_sat: u64,
     },
     /// Generates an URL to buy bitcoin from a 3rd party provider
     BuyBitcoin {
@@ -200,32 +212,49 @@ pub(crate) async fn handle_command(
 ) -> Result<String> {
     Ok(match command {
         Command::ReceivePayment {
+            payment_method,
             payer_amount_sat,
             description,
         } => {
-            let prepare_res = sdk
-                .prepare_receive_payment(&PrepareReceivePaymentRequest { payer_amount_sat })
+            let prepare_response = sdk
+                .prepare_receive_payment(&PrepareReceiveRequest {
+                    payer_amount_sat,
+                    payment_method: payment_method.unwrap_or(PaymentMethod::Lightning),
+                })
                 .await?;
 
             wait_confirmation!(
                 format!(
                     "Fees: {} sat. Are the fees acceptable? (y/N) ",
-                    prepare_res.fees_sat
+                    prepare_response.fees_sat
                 ),
                 "Payment receive halted"
             );
 
             let response = sdk
                 .receive_payment(&ReceivePaymentRequest {
-                    prepare_res,
+                    prepare_response,
                     description,
                 })
                 .await?;
-            let invoice = response.invoice.clone();
 
-            let mut result = command_result!(response);
+            let mut result = command_result!(&response);
             result.push('\n');
-            result.push_str(&build_qr_text(&invoice));
+
+            match parse(&response.destination).await? {
+                InputType::Bolt11 { invoice } => result.push_str(&build_qr_text(&invoice.bolt11)),
+                InputType::LiquidAddress { address } => {
+                    result.push_str(&build_qr_text(&address.to_uri().map_err(|e| {
+                        anyhow::anyhow!("Could not build BIP21 from address data: {e:?}")
+                    })?))
+                }
+                InputType::BitcoinAddress { address } => {
+                    result.push_str(&build_qr_text(&address.to_uri().map_err(|e| {
+                        anyhow::anyhow!("Could not build BIP21 from address data: {e:?}")
+                    })?))
+                }
+                _ => {}
+            }
             result
         }
         Command::FetchLightningLimits => {
@@ -236,9 +265,32 @@ pub(crate) async fn handle_command(
             let limits = sdk.fetch_onchain_limits().await?;
             command_result!(limits)
         }
-        Command::SendPayment { bolt11, delay } => {
+        Command::SendPayment {
+            bolt11,
+            address,
+            amount_sat,
+            delay,
+        } => {
+            let destination = match (bolt11, address) {
+                (None, None) => {
+                    return Err(anyhow::anyhow!(
+                        "Must specify either a `bolt11` invoice or a direct/BIP21 `address`."
+                    ))
+                }
+                (Some(bolt11), None) => bolt11,
+                (None, Some(address)) => address,
+                (Some(_), Some(_)) => {
+                    return Err(anyhow::anyhow!(
+                        "Cannot specify both `bolt11` and `address` at the same time."
+                    ))
+                }
+            };
+
             let prepare_response = sdk
-                .prepare_send_payment(&PrepareSendRequest { invoice: bolt11 })
+                .prepare_send_payment(&PrepareSendRequest {
+                    destination,
+                    amount_sat,
+                })
                 .await?;
 
             wait_confirmation!(
@@ -249,17 +301,20 @@ pub(crate) async fn handle_command(
                 "Payment send halted"
             );
 
+            let send_payment_req = SendPaymentRequest {
+                prepare_response: prepare_response.clone(),
+            };
+
             if let Some(delay) = delay {
                 let sdk_cloned = sdk.clone();
-                let prepare_cloned = prepare_response.clone();
 
                 tokio::spawn(async move {
                     thread::sleep(Duration::from_secs(delay));
-                    sdk_cloned.send_payment(&prepare_cloned).await.unwrap();
+                    sdk_cloned.send_payment(&send_payment_req).await.unwrap();
                 });
                 command_result!(prepare_response)
             } else {
-                let response = sdk.send_payment(&prepare_response).await?;
+                let response = sdk.send_payment(&send_payment_req).await?;
                 command_result!(response)
             }
         }
@@ -268,7 +323,7 @@ pub(crate) async fn handle_command(
             receiver_amount_sat,
             sat_per_vbyte,
         } => {
-            let prepare_res = sdk
+            let prepare_response = sdk
                 .prepare_pay_onchain(&PreparePayOnchainRequest {
                     receiver_amount_sat,
                     sat_per_vbyte,
@@ -278,7 +333,7 @@ pub(crate) async fn handle_command(
             wait_confirmation!(
                 format!(
                     "Fees: {} sat (incl claim fee: {} sat). Are the fees acceptable? (y/N) ",
-                    prepare_res.total_fees_sat, prepare_res.claim_fees_sat
+                    prepare_response.total_fees_sat, prepare_response.claim_fees_sat
                 ),
                 "Payment send halted"
             );
@@ -286,37 +341,16 @@ pub(crate) async fn handle_command(
             let response = sdk
                 .pay_onchain(&PayOnchainRequest {
                     address,
-                    prepare_res,
+                    prepare_response,
                 })
                 .await?;
             command_result!(response)
-        }
-        Command::ReceiveOnchainPayment { payer_amount_sat } => {
-            let prepare_res = sdk
-                .prepare_receive_onchain(&PrepareReceiveOnchainRequest { payer_amount_sat })
-                .await?;
-
-            wait_confirmation!(
-                format!(
-                    "Fees: {} sat. Are the fees acceptable? (y/N) ",
-                    prepare_res.fees_sat
-                ),
-                "Payment receive halted"
-            );
-
-            let response = sdk.receive_onchain(&prepare_res).await?;
-            let bip21 = response.bip21.clone();
-
-            let mut result = command_result!(response);
-            result.push('\n');
-            result.push_str(&build_qr_text(&bip21));
-            result
         }
         Command::BuyBitcoin {
             provider,
             amount_sat,
         } => {
-            let prepare_res = sdk
+            let prepare_response = sdk
                 .prepare_buy_bitcoin(&PrepareBuyBitcoinRequest {
                     provider,
                     amount_sat,
@@ -326,14 +360,14 @@ pub(crate) async fn handle_command(
             wait_confirmation!(
                 format!(
                     "Fees: {} sat. Are the fees acceptable? (y/N) ",
-                    prepare_res.fees_sat
+                    prepare_response.fees_sat
                 ),
                 "Buy Bitcoin halted"
             );
 
             let url = sdk
                 .buy_bitcoin(&BuyBitcoinRequest {
-                    prepare_res,
+                    prepare_response,
                     redirect_url: None,
                 })
                 .await?;
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -278,10 +278,12 @@ enum PaymentError {
     "AlreadyPaid",
     "PaymentInProgress",
     "AmountOutOfRange",
+    "AmountMissing",
     "Generic",
     "InvalidOrExpiredFees",
     "InsufficientFunds",
     "InvalidInvoice",
+    "InvalidNetwork",
     "InvalidPreimage",
     "LwkError",
     "PairsNotFound",
@@ -323,35 +325,53 @@ dictionary GetInfoResponse {
 };
 
 dictionary PrepareSendRequest {
-    string invoice;
+    string destination;
+    u64? amount_sat;
+};
+
+[Enum]
+interface SendDestination {
+    LiquidAddress(LiquidAddressData address_data);
+    Bolt11(LNInvoice invoice);
 };
 
 dictionary PrepareSendResponse {
-    string invoice;
+    SendDestination destination;
     u64 fees_sat;
+};
+
+dictionary SendPaymentRequest {
+    PrepareSendResponse prepare_response;
 };
 
 dictionary SendPaymentResponse {
     Payment payment;
 };
 
-dictionary PrepareReceivePaymentRequest {
-    u64 payer_amount_sat;
+enum PaymentMethod {
+    "Lightning",
+    "BitcoinAddress",
+    "LiquidAddress",
 };
 
-dictionary PrepareReceivePaymentResponse {
-    u64 payer_amount_sat;
+dictionary PrepareReceiveRequest {
+    u64? payer_amount_sat;
+    PaymentMethod payment_method;
+};
+
+dictionary PrepareReceiveResponse {
+    u64? payer_amount_sat;
+    PaymentMethod payment_method;
     u64 fees_sat;
 };
 
 dictionary ReceivePaymentRequest {
-    PrepareReceivePaymentResponse prepare_res;
+    PrepareReceiveResponse prepare_response;
     string? description = null;
 };
 
 dictionary ReceivePaymentResponse {
-    string id;
-    string invoice;
+    string destination;
 };
 
 dictionary Limits {
@@ -383,21 +403,7 @@ dictionary PreparePayOnchainResponse {
 
 dictionary PayOnchainRequest {
     string address;
-    PreparePayOnchainResponse prepare_res;
-};
-
-dictionary PrepareReceiveOnchainRequest {
-    u64 payer_amount_sat;
-};
-
-dictionary PrepareReceiveOnchainResponse {
-    u64 payer_amount_sat;
-    u64 fees_sat;
-};
-
-dictionary ReceiveOnchainResponse {
-    string address;
-    string bip21;
+    PreparePayOnchainResponse prepare_response;
 };
 
 enum BuyBitcoinProvider {
@@ -416,7 +422,7 @@ dictionary PrepareBuyBitcoinResponse {
 };
 
 dictionary BuyBitcoinRequest {
-    PrepareBuyBitcoinResponse prepare_res;
+    PrepareBuyBitcoinResponse prepare_response;
     string? redirect_url = null;
 };
 
@@ -436,19 +442,22 @@ dictionary ListPaymentsRequest {
     u32? limit = null;
 };
 
+[Enum]
+interface PaymentDetails {
+    Lightning(string swap_id, string description, string? preimage, string? bolt11, string? refund_tx_id, u64? refund_tx_amount_sat);
+    Liquid(string destination, string description);
+    Bitcoin(string swap_id, string description, string? refund_tx_id, u64? refund_tx_amount_sat);
+};
+
 dictionary Payment {
+    string? destination;
+    string? tx_id = null;
     u32 timestamp;
     u64 amount_sat;
     u64 fees_sat;
     PaymentType payment_type;
     PaymentState status;
-    string description;
-    string? tx_id = null;
-    string? swap_id = null;
-    string? preimage = null;
-    string? bolt11 = null;
-    string? refund_tx_id = null;
-    u64? refund_tx_amount_sat = null;
+    PaymentDetails? details;
 };
 
 enum PaymentType {
@@ -556,10 +565,10 @@ interface BindingLiquidSdk {
     PrepareSendResponse prepare_send_payment(PrepareSendRequest req);
 
     [Throws=PaymentError]
-    SendPaymentResponse send_payment(PrepareSendResponse req);
+    SendPaymentResponse send_payment(SendPaymentRequest req);
 
     [Throws=PaymentError]
-    PrepareReceivePaymentResponse prepare_receive_payment(PrepareReceivePaymentRequest req);
+    PrepareReceiveResponse prepare_receive_payment(PrepareReceiveRequest req);
 
     [Throws=PaymentError]
     ReceivePaymentResponse receive_payment(ReceivePaymentRequest req);
@@ -575,12 +584,6 @@ interface BindingLiquidSdk {
 
     [Throws=PaymentError]
     SendPaymentResponse pay_onchain(PayOnchainRequest req);
-
-    [Throws=PaymentError]
-    PrepareReceiveOnchainResponse prepare_receive_onchain(PrepareReceiveOnchainRequest req);
-
-    [Throws=PaymentError]
-    ReceiveOnchainResponse receive_onchain(PrepareReceiveOnchainResponse req);
 
     [Throws=PaymentError]
     PrepareBuyBitcoinResponse prepare_buy_bitcoin(PrepareBuyBitcoinRequest req);
--- a/lib/core/Cargo.toml
+++ b/lib/core/Cargo.toml
@@ -17,7 +17,9 @@ bip39 = "2.0.0"
 boltz-client = { git = "https://github.com/dangeross/boltz-rust", branch = "savage-breez-latest" }
 chrono = "0.4"
 env_logger = "0.11"
-flutter_rust_bridge = { version = "=2.2.0", features = ["chrono"], optional = true }
+flutter_rust_bridge = { version = "=2.2.0", features = [
+  "chrono",
+], optional = true }
 log = { workspace = true }
 lwk_common = "0.7.0"
 lwk_signer = "0.7.0"
@@ -36,7 +38,10 @@ openssl = { version = "0.10", features = ["vendored"] }
 tokio = { version = "1", features = ["rt", "macros"] }
 tokio-stream = { version = "0.1.14", features = ["sync"] }
 url = "2.5.0"
-futures-util = { version = "0.3.28", default-features = false, features = ["sink", "std"] }
+futures-util = { version = "0.3.28", default-features = false, features = [
+  "sink",
+  "std",
+] }
 async-trait = "0.1.80"
 hex = "0.4"
 reqwest = { version = "=0.11.20", features = ["json"] }
--- a/lib/core/src/bindings.rs
+++ b/lib/core/src/bindings.rs
@@ -101,15 +101,15 @@ impl BindingLiquidSdk {
 
     pub async fn send_payment(
         &self,
-        req: PrepareSendResponse,
+        req: SendPaymentRequest,
     ) -> Result<SendPaymentResponse, PaymentError> {
         self.sdk.send_payment(&req).await
     }
 
     pub async fn prepare_receive_payment(
         &self,
-        req: PrepareReceivePaymentRequest,
-    ) -> Result<PrepareReceivePaymentResponse, PaymentError> {
+        req: PrepareReceiveRequest,
+    ) -> Result<PrepareReceiveResponse, PaymentError> {
         self.sdk.prepare_receive_payment(&req).await
     }
 
@@ -142,20 +142,6 @@ impl BindingLiquidSdk {
         req: PayOnchainRequest,
     ) -> Result<SendPaymentResponse, PaymentError> {
         self.sdk.pay_onchain(&req).await
-    }
-
-    pub async fn prepare_receive_onchain(
-        &self,
-        req: PrepareReceiveOnchainRequest,
-    ) -> Result<PrepareReceiveOnchainResponse, PaymentError> {
-        self.sdk.prepare_receive_onchain(&req).await
-    }
-
-    pub async fn receive_onchain(
-        &self,
-        req: PrepareReceiveOnchainResponse,
-    ) -> Result<ReceiveOnchainResponse, PaymentError> {
-        self.sdk.receive_onchain(&req).await
     }
 
     pub async fn prepare_buy_bitcoin(
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -194,16 +194,29 @@ pub struct ConnectRequest {
     pub config: Config,
 }
 
+/// The send/receive methods supported by the SDK
+#[derive(Clone, Debug, EnumString, Serialize, Eq, PartialEq)]
+pub enum PaymentMethod {
+    #[strum(serialize = "lightning")]
+    Lightning,
+    #[strum(serialize = "bitcoin")]
+    BitcoinAddress,
+    #[strum(serialize = "liquid")]
+    LiquidAddress,
+}
+
 /// An argument when calling [crate::sdk::LiquidSdk::prepare_receive_payment].
 #[derive(Debug, Serialize)]
-pub struct PrepareReceivePaymentRequest {
-    pub payer_amount_sat: u64,
+pub struct PrepareReceiveRequest {
+    pub payer_amount_sat: Option<u64>,
+    pub payment_method: PaymentMethod,
 }
 
 /// Returned when calling [crate::sdk::LiquidSdk::prepare_receive_payment].
 #[derive(Debug, Serialize)]
-pub struct PrepareReceivePaymentResponse {
-    pub payer_amount_sat: u64,
+pub struct PrepareReceiveResponse {
+    pub payment_method: PaymentMethod,
+    pub payer_amount_sat: Option<u64>,
     pub fees_sat: u64,
 }
 
@@ -211,14 +224,15 @@ pub struct PrepareReceivePaymentResponse {
 #[derive(Debug, Serialize)]
 pub struct ReceivePaymentRequest {
     pub description: Option<String>,
-    pub prepare_res: PrepareReceivePaymentResponse,
+    pub prepare_response: PrepareReceiveResponse,
 }
 
 /// Returned when calling [crate::sdk::LiquidSdk::receive_payment].
 #[derive(Debug, Serialize)]
 pub struct ReceivePaymentResponse {
-    pub id: String,
-    pub invoice: String,
+    /// Either a BIP21 URI (Liquid or Bitcoin), a Liquid address
+    /// or an invoice, depending on the [PrepareReceivePaymentResponse] parameters
+    pub destination: String,
 }
 
 /// The minimum and maximum in satoshis of a Lightning or onchain payment.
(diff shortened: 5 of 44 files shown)


hydra-yse:
CC @roeierez
While implementing BIP21 features is fairly straightforward, there are some questions related to the interface which we would like to expose to the user that need some clarification:

## Send
1. Do we need a separate method to send via BIP21 or should we reuse existing methods (specifically `send_payment` or `pay_onchain`) and rework their logic?
2. If we reuse existing methods, how should the interface look like? Here's the main alternatives we thought of:
```rust
// Option 1
pub struct PrepareSendRequest {
    pub invoice: Option<String>,
    pub bip21: Option<String>,
}

// Option 2
pub enum PrepareSendRequest {
    Bip21 { bip21: String },
    Bolt11 { bolt11: String },
}

// Option 3
pub struct PrepareSendRequest {
    pub input: String, // Gets parsed by the SDK, which then returns the appropriate output type
}
```

In either one of the cases, the `PrepareSendResponse` should look something like:
```rust
// Option 1
pub struct PrepareSendResponse {
    pub invoice: Option<String>,
    pub bip21: Option<String>,
}

// Option 2
pub enum PrepareSendResponse {
    Bip21 { bip21: String },
    Bolt11 { bolt11: String },
}
```
In order to make the output more comprehensible. Like in the request, only one field can be set at once.

## Receive

1. With regards to the interface, same question. New method, or altering `receive_payment`/`receive_onchain`'s logic?

2. In case we keep the same method, we thought adding a flag to the input should be enough; something like:

```rust
// Option 1
pub struct PrepareReceivePaymentRequest {
    pub payer_amount_sat: u64,
    pub use_bip21: bool,
}

// Option 2 (discouraged)
pub enum PrepareReceivePaymentRequest {
    Bip21 { payer_amount_sat: u64 },
    Bolt11 { payer_amount_sat: u64 },
}
```

For the inputs, and:

```rust
// Option 1
// Note this is *NOT* PrepareReceivePaymentResponse
pub struct ReceivePaymentResponse {
    pub id: String, // Swap id
    pub invoice: String,
    pub bip21: String,
}


// Option 2
pub struct ReceivePaymentResponse {
    pub id: String, // Swap id
    pub invoice: Option<String>,
    pub bip21: Option<String>,
}


// Option 3
pub enum ReceivePaymentResponse {
    Bip21 { bip21: String },
    Bolt11 { bolt11: String },
}
```
For the outputs, choosing whether we return both the invoice and the bip21 url, or only one of the two.

roeierez:
@hydra-yse I had a chat with @kingonly to understand better the requriement and one of the key goals here is to make it as simple as we can for the devs not requiring them to fully understand specs and protocols.

Here is a suggestion aligned with the above, let me know what you think:

```
PrepareSendPaymentRequest {
  pub destination: string // didn't find the right name yet but it should get all payment methods (bolt11, bip21, liquid address)
}

PrepareSendPaymentResponse {
  pub parsed_destination: InputType // parsed input  that the sdk is going to pay for (bolt11, liquid address)
  pub fees_sat: u64 // The corresponding fees. If we are going to pay to a liquid address then only the lowball onchain fees.
}
```

hydra-yse:
@roeierez Sure I agree, if the goal here is to make this as direct as possible then I agree we should merge all fields into one. Regarding the output, while returning a field `InputType` is not as direct and comprehensible we assume that the usual case is to just pass the result directly into `send_payment`, so there shouldn't be any issues there. More advanced users can then still pattern match into the field for more specific use-cases.

roeierez:
> @roeierez Sure, if the goal here is to make this as direct as possible then I agree we should merge all fields into one. Regarding the output, while returning a field `InputType` is not as direct and comprehensible we assume that the usual case is to just pass the result directly into `send_payment`, so there shouldn't be any issues there. More advanced users can then still pattern match into the field for more specific use-cases. What about receive? Are you still discussing?

I still think on the receive yes. if for receive (with a flag) we return bip21 then the returned fees should be changes as well as we have two use cases. That's the point I am trying to simplify. If we would return liquid address then no issues becaue the payer has only one choice of paying.

roeierez:
@hydra-yse had a discussion with @kingonly and we got something well defined.
For the receive 
1. We should add another optional parameter that will define the receive method (bolt11, bip21) bolt11 is the default.
2. If user asked for bip21 with X amount then we will put the X amount as the bip21 parameter and in the case we receive to that onchain address we assume no fees on the receiver.

Does it work for you?

lib/core/src/sdk.rs
```diff
@@ -676,23 +676,50 @@ impl LiquidSdk {
...
-        self.ensure_send_is_not_self_transfer(&req.invoice)?;
-        let invoice = self.validate_invoice(&req.invoice)?;
+        let fees_sat;
+        let receiver_amount_sat;
+        let parsed_destination = sdk_common::input_parser::parse(&req.destination).await?;
+        match &parsed_destination {
+            InputType::LiquidAddress {
+                address: liquid_address_data,
+            } => {
+                let Some(amount_sat) = liquid_address_data.amount_sat else {
```
roeierez, line 686: It seems that if the bip21 won't contain any amount (as it is optional) we will fail here.
Do we have to make it required?

lib/core/src/sdk.rs
```diff
@@ -735,48 +762,94 @@ impl LiquidSdk {
...
+            InputType::LiquidAddress {
+                address: liquid_address_data,
+            } => {
+                // We default to paying the invoice encoded within the BIP21
+                // if it is present
+                if let Some(invoice) = &liquid_address_data.invoice {
+                    return self.check_send_payment_invoice(invoice, req.fees_sat).await;
+                }
+
+                let Some(amount_sat) = liquid_address_data.amount_sat else {
```
roeierez, line 775: As bip21 won't necessaryily have an amount perhaps we need to change the send_payment to receive a different argument?
`send_payment(SendPaymentRequest)`

Where

```
SendPaymentRequest {
 prepare_response: PrepareSendResponse
 amount_sat: Option<u64>
}
```

roeierez (changes requested):

lib/core/src/sdk.rs
hydra-yse: What is the expected flow? If the BIP21 doesn't contain any information on the amount we can't calculate fees, etc. which is what the prepare statement is primarily for.

Perhaps we could set the fees as optional and when no amount is given we can just return no fees and give the ability to set an amount in the `send_payment` method?

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -313,11 +324,12 @@ dictionary GetInfoResponse {
...
 
 dictionary PrepareSendRequest {
-    string invoice;
+    string destination;
+    f32? fee_rate;
 };
 
 dictionary PrepareSendResponse {
-    string invoice;
+    InputType parsed_destination;
```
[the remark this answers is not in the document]
hydra-yse: Agreed :+1:

lib/core/src/sdk.rs
hydra-yse: Oh, I see we thought of the same alternative. Then I agree, will fix this way

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -313,11 +324,12 @@ dictionary GetInfoResponse {
 };
 
 dictionary PrepareSendRequest {
-    string invoice;
+    string destination;
+    f32? fee_rate;
```
roeierez, line 328: We don't need fee rate here, we will use the smallest we can on liquid.

roeierez (changes requested):

lib/core/src/sdk.rs
roeierez: I see. 
So perhaps we should pass tothe prepare_send_payment also the amount?

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -313,25 +324,43 @@ dictionary GetInfoResponse {
...
 dictionary PrepareSendRequest {
-    string invoice;
+    string payment_destination;
+    u64? amount_sat;
+};
+
+[Enum]
+interface PaymentDestination {
+    BIP21(LiquidAddressData address);
+    Bolt11(string invoice);
```
roeierez, line 334: Since in Bip21 we have the parsed data (LiquidAddressData) it would be more consistent if the Bolt11 contains the LNInvoice isntead of the string.

roeierez (changes requested):

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -313,35 +324,59 @@ dictionary GetInfoResponse {
...
 
+dictionary SendPaymentRequest {
+    PrepareSendResponse prepare_response;
+};
+
 dictionary SendPaymentResponse {
     Payment payment;
 };
 
+enum ReceiveMethod {
```
roeierez, line 350: Seems like we don't use it anymore.

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -313,35 +324,59 @@ dictionary GetInfoResponse {
...
-    PrepareReceivePaymentResponse prepare_res;
+    PrepareReceivePaymentResponse prepare_response;
     string? description = null;
 };
 
+[Enum]
+interface ReceiveDestination {
+    BIP21(string uri);
+    Liquid(string address):
+    Bolt11(string id, string invoice);
```
roeierez, line 375: Why do we need the id?

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -313,35 +324,59 @@ dictionary GetInfoResponse {
...
 };
 
 dictionary ReceivePaymentRequest {
-    PrepareReceivePaymentResponse prepare_res;
+    PrepareReceivePaymentResponse prepare_response;
     string? description = null;
 };
 
+[Enum]
+interface ReceiveDestination {
```
roeierez, line 372: I am wondering if we should have this struct or keep it simple and return a string in the response.
No strong opinion. No strong opnion but since all of the possible results are strings then I am not sure what is the benefit of returning an enum.

lib/core/src/sdk.rs
```diff
@@ -1187,9 +1299,45 @@ impl LiquidSdk {
...
+            (false, None) => {
+                let address = self.onchain_wallet.next_unused_address().await?.to_string();
+                Ok(ReceivePaymentResponse {
+                    receive_destination: ReceiveDestination::Liquid { address },
+                })
+            }
+            (false, Some(payer_amount_sat)) => {
+                let address = self.onchain_wallet.next_unused_address().await?.to_string();
+                let uri =
+                    utils::new_liquid_bip21(&address, self.config.network, Some(payer_amount_sat));
```
roeierez, line 1327: Should we add the description as "message" or "label" in the resulted bip21 ?

roeierez (changes requested):
Looks good, some final comments.

lib/bindings/src/breez_sdk_liquid.udl
hydra-yse: In the previous receive response we exposed the swap id, but thinking about it now the user doesn't need to be aware of it

lib/core/src/sdk.rs
hydra-yse: I believe it should be the message field, the label is meant for small strings/identifiers as far as I understand

lib/bindings/src/breez_sdk_liquid.udl
roeierez: Yes I also think we don't need it. Let's remove it and we are left only with the result string.

lib/core/src/model.rs
```diff
@@ -197,28 +197,37 @@ pub struct ConnectRequest {
 /// An argument when calling [crate::sdk::LiquidSdk::prepare_receive_payment].
 #[derive(Debug, Serialize)]
 pub struct PrepareReceivePaymentRequest {
-    pub payer_amount_sat: u64,
+    pub payer_amount_sat: Option<u64>,
+    pub use_lightning: bool,
```
ok300, line 201: Doesn't this complicate things? It's now a 2x2 matrix of possible inputs, where only 3 combinations are valid. Also the amount range should only be validated when using swaps.

No strong opinion, but IMO it would be clearer to the caller (and easier for us to validate) if we ask for

```rust
pub enum PrepareReceivePaymentRequest {
    Lightning(payer_amount_sat: u64),
    Onchain(payer_amount_sat: Option<u64>)
}
```

lib/bindings/src/breez_sdk_liquid.udl
ok300: Two NITs:

1) The three options can be reduced to two: `Bolt11` and `BIP21`, since `BIP21` can also cover a pure Liquid address.

2) In case of using Lightning (swaps), technically there is also a `BIP21` string that gets created in the MRH and can be used by onchain wallets as well. So the two options are more like `LightningAndOnchain` and `OnchainOnly` (exact naming TBD). This means if LN is chosen, we can return both the invoice and the BIP21 (decoded from the MRH).

For example like so
```
interface ReceiveDestination {
    OnchainOnly(string bip21):
    LightningAndOnchain(string invoice, string bip21);
}
```

Then it would help callers if we return this struct, even if it has only strings, e.g. for UIs that can alternate between "pay this invoice" and "pay onchain".

lib/core/src/sdk.rs
```diff
@@ -731,52 +775,102 @@ impl LiquidSdk {
...
         lbtc_address: &str,
-        amount_btc: f64,
+        receiver_amount_sat: u64,
+        fees_sat: u64,
     ) -> Result<SendPaymentResponse, PaymentError> {
-        let amount_sat: u64 = (amount_btc * 100_000_000.0) as u64;
-        info!("Found MRH for L-BTC address {lbtc_address} and amount_sat {amount_sat}");
-
-        let receiver_amount_sat = get_invoice_amount!(req.invoice);
+        // TODO Ensure that `None` provides the lowest fees possible (0.01 sat/vbyte)
```
ok300, line 853: Which `None` does this refer to? And is this comment still relevant to the PR, or can be removed?

lib/core/src/utils.rs
```diff
@@ -59,6 +64,22 @@ pub(crate) fn deserialize_tx_hex(tx_hex: &str) -> Result<Transaction> {
...
+    absolute_fees: u64,
+) -> Result<f32> {
+    let standard_fees = wallet
+        .build_tx(None, recipient_address, amount_sat)
+        .await?
+        .all_fees()
+        .values()
+        .sum::<u64>() as f32;
+
+    Ok(0.1 * absolute_fees as f32 / standard_fees)
```
ok300, line 80: Is the `0.1` actually the base feerate that's being "adjusted" with the division? If so, then why not use `STANDARD_FEE_RATE_SAT_PER_VBYTE` constant instead, so it's clear what it represents and it can be replaced easily in the future?

ok300 (commented):
Looks good, found only a few NITs.

lib/core/src/sdk.rs
hydra-yse: It's relevant, it ties in to #423. Once we'll be able to broadcast via our own node we'll need a global method that can tell us exactly which fees to use any broadcast since we won't be limited by Boltz API

lib/core/src/utils.rs
hydra-yse: Agreed, will do

lib/core/src/model.rs
```diff
@@ -14,7 +14,7 @@ use lwk_signer::SwSigner;
 use lwk_wollet::ElementsNetwork;
 use rusqlite::types::{FromSql, FromSqlError, FromSqlResult, ToSqlOutput, ValueRef};
 use rusqlite::ToSql;
-use sdk_common::prelude::*;
+use sdk_common::{liquid::LiquidAddressData, prelude::*};
```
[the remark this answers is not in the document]
hydra-yse: I'm not sure about the ```pub use crate::liquid::*;```, but you're right, I can just replace `LiquidAddressData` with `liquid::LiquidAddressData` and use that instead

lib/core/src/model.rs
ok300: To expand on Roei's comment above:

1. is IMO more aligned with how we have two methods for payment: `pay_onchain` and `send_payment`, i.e. we don't cover both payment modalities in a single method. Overall this would result in three receive variants: `receive_payment`,  `receive_onchain` and `receive_liquid_payment`.

2. is possible because every swap we create also includes a BIP21 fallback encoded in a routing hint. If we return both, we make it easy for the caller to decide what to show the end-user ("pay this invoice" vs "pay to onchain Liquid address"). The one downside here is it's more complicated to communicate the fees, as it depends if the payer chooses the invoice or the BIP21.

lib/core/src/model.rs
```diff
@@ -194,31 +194,42 @@ pub struct ConnectRequest {
...
 }
 
 /// Returned when calling [crate::sdk::LiquidSdk::receive_payment].
 #[derive(Debug, Serialize)]
 pub struct ReceivePaymentResponse {
-    pub id: String,
-    pub invoice: String,
+    /// Either a BIP21 URI (Liquid or Bitcoin), a Liquid address
+    /// or an invoice, depending on the [PrepareReceivePaymentResponse] parameters
+    pub receive_destination: String,
```
roeierez, line 232: NIT: just `destination` ?

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -313,35 +325,47 @@ dictionary GetInfoResponse {
...
 
 dictionary PrepareSendRequest {
-    string invoice;
+    string payment_destination;
+    u64? amount_sat;
+};
+
+[Enum]
+interface SendDestination {
+    BIP21(LiquidAddressData address_data);
```
roeierez, line 334: perhaps rename to `LiquidAddress(LiquidAddressData address_data);`
So we are consistent with the receive ?

lib/core/src/model.rs
```diff
@@ -250,16 +261,39 @@ pub struct OnchainPaymentLimitsResponse {
 /// An argument when calling [crate::sdk::LiquidSdk::prepare_send_payment].
 #[derive(Debug, Serialize, Clone)]
 pub struct PrepareSendRequest {
-    pub invoice: String,
+    /// The destination we intend to pay to.
+    /// Currently supports BIP21 URIs, BOLT11 invoices and Liquid addresses
+    pub payment_destination: String,
```
roeierez, line 266: NIT: destination

lib/core/src/model.rs
```diff
@@ -250,16 +261,39 @@ pub struct OnchainPaymentLimitsResponse {
...
+
+    /// Should only be set when paying directly onchain or to a BIP21 URI
+    /// where no amount is specified
+    pub amount_sat: Option<u64>,
+}
+
+/// Specifies the supported destinations which can be payed by the SDK
+#[derive(Clone, Debug, Serialize)]
+pub enum SendDestination {
+    BIP21 {
```
roeierez, line 276: LiquidAddress instead of BIP21 to be consistent with the receive ?

lib/core/src/model.rs
```diff
@@ -250,16 +261,39 @@ pub struct OnchainPaymentLimitsResponse {
...
+    Bolt11 {
+        invoice: LNInvoice,
+    },
 }
 
 /// Returned when calling [crate::sdk::LiquidSdk::prepare_send_payment].
 #[derive(Debug, Serialize, Clone)]
 pub struct PrepareSendResponse {
-    pub invoice: String,
+    pub payment_destination: SendDestination,
```
roeierez, line 287: NIT: destination

lib/core/src/sdk.rs
```diff
@@ -676,23 +676,67 @@ impl LiquidSdk {
...
-            .amount_milli_satoshis()
-            .ok_or(PaymentError::AmountOutOfRange)?
-            / 1000;
-        let lbtc_pair = self.validate_submarine_pairs(receiver_amount_sat)?;
+        match sdk_common::input_parser::parse(&req.payment_destination).await? {
+            InputType::LiquidAddress {
+                address: mut liquid_address_data,
+            } => {
+                let Some(amount_sat) = req.amount_sat.or(liquid_address_data.amount_sat) else {
+                    return Err(PaymentError::AmountOutOfRange);
```
roeierez, line 688: Should it be AmountMissing ?

lib/core/src/sdk.rs
```diff
@@ -676,23 +676,67 @@ impl LiquidSdk {
...
             }
-            None => {
-                let lockup_fees_sat = self.estimate_lockup_tx_fee(receiver_amount_sat).await?;
-                lbtc_pair.fees.total(receiver_amount_sat) + lockup_fees_sat
+            InputType::Bolt11 { invoice } => {
+                self.ensure_send_is_not_self_transfer(&invoice.bolt11)?;
+                self.validate_invoice(&invoice.bolt11)?;
+
+                receiver_amount_sat =
+                    invoice.amount_msat.ok_or(PaymentError::AmountOutOfRange)? / 1000;
```
roeierez, line 712: Should it be AmountMissing ?

lib/core/src/sdk.rs
```diff
@@ -731,52 +775,102 @@ impl LiquidSdk {
...
+                    liquid_address_data.network == self.config.network.into(),
+                    PaymentError::Generic {
+                        err: "Cannot pay to the given network".to_string()
+                    }
+                );
 
-        let amount_sat = get_invoice_amount!(&req.invoice);
-        let payer_amount_sat = amount_sat + req.fees_sat;
+                let Some(amount_sat) = liquid_address_data.amount_sat else {
+                    return Err(PaymentError::AmountOutOfRange);
```
roeierez, line 799: AmountMissing ?

lib/core/src/sdk.rs
```diff
@@ -731,52 +775,102 @@ impl LiquidSdk {
...
+                ensure_sdk!(
+                    payer_amount_sat <= self.get_info().await?.balance_sat,
+                    PaymentError::InsufficientFunds
+                );
+
+                self.send_direct_payment(&liquid_address_data.address, amount_sat, *fees_sat)
+                    .await
+            }
+            SendDestination::Bolt11 { invoice } => {
+                self.check_send_payment_invoice(&invoice.bolt11, *fees_sat)
```
roeierez, line 812: NIT: `pay_invoice` ?

lib/core/src/sdk.rs
```diff
@@ -731,52 +775,102 @@ impl LiquidSdk {
...
+                    return Err(PaymentError::AmountOutOfRange);
+                };
+
+                let payer_amount_sat = amount_sat + fees_sat;
+                ensure_sdk!(
+                    payer_amount_sat <= self.get_info().await?.balance_sat,
+                    PaymentError::InsufficientFunds
+                );
+
+                self.send_direct_payment(&liquid_address_data.address, amount_sat, *fees_sat)
```
roeierez, line 808: NIT: `pay_liquid_address` ?

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -313,35 +325,47 @@ dictionary GetInfoResponse {
 };
 
 dictionary PrepareSendRequest {
-    string invoice;
+    string payment_destination;
```
roeierez, line 328: NIT: destination

roeierez (commented):
Looks good! Only final NIT and code generation missing.

roeierez (approved):
LGTM

dangeross (changes requested):
I still see the use of `prepare_pay_onchain` and `pay_onchain`, are they still needed?

Also looks like Flutter / React Native need regenerating

The `lib/bindings/src/lib.rs` file needs updating similarly to bindings.rs

cli/src/commands.rs
```diff
@@ -46,18 +56,19 @@ pub(crate) enum Command {
...
     ReceivePayment {
+        /// Whether or not to receive via lightning
+        #[arg(short = 'm', long = "method")]
+        payment_method: Option<PaymentMethod>,
+
         /// Amount the payer will send, in satoshi
-        payer_amount_sat: u64,
+        /// If not specified, it will generate a BIP21 URI/Liquid address with no amount
+        #[arg(short, long)]
+        amount_sat: Option<u64>,
```
ok300, line 66: Isn't this still the `payer_amount_sat`?

lib/core/src/bindings.rs
```diff
@@ -107,15 +107,15 @@ impl BindingLiquidSdk {
...
     ) -> Result<SendPaymentResponse, PaymentError> {
         self.sdk.send_payment(&req).await
     }
 
     pub async fn prepare_receive_payment(
         &self,
-        req: PrepareReceivePaymentRequest,
-    ) -> Result<PrepareReceivePaymentResponse, PaymentError> {
+        req: PrepareReceiveRequest,
+    ) -> Result<PrepareReceiveResponse, PaymentError> {
```
ok300, line 118: Why were these renamed? Doesn't this go against the convention of "method name + Request/Response"?

lib/core/src/model.rs
```diff
@@ -194,31 +194,42 @@ pub struct ConnectRequest {
     pub config: Config,
 }
 
+/// The receive methods supported by the SDK
```
ok300, line 197:
```suggestion
/// The send / receive methods supported by the SDK
```

This will probably be used for `Payment`, so it covers the send methods supported as well.

cli/src/commands.rs
```diff
@@ -46,18 +56,19 @@ pub(crate) enum Command {
     },
     /// Receive lbtc and send btc through a swap
     ReceivePayment {
+        /// Whether or not to receive via lightning
```
ok300, line 59:
```suggestion
        /// Indicates desired payment method. If not set, indicates payment via Lightning.
```

ok300 (commented):
Found a few NITs and have one bigger question:

The Liquid BIP21 URIs insist on having the asset ID if an amount is specified:
```
// "assetid" MUST be provided if "amount" is present
// See https://github.com/ElementsProject/elements/issues/805#issuecomment-576743532
```

However this PR adds send / receive support for
- BIP21 (with asset ID check)
- plain Liquid address with optional amount (without asset ID check)

Doesn't this mean we risk receiving other Liquid assets in the second case?

A simple solution would be to subsume "plain Liquid address" under BIP21, as there can be a BIP21 with only the address (and if amount is present, it automatically checks that asset ID is also present).

What do you think?

lib/core/src/bindings.rs
hydra-yse: Previously the methods were named `PrepareReceivePaymentReq/Res` and `PrepareReceiveOnchainReq/Res`, but the distinction is now not necessary as the two have been merged. This is now the same as send, where it's simply `PrepareSendReq/Res`

lib/core/src/sdk.rs
```diff
@@ -1526,12 +1725,19 @@ impl LiquidSdk {
     ///     * `redirect_url` - the optional redirect URL the provider should redirect to after purchase
```
dangeross, line 1725: Replace above in the comments `prepare_res` with `prepare_response`

packages/react-native/src/index.ts
```diff
@@ -19,29 +19,29 @@ const BreezSDKLiquid = NativeModules.RNBreezSDKLiquid
 
 const BreezSDKLiquidEmitter = new NativeEventEmitter(BreezSDKLiquid)
 
-export interface AesSuccessActionDataDecrypted {
+export type AesSuccessActionDataDecrypted = {
```
dangeross, line 22: Seems to be producing old syntax. What's your `tslint` and `typescript` versions?
```
tslint --version
tsc --version
```
I have:
```
tslint 6.1.3
tsc 5.4.5
```

lib/core/src/model.rs
```diff
@@ -951,19 +977,86 @@ pub struct PaymentSwapData {
...
     /// Payment status derived from the swap status
     pub status: PaymentState,
 }
 
+/// The specific details of a payment, depending on its type
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub enum PaymentDetails {
+    Lightning {
+        /// The swap ID, if any swap is associated with this payment
+        swap_id: String,
```
erdemyerebasmaz, line 994: Is the documentation wrong here, or the type? As "if any" means this field is nullable.

lib/core/src/model.rs
```diff
@@ -951,19 +977,86 @@ pub struct PaymentSwapData {
...
+    /// In case of an incoming chain swap, it's the Liquid address which will receive the funds
+    pub claim_address: Option<String>,
+
     /// Payment status derived from the swap status
     pub status: PaymentState,
 }
 
+/// The specific details of a payment, depending on its type
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub enum PaymentDetails {
```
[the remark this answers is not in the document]
erdemyerebasmaz: One benefit is, for some of the binding languages, you can't match directly on the enum and even if you did, you'll need to assign the `details` to a variable to access it's fields. Otherwise it will always be recognized as `PaymentDetails`.

packages/react-native/src/index.ts
erdemyerebasmaz:
```
tslint: 6.1.3
tsc: 5.5.4
```

erdemyerebasmaz: Unless I'm missing something CI used the exact same versions on my env, but running the codegen does not introduce changes for me.

```
tslint: 6.1.3
typescript: 5.5.4
kotlin: 2.0.10 
swiftformat: 0.54.3
ktlint: 1.3.1
```

erdemyerebasmaz: The difference was on `packages/react-native`. Running `make all` on `packages/react-native/` after fixing Payment on UDL file has addressed this & `cargo clippy` issues.

- [Update Payment on udl file](https://github.com/breez/breez-sdk-liquid/pull/414/commits/a7d43d8a77e7a2510981d5bdbd98b4d7d1fce14e)
- [chore: run 'make all' on 'packages/react-native/'](https://github.com/breez/breez-sdk-liquid/pull/414/commits/fe981e6cf435a9fa8d2daf1743a025b04673913c)

lib/core/src/model.rs
hydra-yse: Good catch Erdem, previously it was an `Option` but then I decided to leave it as string since we're sure that if the details exist so will the swap_id (in case of Bitcoin/Lightning details)

lib/core/src/sdk.rs
```diff
@@ -1140,32 +1277,63 @@ impl LiquidSdk {
...
+                debug!(
+                    "Preparing Liquid Receive Swap with: amount_sat {amount_sat:?} sat, fees_sat {fees_sat} sat"
+                );
+            }
+        };
 
-        Ok(PrepareReceivePaymentResponse {
-            payer_amount_sat,
+        Ok(PrepareReceiveResponse {
+            amount_sat,
```
ok300, line 1334: Why was this renamed from `payer_amount_sat`? AFAIK it still represents the payer amount.

lib/core/src/sdk.rs
```diff
@@ -1180,16 +1348,71 @@ impl LiquidSdk {
...
+
+                let receive_destination = match amount_sat {
+                    Some(amount_sat) => LiquidAddressData {
+                        address: address.to_string(),
+                        network: self.config.network.into(),
+                        amount_sat: Some(*amount_sat),
+                        asset_id: Some(match self.config.network {
+                            LiquidNetwork::Mainnet => AssetId::LIQUID_BTC.to_hex(),
+                            LiquidNetwork::Testnet => {
+                                "6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d"
```
ok300, line 1389: This is the same as `AssetId::LIQUID_BTC` :
https://github.com/ElementsProject/rust-elements/blob/f875ae2ac247a44ca1bbbc4aa16a3c7adbad6c4d/src/issuance.rs#L367-L370

This block should be replaced with
```rust
asset_id: Some(AssetId::LIQUID_BTC.to_hex()),
```

packages/flutter/ffigen.yaml
```diff
@@ -18,3 +18,5 @@ comments:
   style: any
   length: full
 ignore-source-errors: true
+llvm-path:
+  - '/nix/store/3x48dmkw2z70mscl68a5xhf9rpr57qw2-clang-18.1.7-lib/'
```
ok300, line 22: Committed by mistake?

lib/core/src/model.rs
```diff
@@ -912,6 +926,9 @@ pub struct PaymentTxData {
     /// The tx ID of the transaction
     pub tx_id: String,
 
+    /// The destination address of the payment, only present when paying directly to a Liquid address
+    pub destination_address: Option<String>,
```
ok300, line 930: IMO this is not clear enough.

- Does this work only for "pay to Liquid" type Send? All Receive payments are technically paying to a Liquid address, so they match the description.
- Does this also work for "Send to invoice" type Send, where we default to the MRH? In that case, we're also sending to a Liquid address.
- Does this also work for refund txs for failed Sends? In that case, the destination address is also a Liquid address.
- Since this is a struct that represents an onchain tx, what type of tx will this field be filled for? (lockup, claim, refund)

Please update the description accordingly.

lib/core/src/persist/migrations.rs
```diff
@@ -65,5 +65,6 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
         ALTER TABLE send_swaps ADD COLUMN description TEXT;
         ALTER TABLE chain_swaps ADD COLUMN description TEXT;
         ",
+        "ALTER TABLE payment_tx_data ADD COLUMN destination_address TEXT;",
```
ok300, line 68: From what I understand, this is only set for chain swaps, and it's set to the swaps `claim_address`. There is already a field called `claim_address` in the `chain_swaps` table.

Instead of inserting it here (duplicating it) in the table for tx data, why not update the `select_payment_query` to read the `claim_address` directly from `chain_swaps`?

ok300 (changes requested):

lib/core/src/sdk.rs
hydra-yse: Indeed, might get confused with the receiver amount. I'll rollback the change.

lib/core/src/model.rs
hydra-yse: I've thought up a better way to go about this, as I agree that using the `PaymentTxData` struct is misleading for other payment types.

lib/core/src/persist/migrations.rs
hydra-yse: That's what currently happens for chain swaps (we don't use the `PaymentTxData::destination_address` but just the `claim_address` column from the swap data. 
The reason for creating this is related only with direct onchain (Liquid) payments where we want to rebuild the destination within the `PaymentDetails`. I'll suggest an alternative solution, let me know if it makes more sense.

packages/react-native/src/index.ts
```diff
@@ -158,172 +158,172 @@ export type LnUrlPayRequestData = {
...
+export interface LnUrlWithdrawSuccessData {
     invoice: LnInvoice
 }
 
-export type LocaleOverrides = {
+export interface LocaleOverrides {
     locale: string
     spacing?: number
-    symbol: SymbolType
+    symbol: Symbolinterface
```
dangeross, line 187:
```suggestion
    symbol: SymbolType
```

packages/react-native/src/index.ts
```diff
@@ -158,172 +158,172 @@ export type LnUrlPayRequestData = {
...
 
-export type Payment = {
+export interface Payment {
     destination?: string
     txId?: string
     timestamp: number
     amountSat: number
     feesSat: number
-    paymentType: PaymentType
+    paymentinterface: PaymentType
```
dangeross, line 220:
```suggestion
    paymentType: PaymentType
```

packages/react-native/src/index.ts
```diff
@@ -52,34 +52,34 @@ export type Config = {
...
 
-export type CurrencyInfo = {
+export interface CurrencyInfo {
     name: string
     fractionSize: number
     spacing?: number
-    symbol?: SymbolType
-    uniqSymbol?: SymbolType
+    symbol?: Symbolinterface
+    uniqSymbol?: Symbolinterface
```
dangeross, line 65:
```suggestion
    symbol?: SymbolType
    uniqSymbol?: SymbolType
```

packages/react-native/src/index.ts
```diff
@@ -114,39 +114,39 @@ export type LiquidAddressData = {
     message?: string
 }
 
-export type ListPaymentsRequest = {
-    filters?: PaymentType[]
+export interface ListPaymentsRequest {
+    filters?: Paymentinterface[]
```
dangeross, line 118:
```suggestion
    filters?: PaymentType[]
```

dangeross (changes requested):

lib/core/src/sdk.rs
```diff
@@ -537,14 +543,26 @@ impl LiquidSdk {
...
                         confirmed_sent_sat += p.amount_sat;
-                        confirmed_received_sat += p.refund_tx_amount_sat.unwrap_or_default();
+                        if let Some(details) = p.details {
+                            confirmed_received_sat +=
+                                details.get_refund_tx_amount_sat().unwrap_or_default();
+                        }
                     }
-                    Pending => match p.refund_tx_amount_sat {
-                        Some(refund_tx_amount_sat) => {
+                    Pending => match p.details {
```
ok300, line 551: How about rewriting this as

```rust
Pending => match p.details.and_then(|d| d.get_refund_tx_amount_sat()) {
    Some(refund_tx_amount_sat) => {
        confirmed_sent_sat += p.amount_sat;
        pending_receive_sat += refund_tx_amount_sat;
    }
    None => pending_send_sat += p.amount_sat,
},
```

Looks simpler.

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -313,35 +325,53 @@ dictionary GetInfoResponse {
...
+enum PaymentMethod {
+    "Lightning",
+    "BitcoinAddress",
+    "LiquidAddress",
 };
 
-dictionary PrepareReceivePaymentResponse {
-    u64 payer_amount_sat;
+dictionary PrepareReceiveRequest {
+    u64? amount_sat;
```
ok300, line 358: Why was this renamed from `payer_amount_sat`? If that's still what it represents, it should remain unchanged IMO.

lib/core/src/persist/mod.rs
```diff
@@ -99,14 +103,30 @@ impl Persister {
...
                 ptx.is_confirmed,
             ),
         )?;
+        if let Some(PaymentDetails::Liquid {
+            destination,
+            description,
+        }) = details
+        {
+            tx.execute(
+                "INSERT OR REPLACE INTO payment_details (
```
ok300, line 120: If the `payment_details` table only contains payment details for direct Liquid txs, shouldn't it be given a clearer name? Like `direct_tx_details` , `direct_onchain_tx_details` or similar?

lib/core/src/sdk.rs
```diff
@@ -1099,35 +1252,36 @@ impl LiquidSdk {
...
                             }
-                            false => {
-                                debug!("Received Send Payment pending event, waiting for confirmation");
-                                maybe_payment = Some(details);
+                        },
+                    Ok(SdkEvent::PaymentSucceeded { details: payment }) =>
+                        if let Some(PaymentDetails::Lightning { swap_id, .. } | PaymentDetails::Bitcoin { swap_id, .. }) = &payment.details {
+                            if swap_id.clone() == expected_swap_id {
+                                debug!("Received Send Payment succeed event");
+                                return Ok(payment);
```
ok300, line 1282:
```suggestion
                        let maybe_payment_swap_id = payment.details.as_ref().and_then(|d|d.get_swap_id());
                        if matches!(maybe_payment_swap_id, Some(swap_id) if swap_id == expected_swap_id) {
                            debug!("Received Send Payment succeed event");
                            return Ok(payment);
```

lib/core/src/sdk.rs
```diff
@@ -1099,35 +1252,36 @@ impl LiquidSdk {
...
                         return Err(PaymentError::PaymentTimeout)
                     },
                 },
                 event = events_stream.recv() => match event {
-                    Ok(SdkEvent::PaymentPending { details }) => match details.swap_id.clone() {
-                        Some(id) if id == swap_id => match accept_zero_conf {
-                            true => {
-                                debug!("Received Send Payment pending event with zero-conf accepted");
-                                return Ok(details)
+                    Ok(SdkEvent::PaymentPending { details: payment }) =>
```
ok300, line 1263: How about rewriting this branch as
```rust
Ok(SdkEvent::PaymentPending { details: payment }) => {
    let maybe_payment_swap_id = payment.details.as_ref().and_then(|d|d.get_swap_id());
    if matches!(maybe_payment_swap_id, Some(swap_id) if swap_id == expected_swap_id) {
        match accept_zero_conf {
            true => {
                debug!("Received Send Payment pending event with zero-conf accepted");
                return Ok(payment)
            }
            false => {
                debug!("Received Send Payment pending event, waiting for confirmation");
                maybe_payment = Some(payment);
            }
        }
    };
}
```

Less nested blocks (so less indentation) and IMO simpler to read.

hydra-yse: Agreed, it does look simpler to read

lib/core/src/model.rs
```diff
@@ -951,19 +977,94 @@ pub struct PaymentSwapData {
...
+
 /// Represents an SDK payment.
 ///
 /// By default, this is an onchain tx. It may represent a swap, if swap metadata is available.
 #[derive(Debug, Clone, PartialEq, Serialize)]
 pub struct Payment {
-    pub tx_id: Option<String>,
+    /// The destination associated with the payment, if it was created via our SDK.
+    /// Can be either a BIP21 URI (Liquid/Bitcoin), a Liquid address or an invoice
+    pub destination: Option<String>,
```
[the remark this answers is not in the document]
hydra-yse: Right, since we're utilizing the `claim_address`, meaning we actually never use Bitcoin BIP21.

lib/core/src/persist/mod.rs
ok300: I see, makes sense.

lib/core/src/sdk.rs
ok300: NIT: would improve readability and remove nested blocks.

ok300 (approved):
Just one NIT, otherwise LGTM.

dangeross:
Seems to be conflicts

lib/core/Cargo.toml
```diff
@@ -36,11 +38,15 @@ openssl = { version = "0.10", features = ["vendored"] }
...
-futures-util = { version = "0.3.28", default-features = false, features = ["sink", "std"] }
+futures-util = { version = "0.3.28", default-features = false, features = [
+  "sink",
+  "std",
+] }
 async-trait = "0.1.80"
 hex = "0.4"
 reqwest = { version = "=0.11.20", features = ["json"] }
 electrum-client = { version = "0.19.0" }
+querystring = "1.1.0"
```
dangeross, line 49: Is this used still?

dangeross (approved):
LGTM. Just need to check `querystring` is still needed and resolve conflicts

packages/react-native/src/index.ts
```diff
@@ -545,7 +591,6 @@ export const connect = async (req: ConnectRequest): Promise<void> => {
 export const addEventListener = async (listener: EventListener): Promise<string> => {
     const response = await BreezSDKLiquid.addEventListener()
     BreezSDKLiquidEmitter.addListener(`event-${response}`, listener)
-
```
dangeross, line 548: This is causing CI issues

erdemyerebasmaz (approved):
LGTM

Have some nit comments and a note on a calculation which needs to be tested for floating point precision errors.

PR #424, opened 2024-07-27
Pause event notifications on first sync

ok300:
This PR temporarily disables the emitting of events during the first `sync`.

Fixes #300

PR #426, opened 2024-08-01
Update Dart project depedencies to latest

erdemyerebasmaz:
Ran `melos pub-upgrade` to address CI errors.

We should automate this;
- Run `melos pub-upgrade` and create a PR if there's an error on `Check git status` step of  `Check Dart/Flutter bindings` job because of changes in `pubspec.lock` files.


erdemyerebasmaz:
Fixed as part of #424. Closing, but my point stands, this process should be automated.

PR #433, opened 2024-08-06
Android notification plugin

dangeross:
This PR adds the Android notification plugin to handle lnurlpay_info, lnurlpay_invoice and swap_updated push notifications.

Builds on PR #410

### DIFF
--- a/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/NotificationHelper.kt
+++ b/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/NotificationHelper.kt
@@ -1,0 +1,364 @@
+package breez_sdk_liquid_notification
+
+import android.Manifest
+import android.annotation.SuppressLint
+import android.app.Notification
+import android.app.NotificationChannel
+import android.app.NotificationChannelGroup
+import android.app.NotificationManager
+import android.app.PendingIntent
+import android.app.TaskStackBuilder
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.os.Build
+import android.util.Log
+import androidx.annotation.RequiresApi
+import androidx.core.app.ActivityCompat
+import androidx.core.app.NotificationCompat
+import androidx.core.app.NotificationManagerCompat
+import breez_sdk_liquid_notification.Constants.DEFAULT_FOREGROUND_SERVICE_NOTIFICATION_CHANNEL_DESCRIPTION
+import breez_sdk_liquid_notification.Constants.DEFAULT_FOREGROUND_SERVICE_NOTIFICATION_CHANNEL_NAME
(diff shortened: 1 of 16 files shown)


roeierez (commented):
Looks very good! Added a small comment and I think we should wait for https://github.com/breez/breez-sdk-liquid/pull/414 before merging as it will change slightly the receive interface

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/NotificationHelper.kt
```diff
@@ -0,0 +1,364 @@
...
+
+            getNotificationManager(context)?.also { manager ->
+                createNotificationChannelGroups(context, manager)
+                createNotificationChannels(context, manager)
+                Log.d(TAG, "Registered notification channels")
+            }
+        }
+
+        @RequiresApi(Build.VERSION_CODES.O)
+        private fun createNotificationChannels(
```
erdemyerebasmaz, line 134: Reviewing this again made me aware of parts we could improve. I really want to refactor this function and make it more modular :)

Which poses a question: How can we ensure changes made in this plugin are also reflected in the GL notification plugin, and vice versa?

erdemyerebasmaz (commented):
Changes look good to me!

I agree with @roeierez's sentiment to wait for https://github.com/breez/breez-sdk-liquid/pull/414 before merging, as it has breaking changes on how to address his suggestion:
> Should we compare to the notified swap id before notifying success?

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/NotificationHelper.kt
dangeross: I agree, there is overlap. I think we agreed in the dev call to keep it separate to begin with, then work together a more modular approach. I would suggest a separate repo for the core plugin components

roeierez: Let's follow our decision to first implement a workable solution then we will better understand on how to proceed.

roeierez (approved):
LGTM

erdemyerebasmaz (approved):

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
```diff
@@ -49,7 +50,7 @@ class SwapUpdatedJob(
             is SdkEvent.PaymentSucceeded -> {
                 val payment = e.details
 
-                payment.swapId?.let { swapId ->
+                (payment.details as? PaymentDetails.Bitcoin)?.swapId?.let { swapId ->
```
dangeross, line 53: This should handle both `Lightning` and `Bitcoin` payment details

PR #428, opened 2024-08-01
Generate bindings for sdk-common w/ liquid feature

erdemyerebasmaz:

PR #434, opened 2024-08-07
Allow working with local library on iOS

erdemyerebasmaz:
This PR allows working with local library on iOS by using a different `podspec` for local & production environments.

###  iOS changes:
- Create a `podspec` for local development 
 - Use production `podspec` when publishing
---
### Other changelist:
These changes addresses CI warnings & a bug when running `just codegen` for the first time.
- ####  Flutter 3.24 changes
  - Update SDK ranges
    - Dart >= `3.5.0`
    - Flutter >= `3.24.0`
  - Update dependencies to latest using `melos pub-upgrade`

- ####  Bugfix:
  - Correct the output folder creation on `codegen` just recipe
    - This would have caused an error on devices that would run `just codegen` for the first time, as the output folder wouldn't have existed

### DIFF
--- a/packages/flutter/ios/flutter_breez_liquid.podspec
+++ b/packages/flutter/ios/flutter_breez_liquid.podspec
@@ -34,7 +34,6 @@ Pod::Spec.new do |spec|
   spec.osx.deployment_target = '10.11'
 
   spec.dependency 'Flutter'
-  spec.platform = :ios, '11.0'
   spec.static_framework = true
 
   # Flutter.framework does not contain a i386 slice.
--- a/packages/flutter/ios/flutter_breez_liquid.podspec.production
+++ b/packages/flutter/ios/flutter_breez_liquid.podspec.production
@@ -1,4 +1,4 @@
-version = '0.2.1' # generated; do not edit
+version = '0.1.0' # generated; do not edit
 tag_name = "v#{version}"
 release_tag_name = "breez_liquid-#{tag_name}"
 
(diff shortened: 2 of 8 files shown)


erdemyerebasmaz:
CI errors can be ignored, `cargo clippy` warnings will be addressed by 
- #428

or these can be addressed separately on top of `main` branch.

packages/flutter/ios/flutter_breez_liquid.podspec
```diff
@@ -25,16 +26,18 @@ Pod::Spec.new do |spec|
...
-
     unzip #{local_zip_name}
     cd -
   CMD
   
   spec.ios.deployment_target = '12.0'
   spec.osx.deployment_target = '10.11'
+
+  spec.dependency 'Flutter'
+  spec.platform = :ios, '11.0'
```
dangeross, line 37: This is not needed because the deployment_targets are set

https://guides.cocoapods.org/syntax/podspec.html#platform
```
  spec.ios.deployment_target = '12.0'
  spec.osx.deployment_target = '10.11'
```

packages/flutter/ios/flutter_breez_liquid.podspec.production
dangeross, line 1: Does this need to be included in the production podspec also? It seems to be in the breez-sdk-greenlight production podspec
```
  spec.dependency 'Flutter'

  # Flutter.framework does not contain a i386 slice.
  spec.pod_target_xcconfig = {'STRIP_STYLE' => 'non-global', 'DEFINES_MODULE' => 'YES', 'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'i386' }
  spec.swift_version = '5.0'
```

dangeross (changes requested):

packages/flutter/ios/flutter_breez_liquid.podspec
erdemyerebasmaz: Removed f79d2122613116aacc7f9332520c8d1a95becc60

packages/flutter/ios/flutter_breez_liquid.podspec.production
erdemyerebasmaz: No, it's not needed and it could be removed from breez-sdk-greenlight production podspec as well.

This workaround is to ensure library works with non-physical devices, i.e.; iOS Simulator.

dangeross (approved):

PR #440, opened 2024-08-09
Add description hash to receive payment

dangeross:
This PR adds support for using a description hash in the create reverse swap request. This can then be used in the LNURL-pay flow to hash the metadata. 

Even though there is broad adoption to drop the validation of the metadata description hash, it is as yet not merged: https://github.com/lnurl/luds/pull/234

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -500,7 +500,7 @@ dependencies = [
 [[package]]
 name = "boltz-client"
 version = "0.1.3"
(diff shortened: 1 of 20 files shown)


ok300 (changes requested):
Calling `receive_payment` in the CLI panics:

```
receive-payment 60000
thread 'main' panicked at ...
Command receive-payment: Short option names must be unique for each argument, but '-h' is in use by both 'use_description_hash' and 'help' (call `cmd.disable_help_flag(true)` to remove the auto-generated `--help`)
```

ok300 (approved):

roeierez (approved):

PR #443, opened 2024-08-13
Add `sign_message` and `check_message`

### LINKED ISSUE #277: Receive via lightning address

dangeross:
This PR adds message signing and checking methods to help sign requests to be sent to the LNURL service

Fixes #277

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -569,6 +569,7 @@ dependencies = [
  "tokio-stream",
  "tokio-tungstenite",
  "url",
+ "zbase32",
 ]
 
 [[package]]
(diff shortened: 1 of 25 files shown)


ok300:
There is some overlap with `sign_message` / `check_message` in the main SDK, including the use of `zbase32`.

Would it make sense to expose the common methods in `sdk-common` and re-use them here, as much as possible?

dangeross:
@ok300 I don't think there's enough overlap to justify moving it. The GL SDK uses the node to sign the message. I have updated the PR to use the lightning message_signing module to verify like in the GL SDK

ok300 (approved):
You're right.

roeierez (approved):

PR #447, opened 2024-08-14
Bugfix: Update production podpsec's version with version script

### LINKED ISSUE #417: Run version script for Flutter after bumping versions
[lib/bindings/langs/flutter/scripts/version.sh](https://github.com/breez/breez-sdk-liquid/blob/main/lib/bindings/langs/flutter/scripts/version.sh) script should be run after version bumping to ensure applications using our Flutter plugin can locate and extract binary zip files properly.

To prevent issues from not running this script prior release, I suggest we include version bumping & run the `version.sh` script as part of a workflow job and make it a pre-requisite for release/publishing.

erdemyerebasmaz:
Closes #417

This PR addresses a regression caused with changes on 
- #434

and makes several improvements to version script. Making it easier to use with the addition of `just version` recipe & correctly sets `ROOT` relative to the script's location.

### Changelist:
- fix: Update production podpsec's version with version script
- Resolve relative path of root from script directory
- Ran 'just version'
- Ran `melos pub-upgrade`
- Updated melos dependency
- Update versions on podspec & cmake scripts when publishing Flutter plugin
- Check if versions on podspec & cmake scripts were updated using version script on CI

### DIFF
--- a/.github/workflows/publish-flutter.yml
+++ b/.github/workflows/publish-flutter.yml
@@ -92,6 +92,10 @@ jobs:
           rm android/build.gradle.bak
           rm android/CMakeLists.txt.bak
           rm ios/flutter_breez_liquid.podspec.bak
+      
+      - name: Update version number on podspec files & CMake scripts
+        working-directory: build/lib/bindings/langs/flutter/
+        run: bash scripts/version.sh
 
       - name: Archive flutter release
         uses: actions/upload-artifact@v4
--- a/lib/bindings/langs/flutter/justfile
+++ b/lib/bindings/langs/flutter/justfile
@@ -86,6 +86,23 @@ link-other:
 	-ln -sf $(pwd)/platform-build/other.tar.gz ../../../../packages/flutter/linux/{{curr_version}}.tar.gz
 	-ln -sf $(pwd)/platform-build/other.tar.gz ../../../../packages/flutter/windows/{{curr_version}}.tar.gz
 
+# Copies library archives from platform-build to their expected locations
+copy:
+	just copy-apple
+	just copy-android
+	just copy-other
+
+copy-apple:
+	cp $(pwd)/platform-build/breez_sdk_liquid.xcframework.zip ../../../../packages/flutter/macos/Frameworks/{{curr_version}}.zip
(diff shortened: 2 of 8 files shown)


lib/bindings/langs/flutter/justfile
```diff
@@ -87,6 +87,10 @@ init *args:
 init-sdk:
 	brew install protobuf
 
+# Update version number on podspec files & CMake scripts
+version:
```
dangeross, line 91: Should this be run during the publish build? I imagine the pubspec version being updated but `just version` not run. I don't see the need to keep commiting the podspec.production with version changes

erdemyerebasmaz: > Should this be run during the publish build? I imagine the pubspec version being updated but `just version` not run. I don't see the need to keep commiting the podspec.production with version changes

Yes, I was preparing a separate PR for these changes.

I am also adding a step to main CI to make sure version on podspec files & CMake scripts were updated using this recipe.

A little note: This script doesn't cover the case where version on Flutter plugin's `pubspec.yaml`  wasn't updated properly, which is used as the source version number on this script.

erdemyerebasmaz: @dangeross Done.
- Update versions on podspec & cmake scripts on publish d35ac7e
- Check if versions on podspec & cmake scripts were updated properly on CI 19512d1

.github/workflows/publish-flutter.yml
```diff
@@ -92,6 +92,10 @@ jobs:
           rm android/build.gradle.bak
           rm android/CMakeLists.txt.bak
           rm ios/flutter_breez_liquid.podspec.bak
+      
+      - name: Update version number on podspec files & CMake scripts
+        working-directory: dist
```
dangeross, line 97: Is there a reason not to set the working dir here?
```
        working-directory: build/lib/bindings/langs/flutter
```

erdemyerebasmaz: > Is there a reason not to set the working dir here?
> 
> ```
>         working-directory: build/lib/bindings/langs/flutter
> ```

No, fixed: 26b8f83

dangeross (approved):

PR #448, opened 2024-08-15
Receive payment: remove MRH BIP21 amount validation

ok300:
This PR removes the amount validation of the BIP21 in the magic routing hint.


ok300:
Fixes https://github.com/breez/misty-breez/issues/142

dangeross (approved):

PR #450, opened 2024-08-15
Confirm chain swap claims

dangeross:
This PR changes the chain swap so that once the claim tx is broadcast it keeps the state pending (waiting confirmation event is emitted) and waits for confirmation before setting it to complete. For the incoming chain swap the `sync()` monitors the claim tx, for outgoing chain swap the claim address is checked periodically to see if the tx is confirmed.

See: https://github.com/breez/breez-sdk-liquid/pull/399#discussion_r1716713556

### DIFF
--- a/lib/core/src/chain_swap.rs
+++ b/lib/core/src/chain_swap.rs
@@ -4,7 +4,7 @@ use std::{str::FromStr, sync::Arc};
 use anyhow::{anyhow, Result};
 use boltz_client::swaps::boltz::{self, SwapUpdateTxDetails};
 use boltz_client::swaps::{boltz::ChainSwapStates, boltz::CreateChainResponse};
-use boltz_client::Secp256k1;
+use boltz_client::{Address, Secp256k1, ToHex};
 use log::{debug, error, info, warn};
 use lwk_wollet::elements::Transaction;
 use tokio::sync::{broadcast, watch, Mutex};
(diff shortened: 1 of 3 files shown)


lib/core/src/persist/chain.rs
```diff
@@ -221,6 +221,23 @@ impl Persister {
         Ok(res)
     }
 
+    pub(crate) fn list_pending_chain_swaps_by_claim_tx_id(
```
ok300, line 224:
```suggestion
    /// This only returns the swaps that have a claim tx, skipping the pending ones that are being refunded.
    pub(crate) fn list_pending_chain_swaps_by_claim_tx_id(
```

ok300 (commented):
Looks good, just a NIT or two:
- in the changed tests, there are comments just above the changes which talk about expecting a transition to `Complete` (L2336 and L2306). They should be changed to mention `Pending`.

ok300 (approved):

lib/core/src/chain_swap.rs
```diff
@@ -177,6 +180,54 @@ impl ChainSwapStateHandler {
...
+            })
+            .collect();
+        info!(
+            "Rescanning {} outgoing Chain Swap(s) at height {}",
+            chain_swaps.len(),
+            current_height
+        );
+        for swap in chain_swaps {
+            if let Err(e) = self.rescan_outgoing_chain_swap(&swap).await {
+                error!("Error rescanning incoming Chain Swap {}: {e:?}", swap.id);
```
erdemyerebasmaz, line 202: Typo
```suggestion
                error!("Error rescanning outgoing Chain Swap {}: {e:?}", swap.id);
```

lib/core/src/chain_swap.rs
```diff
@@ -177,6 +180,54 @@ impl ChainSwapStateHandler {
...
+            .bitcoin_chain_service
+            .lock()
+            .await
+            .get_script_history(script_pubkey.as_script())?;
+        let claim_tx_history = script_history
+            .iter()
+            .find(|h| h.txid.to_hex().eq(&claim_tx_id) && h.height > 0);
+        if claim_tx_history.is_some() {
+            info!(
+                "Incoming Chain Swap {} claim tx is confirmed. Setting the swap to Complete",
```
erdemyerebasmaz, line 222: Typo
```suggestion
                "Outgoing Chain Swap {} claim tx is confirmed. Setting the swap to Complete",
```

erdemyerebasmaz (approved):
Looks good.

Found two typos in log statements.

PR #454, opened 2024-08-26
Send: Consider request amount when preparing the payment

hydra-yse:


lib/core/src/sdk.rs
```diff
@@ -701,7 +701,7 @@ impl LiquidSdk {
             InputType::LiquidAddress {
                 address: mut liquid_address_data,
             } => {
-                let Some(amount_sat) = liquid_address_data.amount_sat else {
+                let Some(amount_sat) = liquid_address_data.amount_sat.or(req.amount_sat) else {
```
roeierez, line 704: Shouldn't we give more priority to the amount the user put in the request rather than in the bip21 input?

roeierez (changes requested):

lib/core/src/sdk.rs
roeierez: I think the same as a user can pay different amount in lightning for an bolt11 invoice that has a specific amount, we should allow here the user to override the bip21 amount (as nothing can enforce that amount really) and pay different amount. 
So IMO the amount in the request (if exists) takes priority over the amount in the bip21 as it was explicitly set by the user.

roeierez (approved):

PR #464, opened 2024-08-30
prepare-pay-onchain: add option for drain

### LINKED ISSUE #211: Add support for `drain` (send all L-BTC)

ok300:
Fixes #211

Open questions
- [ ] When locking up funds in `ChainSwapStateHandler`: how to tell if we're dealing with drain or normal lockup?
- [ ] Should chain swap non-drain lockup tx use the standard feerate too, just like the drain tx?

Open tasks
- [ ] Calling `prepare_pay_onchain` with the right `receiver_amount_sat` is essentially a drain, even if it's called with `drain=false`. The `prepare_pay_onchain` and `pay_onchain` should recognize and handle that correctly.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -50,9 +50,12 @@ pub(crate) enum Command {
         /// Amount that will be received, in satoshi
         receiver_amount_sat: u64,
 
-        // The optional fee rate to use, in satoshi/vbyte
+        /// The optional fee rate to use, in satoshi/vbyte
         #[clap(short = 'f', long = "fee_rate")]
         sat_per_vbyte: Option<u32>,
+
+        #[arg(short, long)]
+        drain: Option<bool>,
     },
     /// Receive lbtc and send btc through a swap
     ReceivePayment {
@@ -342,11 +345,14 @@ pub(crate) async fn handle_command(
             address,
             receiver_amount_sat,
             sat_per_vbyte,
+            drain,
         } => {
+            let drain = drain.unwrap_or_default();
             let prepare_response = sdk
                 .prepare_pay_onchain(&PreparePayOnchainRequest {
                     receiver_amount_sat,
                     sat_per_vbyte,
+                    drain,
                 })
                 .await?;
 
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -413,6 +413,7 @@ dictionary OnchainPaymentLimitsResponse {
 dictionary PreparePayOnchainRequest {
     u64 receiver_amount_sat;
     u32? sat_per_vbyte = null;
+    boolean drain;
 };
 
 dictionary PreparePayOnchainResponse {
--- a/lib/core/src/chain_swap.rs
+++ b/lib/core/src/chain_swap.rs
@@ -582,14 +582,24 @@ impl ChainSwapStateHandler {
             lockup_details.amount, lockup_details.lockup_address
         );
 
-        let lockup_tx = self
-            .onchain_wallet
-            .build_tx(
-                None,
-                &lockup_details.lockup_address,
-                lockup_details.amount as u64,
-            )
-            .await?;
+        // TODO How to check if this is a drain? We'd need to know balance_sat, but that doesn't belong in ChainSwapStateHandler
+        let is_drain = true;
+        let lockup_tx = match is_drain {
+            true => {
+                self.onchain_wallet
+                    .build_drain_tx(None, &lockup_details.lockup_address)
+                    .await?
+            }
+            false => {
+                self.onchain_wallet
+                    .build_tx(
+                        None,
+                        &lockup_details.lockup_address,
+                        lockup_details.amount as u64,
+                    )
+                    .await?
+            }
+        };
 
         let lockup_tx_id = self
             .liquid_chain_service
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -310,8 +310,13 @@ pub struct SendPaymentResponse {
 /// An argument when calling [crate::sdk::LiquidSdk::prepare_pay_onchain].
 #[derive(Debug, Serialize, Clone)]
 pub struct PreparePayOnchainRequest {
+    /// The amount in satoshi that will be received
     pub receiver_amount_sat: u64,
+    /// The optional fee rate of the Bitcoin claim transaction. Defaults to the swapper estimated claim fee.
     pub sat_per_vbyte: Option<u32>,
+    /// If set to true, the chosen `receiver_amount_sat` will be ignored. Instead, amounts and fees
+    /// will be returned such that all the wallet's funds are used.
+    pub drain: bool,
 }
 
 /// Returned when calling [crate::sdk::LiquidSdk::prepare_pay_onchain].
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -688,13 +688,17 @@ impl LiquidSdk {
             .sum())
     }
 
-    async fn estimate_lockup_tx_fee(&self, amount_sat: u64) -> Result<u64, PaymentError> {
+    fn get_temp_p2tr_addr(&self) -> &str {
         // TODO Replace this with own address when LWK supports taproot
         //  https://github.com/Blockstream/lwk/issues/31
-        let temp_p2tr_addr = match self.config.network {
+        match self.config.network {
             LiquidNetwork::Mainnet => "lq1pqvzxvqhrf54dd4sny4cag7497pe38252qefk46t92frs7us8r80ja9ha8r5me09nn22m4tmdqp5p4wafq3s59cql3v9n45t5trwtxrmxfsyxjnstkctj",
             LiquidNetwork::Testnet => "tlq1pq0wqu32e2xacxeyps22x8gjre4qk3u6r70pj4r62hzczxeyz8x3yxucrpn79zy28plc4x37aaf33kwt6dz2nn6gtkya6h02mwpzy4eh69zzexq7cf5y5"
-        };
+        }
+    }
+
+    async fn estimate_lockup_tx_fee(&self, amount_sat: u64) -> Result<u64, PaymentError> {
+        let temp_p2tr_addr = self.get_temp_p2tr_addr();
 
         self.estimate_onchain_tx_fee(
             amount_sat,
@@ -704,6 +708,25 @@ impl LiquidSdk {
                 .map(|v| v as f32),
         )
         .await
+    }
+
+    async fn estimate_drain_tx_fee(&self) -> Result<u64, PaymentError> {
+        let temp_p2tr_addr = self.get_temp_p2tr_addr();
+
+        let fee_sat = self
+            .onchain_wallet
+            .build_drain_tx(
+                // TODO If this is None (defaulting to standard fee of 0.1),
+                //  shouldn't it be None for estimate_onchain_tx_fee() too? (chain swap non-drain lockup fee)
+                None,
+                temp_p2tr_addr,
+            )
+            .await?
+            .all_fees()
+            .values()
+            .sum();
+
+        Ok(fee_sat)
     }
 
     /// Prepares to pay a Lightning invoice via a submarine swap.
@@ -1132,19 +1155,13 @@ impl LiquidSdk {
     }
 
     /// Prepares to pay to a Bitcoin address via a chain swap.
-    ///
-    /// # Arguments
-    ///
-    /// * `req` - the [PreparePayOnchainRequest] containing:
-    ///     * `receiver_amount_sat` - the amount in satoshi that will be received
-    ///     * `sat_per_vbyte` - the optional fee rate of the Bitcoin claim transaction. Defaults to the swapper estimated claim fee
     pub async fn prepare_pay_onchain(
         &self,
         req: &PreparePayOnchainRequest,
     ) -> Result<PreparePayOnchainResponse, PaymentError> {
         self.ensure_is_started().await?;
 
-        let receiver_amount_sat = req.receiver_amount_sat;
+        let balance_sat = self.get_info().await?.balance_sat;
         let pair = self.get_chain_pair(Direction::Outgoing)?;
         let claim_fees_sat = match req.sat_per_vbyte {
             Some(sat_per_vbyte) => ESTIMATED_BTC_CLAIM_TX_VSIZE * sat_per_vbyte as u64,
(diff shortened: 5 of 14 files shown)


lib/core/src/model.rs
```diff
@@ -310,7 +310,12 @@ pub struct SendPaymentResponse {
 /// An argument when calling [crate::sdk::LiquidSdk::prepare_pay_onchain].
 #[derive(Debug, Serialize, Clone)]
 pub struct PreparePayOnchainRequest {
+    /// The amount in satoshi that will be received
     pub receiver_amount_sat: u64,
+    /// If set to true, the chosen `receiver_amount_sat` will be ignored. Instead, amounts and fees
+    /// will be returned such that all the wallet's funds are used.
+    pub drain: bool,
```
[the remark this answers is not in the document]
ok300: Done: 6c0866d2c53c5457075f4d7c69563c5beebbdf76

roeierez (changes requested):

lib/core/src/sdk.rs
```diff
@@ -1204,20 +1269,33 @@ impl LiquidSdk {
...
+            .ceil() as u64;
+        let boltz_fee_sat = user_lockup_amount_sat - user_lockup_amount_sat_without_service_fee;
         self.validate_user_lockup_amount_for_chain_pair(&pair, user_lockup_amount_sat)?;
-        let lockup_fees_sat = self.estimate_lockup_tx_fee(user_lockup_amount_sat).await?;
+
+        let payer_amount_sat = req.prepare_response.total_fees_sat + receiver_amount_sat;
+
+        let lockup_fees_sat = match payer_amount_sat == balance_sat {
+            true => self.estimate_drain_tx_fee().await?,
+            false => {
```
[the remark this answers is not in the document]
ok300: You're right:
- if `payer_amount == balance`, then it's a perfect drain (all good)
- if `payer_amount < balance` with a difference of 1-200 sats, then `estimate_lockup_tx_fee_chain_send` fails with insufficient funds
- if `payer_amount < balance` with a bigger difference (> 200 sat), then `estimate_lockup_tx_fee_chain_send` succeeds

It probably depends on the tx size: when we expect change, we have an extra output. When it's a `drain`, there is no change output. So the tx size is bigger in the first case, which is why the minimum fee (and minimum spendable amount) is higher than in the case of `drain`.

What this means for the user:
- if they choose `drain = true`, it will work
- if they choose `drain = false`, but a `receiver_amount` high enough that the `payer_amount` gets very close to the `balance` (within 1-200 sat), `prepare_pay_onchain` will fail with insufficient funds
- if they choose `drain = false`, but a `receiver_amount` of any value below that critical threshold, it will work

IMO this is reasonable. If the user wants to send everything, they should use `drain`.

What do you think?

lib/core/src/chain_swap.rs
```diff
@@ -582,14 +582,24 @@ impl ChainSwapStateHandler {
...
                 &lockup_details.lockup_address,
                 lockup_details.amount as u64,
             )
-            .await?;
+            .await
+        {
+            Err(PaymentError::InsufficientFunds) => {
+                warn!("Cannot build normal lockup tx due to insufficient funds, attempting to build drain tx");
+                self.onchain_wallet
+                    .build_drain_tx(None, &lockup_details.lockup_address)
```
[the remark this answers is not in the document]
ok300: Done: 1fa49259d1416138b7ee4f8fb29844dda4d2e8d4

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -413,6 +413,7 @@ dictionary OnchainPaymentLimitsResponse {
 dictionary PreparePayOnchainRequest {
     u64 receiver_amount_sat;
     u32? sat_per_vbyte = null;
+    boolean? drain = null;
```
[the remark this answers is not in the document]
ok300: To summarize the options available to the user:
- to use `drain` caller has to
  - set `drain = true` and not set `receiver_amount_sat`, or
  - set `drain = false or null` and set `receiver_amount_sat` to the amount that would spend everything
- to not use `drain` caller has to
  - set `drain = false or null` and set `receiver_amount_sat`
 
IMO that's overly complex.

To simplify, how about the following:
- remove `drain` flag and make `receiver_amount_sat` optional
- if it's set, it is non-drain (normal usage)
- if it's not set, it's considered `drain`

What do you think, would that be a good option?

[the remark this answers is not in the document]
roeierez: I think treating null amount as drain is not intuitive.

What do you think about using an  enum?

```
Enum SendPaymentAmount {
 ReceiverAmount(amount: int),
 Drain,
}
```

This also gives us the option to add another variant `Sender(int)` in the future.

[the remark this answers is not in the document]
ok300: Good idea, done: 38f328e35ae550677159ddbcd638e9acbe9ec7f5

lib/core/src/sdk.rs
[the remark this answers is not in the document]
ok300: Finding the drain fee will find the max amount the user can pay, but only as a discrete amount (C). Anything before it (in the interval `(B, C)`) will throw "insufficient funds" error.

```
Amounts user can pay (payer_amount)

   A                B    C
---++++++++++++++++++----+----

A = smallest payer_amount = chain swap min + lockup fee
B = C - 200 sat (approx)
C = balance = biggest payer_amount = max user lockup + drain fee

+ = valid payer_amount
- = invalid payer_amount
```

So we actually have 2 maximums: B (for normal payments) and C (for drain), where any amount between them would fail.

For the current behavior, `pay_onchain` will throw an "insufficient funds" error if user chooses an amount between B and C (excluding B and C). As a way to avoid that, I can find B and include it in the `prepare_pay_onchain` response as a new field. Should I go for that?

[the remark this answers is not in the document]
roeierez: I don't expect users to understand there are "two" maximum. We should have one max value where all amounts underneath should work.

lib/core/src/sdk.rs
```diff
@@ -1132,48 +1165,146 @@ impl LiquidSdk {
...
 
+    /// Find the maximum `receiver_amount_sat` that can be used as argument in `prepare_pay_onchain`.
+    ///
+    /// Note that when using this amount in `prepare_pay_onchain`, the resulting `pay_onchain` will
+    /// leave 1 sat in the wallet. To send all funds and leave no sats in the wallet,
+    /// `prepare_pay_onchain` should be called with `PayOnchainAmount::Drain`.
+    ///
+    /// ### Arguments
+    /// - `sat_per_vbyte`: optional user-chosen BTC claim tx feerate
+    pub async fn get_max_receiver_amount_sat(
```
roeierez, line 1255: I don't understand why when we use lwk drain option we get higher maximum than we specify explicitly.
The drain as far as I understand uses all utxos so the tx size should be the largest possible and more than any other tx created in a different way. Am I missing something?

lib/core/src/sdk.rs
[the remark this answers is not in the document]
ok300: The reason I explained the 2 maximums above was to show there is no "maximum" as for other swaps:
- B is the closest to what you asked, but it's no maximum because there is a higher value (C)
- C is the highest value, but some amounts just below won't work.

This has implications for the UX, because if the UX allows B as maximum and the user choses it, this will result in a 1 sat change output. If there is 1 sat left in the wallet after choosing the "maximum", obviously users can get confused. However sending that amount +1 will also not work. (cc @erdemyerebasmaz ). What will work for sending all funds is explicitly calling drain.

We can however call B a "soft maximum" or find a better name. Key point is this is different than for all other swaps.

I implemented a way to get this "soft maximum" (B) in 60ee1c768021810f72bc64a8ada69d35b638185e . Is this going in the right direction?

If yes, it makes sense to include it in the `fetch_onchain_limits` response. Tricky part here is that finding B depends on an optional claim feeerate, so `fetch_onchain_limits` would get an optional feerate argument, which doesn't seem right. However IMO we cannot treat this feerate as empty, as it is a valid input in `prepare_pay_onchain` so it therefore influences the "soft maximum" that can be sent.

IMO exposing such a maximum would confuse the users more than help. Instead, I think the current behavior is reasonable:
- for UX validation, app can allow sending
  - `PayOnchainAmount::Receiver`: any amount between minimum and total balance
  - `PayOnchainAmount::Drain`: with no amount input from the user, but which internally chooses the maximum possible (C)
- when user confirms selection, app calls `prepare_pay_onchain`
- if user chose a `PayOnchainAmount::Receiver` amount between B and C, `prepare_pay_onchain` will throw an `InsufficientBalance` error
  - user is asked to try again with a different amount


Let's discuss in the call tomorrow.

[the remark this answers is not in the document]
roeierez: Do you understand why lwk wallet fails in amount less than the drain amount? See my question here: https://github.com/breez/breez-sdk-liquid/pull/464#discussion_r1743614630

lib/core/src/sdk.rs
ok300: The reason for that difference is that the drain option doesn't expect a change output, whereas for a normal send LWK includes a change output. So the tx size is higher for normal send, which affects the fee, which affects the max amount sendable.

roeierez: If that's the case then I think it is normal and we shouldn't do anything about it.
The fact that lwk wallet added a drain flag that behaves differently indicates it is a different cast.
If the user passes drain then the prepare will return X amount.
The user won't be able to send that X amount without the drain and will get insufficient funds error which seems OK to me.

ok300: As discussed, I reverted the previous commit.

Now we're not exposing a max for the custom amount, but the user can send the full balance with `drain`.

lib/core/src/sdk.rs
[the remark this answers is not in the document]
ok300: Addressed in the discussion today (not exposing the max anymore).

roeierez (approved):

lib/core/src/chain_swap.rs
```diff
@@ -582,14 +582,28 @@ impl ChainSwapStateHandler {
...
             .onchain_wallet
             .build_tx(
                 None,
                 &lockup_details.lockup_address,
                 lockup_details.amount as u64,
             )
-            .await?;
+            .await
+        {
+            Err(PaymentError::InsufficientFunds) => {
```
roeierez, line 594: Why do we need the fallback to drain? Why not just return error? As far as I understand the user didn't use explicitly "drain" flag in this flow.

lib/core/src/sdk.rs
```diff
@@ -1132,42 +1165,79 @@ impl LiquidSdk {
...
-        let user_lockup_amount_sat = user_lockup_amount_sat_without_service_fee + boltz_fees_sat;
-        self.validate_user_lockup_amount_for_chain_pair(&pair, user_lockup_amount_sat)?;
-        let lockup_fees_sat = self.estimate_lockup_tx_fee(user_lockup_amount_sat).await?;
+        let drain_payer_amount_sat = balance_sat;
+        let drain_lockup_fees_sat = self.estimate_drain_tx_fee().await?;
+        let drain_user_lockup_amount_sat = drain_payer_amount_sat - drain_lockup_fees_sat;
+        let drain_boltz_fees_sat = pair.fees.boltz(drain_user_lockup_amount_sat);
+        let drain_total_fees_sat =
+            drain_boltz_fees_sat + drain_lockup_fees_sat + claim_fees_sat + server_fees_sat;
+        let drain_receiver_amount_sat = drain_payer_amount_sat - drain_total_fees_sat;
```
roeierez, line 1188: Perhaps it is better to calculate all the drain related variables inside the is_drain branch bellow?

roeierez (changes requested):

lib/core/src/chain_swap.rs
ok300: At this point (in `pay_onchain` > changed status to `Created` > `on_new_outgoing_status` > `lockup_funds`)

https://github.com/breez/breez-sdk-liquid/blob/b20c27a3998f75bb80f80c3c42352303d774e3df/lib/core/src/chain_swap.rs#L399

we don't know what the user chose. We only know we should lockup `lockup_details.amount` sats.

So the logic iin `lockup_funds` becomes:
- we start by assuming it's a normal payment (non-drain)
- if `InssuficientFunds`, there is a chance this might have been a drain, not a normal payment
  - we try drain
    - if drain succeeds => `Ok`
    - if drain fails => propagate the error

lib/core/src/sdk.rs
ok300: Agree, that's how I started, but then the question arose
> what if the user chooses `PayOnchainAmount::Receiver` of the exact receiver amount that would have been calculated by `drain` as well?

Then, that would be equivalent to choosing `PayOnchainAmount::Drain`.

If so, it seems natural to do `drain` instead of throwing an error.

To do that, we have to know what the amounts would have been in case of `drain`, which is why they're calculated before the `is_drain` branch below.

If instead we throw an error, let me know and this can definitely be simplified.

lib/core/src/chain_swap.rs
roeierez: I see but there is still a risk here that we should handle.
Does the drain uses the given lockup amount? I am asking because there might be more funds in the wallet were when the swap was created and now we risk in paying lots of fees?
If we use drain here I think we should make sure the amount and fees we pay are as expected.

lib/core/src/sdk.rs
roeierez: > Then, that would be equivalent to choosing `PayOnchainAmount::Drain`.

No, we agreed that this would throw an error. The drain tx is only created if the user asks explicitly for drain, that was the whole point of the flag.

ok300: Done: 62751b8448ff2f2a2c147e3bb43b1c709d7dc924

lib/core/src/chain_swap.rs
ok300: > If we use drain here I think we should make sure the amount and fees we pay are as expected

We do, I added it in 1fa49259d1416138b7ee4f8fb29844dda4d2e8d4 .

The way we build the drain tx takes the expected amount as arg:

https://github.com/breez/breez-sdk-liquid/blob/62751b8448ff2f2a2c147e3bb43b1c709d7dc924/lib/core/src/chain_swap.rs#L594-L603

If LWK builds a drain tx of different amount, for any reason, this `build_drain_tx` will fail with a validation error.

ok300: So the edge case you brought up is already covered by this drain amount validation.

cli/src/commands.rs
```diff
@@ -50,9 +50,12 @@ pub(crate) enum Command {
         /// Amount that will be received, in satoshi
         receiver_amount_sat: u64,
 
-        // The optional fee rate to use, in satoshi/vbyte
+        /// The optional fee rate to use, in satoshi/vbyte
         #[clap(short = 'f', long = "fee_rate")]
         sat_per_vbyte: Option<u32>,
+
+        #[arg(short, long)]
```
dangeross, line 57: Can you add a docs comment?

lib/core/src/sdk.rs
```diff
@@ -1132,42 +1165,71 @@ impl LiquidSdk {
     }
 
     /// Prepares to pay to a Bitcoin address via a chain swap.
-    ///
-    /// # Arguments
-    ///
-    /// * `req` - the [PreparePayOnchainRequest] containing:
-    ///     * `receiver_amount_sat` - the amount in satoshi that will be received
-    ///     * `sat_per_vbyte` - the optional fee rate of the Bitcoin claim transaction. Defaults to the swapper estimated claim fee
```
dangeross, line 1140: Can you update the docs comment with the `amount: PayOnchainAmount`?

dangeross (commented):
Looks good. Just some small nitty bits

roeierez (approved):
LGTM

cli/src/commands.rs
```diff
@@ -50,9 +50,12 @@ pub(crate) enum Command {
         /// Amount that will be received, in satoshi
         receiver_amount_sat: u64,
```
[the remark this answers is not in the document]
ok300: Correct. Done in e0015b77d605cfad5e78fac2616251b1c0b5102d .

cli/src/commands.rs
ok300: Done: 5df694a0eef34fb8e38e9efedee93ba78cad35cd

lib/core/src/sdk.rs
ok300: Done: 2c3fe0a3527adb3d1d331cd4f5491966fb2b24ee

dangeross (approved):
LGTM

PR #467, opened 2024-09-03
Fix Swift macOS platform definition

dangeross:
This PR fixes the Swift macOS platform definition for pre version 6 of Swift tools where `.v15` is not defined. By changing to a string literal version the package can still be used with Swift 5+.

https://forums.swift.org/t/setting-swift-package-platform-to-ios-15-macos-12-watchos-8-not-available/49425


roeierez (approved):
LGTM

PR #470, opened 2024-09-03
sync: correctly detect payment by refund tx ID

### LINKED ISSUE #364: Refund leads to `Payment not found` log entries on `sync`
A refund causes a payment to receive a `refund_tx_id`.

This execution path catches these refund txs

https://github.com/breez/breez-liquid-sdk/blob/86d778f3b6fb9c0d65bc9ad8041cedc8d9a3cbe5/lib/core/src/sdk.rs#L1454-L1462

and triggers a payment lookup by `refund_tx_id`, which leads to
```
[DEBUG breez_liquid_sdk::sdk:485] Payment not found: <refund_tx_id>
```
on every `sync`.

One solution is to expand our `persister.get_payment` lookup to fallback to a lookup by `refund_tx_id` if nothing is found.

Another idea is to also check if an existing`self.persister.list_pending_send_swaps_by_refund_tx_id()` is found for this `refund_tx_id` in the snippet above, before emitting an event for this possibly-new tx.

ok300:
Fixes #364

### DIFF
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -1271,6 +1271,16 @@ impl Payment {
             },
         }
     }
+
+    pub(crate) fn get_refund_tx_id(&self) -> Option<String> {
+        match self.details.clone() {
+            Some(PaymentDetails::Lightning { refund_tx_id, .. }) => Some(refund_tx_id),
+            Some(PaymentDetails::Bitcoin { refund_tx_id, .. }) => Some(refund_tx_id),
+            Some(PaymentDetails::Liquid { .. }) => None,
+            None => None,
+        }
+        .flatten()
+    }
 }
 
 /// Returned when calling [crate::sdk::LiquidSdk::recommended_fees].
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -1874,9 +1874,25 @@ impl LiquidSdk {
             .list_payments(&ListPaymentsRequest::default())
             .await?
             .into_iter()
-            .filter_map(|payment| {
+            .flat_map(|payment| {
                 let tx_id = payment.tx_id.clone();
-                tx_id.map(|tx_id| (tx_id, payment))
+                let refund_tx_id = match payment.details.clone() {
+                    None => None,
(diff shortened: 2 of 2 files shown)


ok300:
@dangeross the swap is not found in that branch because it's checking for Pending Swaps and the case this PR covers is when the refund tx is confirmed (i.e. the swap is not Pending anymore).

> even if it does go into the non-swap payments branch, don't we create payment before calling emit_payment_updated(), so it should find the payment?

If it goes in the [non-swap payments branch](https://github.com/breez/breez-sdk-liquid/blob/86d778f3b6fb9c0d65bc9ad8041cedc8d9a3cbe5/lib/core/src/sdk.rs#L1454-L1462), the reason it doesn't find the payment is that `payments_before_sync` indexes payments by `tx_id`, which is the lockup or claim tx, but we're now processing a refund tx ID. This is also the crux of the fix in the PR, namely we expand `payments_before_sync` to also index the payments by refund tx ID.

dangeross (commented):
Thanks for clarifying, so it just stops `emit_payment_updated()` being called repeatedly. I still wonder why the payment is not found in `emit_payment_updated()` though, if we are adding a payment for the refund tx id

ok300:
> so it just stops `emit_payment_updated()` being called repeatedly.

Correct.

> I still wonder why the payment is not found in `emit_payment_updated()` though, if we are adding a payment for the refund tx id

Well, technically we're not adding a `Payment` for the `refund_tx_id`, we're just indexing the refunded `Payment` by `refund_tx_id` in the local cache `payments_before_sync`. So in `sync_payments_with_chain_data`, the same `Payment` is findable by both `tx_id` (lockup or claim) and by `refund_tx_id` in the local cache.

The reason `persister.get_payment` doesn't find it inside `emit_payment_updated` is that this lookup is done on the DB, in the payment query result (`select_payment_query`). That query only returns the `Payment` once and `persister.get_payment` looks it up by `tx_id`, so it won't find it.

Fortunately, we don't need it to find it, as we're using the local index to retrieve it by `refund_tx_id`.

dangeross (approved):

roeierez (approved):

PR #471, opened 2024-09-04
Receive Chain Swap: support refund even when lockup address is re-used

### LINKED ISSUE #432: Support address re-use in Receive Chain Swap
Scenario:
- Receive Chain Swap is created and successfully completed
- 24h later, new (BTC) funds are sent to the same (BTC) Lockup Address
- 10h later:
  - on Boltz, the swap state is `transaction.claimed` (the final status from the initial completed swap)
  - on SDK, the swap status is `Refundable`

Example instance: testnet swap `7u5RFB8HQNdp`

* (BTC) lockup address: `tb1pnvkd06a3q7e45pdtkyz4x3kdhnkarxespnz8uwsf7386kt2x2kpsqxaku7` ( [open in mempool explorer](https://mempool.space/testnet/address/tb1pnvkd06a3q7e45pdtkyz4x3kdhnkarxespnz8uwsf7386kt2x2kpsqxaku7) )

On SDK startup:
```
[2024-08-02 10:21:36.862 INFO breez_sdk_liquid::chain_swap:142] Chain Swap 7u5RFB8HQNdp has 5000 confirmed and 0 unconfirmed sats
[2024-08-02 10:21:36.862 INFO breez_sdk_liquid::chain_swap:153] Chain Swap 7u5RFB8HQNdp has 5000 unspent sats. Setting the swap to refundable
[2024-08-02 10:21:36.862 INFO breez_sdk_liquid::chain_swap:547] Transitioning Chain swap 7u5RFB8HQNdp to Refundable (server_lockup_tx_id = None, user_lockup_tx_id = None, claim_tx_id = None), refund_tx_id = None)
```

ok300:
Fixes #432


roeierez (approved):
LGTM

dangeross (approved):

hydra-yse (approved):
LGTM

lib/core/src/sdk.rs
```diff
@@ -1822,6 +1822,28 @@ impl LiquidSdk {
...
+            .update_swap_info(
+                &swap.id,
+                RefundPending,
+                None,
+                None,
+                None,
+                Some(&refund_tx_id),
+            )
+            .await?;
+
```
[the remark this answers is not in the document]
ok300: Done: 29872000a802b475408ee19bd4aa0a0b7c3af9d2

lib/core/Cargo.toml
```diff
@@ -14,7 +14,7 @@ frb = ["dep:flutter_rust_bridge"]
 [dependencies]
 anyhow = { workspace = true }
 bip39 = "2.0.0"
-boltz-client = { git = "https://github.com/hydra-yse/boltz-rust", branch = "yse-fee-calculation" }
+boltz-client = { git = "https://github.com/ok300/boltz-rust", branch = "ok300-chain-swap-refund-check-all-utoxs" }
```
dangeross, line 17: Do this include @hydra-yse's fee calculation fix?

ok300: Good point. No, because we started out branches more or less at the same time.

@hydra-yse can your branch be merged upstream? It's a fairly small change and should be easy to test and review.

lib/core/src/chain_swap.rs
```diff
@@ -818,18 +828,25 @@ impl ChainSwapHandler {
...
         let refund_tx_id = bitcoin_chain_service.broadcast(&refund_tx)?.to_string();
 
-        info!(
-            "Successfully broadcast refund for incoming Chain Swap {}, is_cooperative: {is_cooperative}",
-            swap.id
-        );
+        info!("Successfully broadcast refund for incoming Chain Swap {id}, is_cooperative: {is_cooperative}");
+
+        // After refund tx is broadcasted, set the payment state to `RefundPending`. This ensures:
+        // - the swap is not shown in `list-refundables` anymore
```
dangeross, line 839: It's my understanding that this PR fixes that the refundable swap is no longer shown in `list_refundables`. I have two questions, after executing a refund:
1. Is the payment shown with the refund_tx_id in `list_payments`?
2. Is it possible to broadcast another tx to effectively bump the fee?

My concern is the swap is no longer refundable if broadcasting the refund with too low a fee

ok300: >  Is the payment shown with the refund_tx_id in `list_payments`?

Yes.
 
>  Is it possible to broadcast another tx to effectively bump the fee?

Only for cooperative refunds. The non-coop refunds don't have RBF enabled.

However there is no timeout after which cooperative refunds are impossible. The only way I could test non-cooperative refunds was by commenting out the coop refund call. Boltz participated in the coop refund even weeks after a swap failed.

> My concern is the swap is no longer refundable if broadcasting the refund with too low a fee

From what I can tell, this can only happen if
1. Boltz is not reachable for the coop refund and
2. the user chooses a feerate that is too low

IMO this is unlikely because of the 1st point.

dangeross: > Only for cooperative refunds. The non-coop refunds don't have RBF enabled

I tried on main branch to do this, but I get a "No UTXO found" error. Probably because the electrum [script_list_unspent](https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain-scripthash-listunspent) excludes them **_"Any output that is spent in the mempool does not appear"_**

> Boltz participated in the coop refund even weeks after a swap failed

I think Boltz will always try to cooperate with a refund (always if the lockup failed _or_ if the swap expired after they refund themselves). So it's more doomsday handling

lib/core/Cargo.toml
ok300: The calculation fix was merged in upstream and in my branch: 87e9327d24e1da02febf3f0e7bba08570111f29e

dangeross (approved):
LGTM. Can you make an issue to handle increasing the fee based on an already broadcast refund tx? As discussed in the dev call

lib/core/src/chain_swap.rs
ok300: Tracking in #539

ok300:
> Can you make an issue to handle increasing the fee based on an already broadcast refund tx?

Done, opened a separate issue: #539

PR #472, opened 2024-09-05
Set optional prepare params as null

dangeross:
This PR sets the optional params in the UDL as default `null` so that they are not required to be past into the binding classes (e.g. python, kotlin, swift)


hydra-yse (approved):
LGTM

PR #474, opened 2024-09-05
Set `PaymentDetails` as mandatory, default to LiquidAddress

### LINKED ISSUE #469: Use LiquidAddress in payment details for onchain payments

hydra-yse:
Closes #469.

### DIFF
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -475,9 +475,9 @@ dictionary Payment {
     u64 fees_sat;
     PaymentType payment_type;
     PaymentState status;
+    PaymentDetails details;
     string? destination = null;
     string? tx_id = null;
-    PaymentDetails details;
 };
 
 enum PaymentType {
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -477,7 +477,7 @@ dictionary Payment {
     PaymentState status;
     string? destination = null;
     string? tx_id = null;
-    PaymentDetails? details = null;
+    PaymentDetails details;
 };
 
 enum PaymentType {
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -1176,9 +1176,8 @@ impl Payment {
     pub(crate) fn from_tx_data(
         tx: PaymentTxData,
         swap: Option<PaymentSwapData>,
-        payment_details: Option<PaymentDetails>,
+        details: PaymentDetails,
     ) -> Payment {
-        let description = swap.as_ref().map(|s| s.description.clone());
         Payment {
             tx_id: Some(tx.tx_id),
             // When the swap is present and of type send and receive, we retrieve the destination from the invoice.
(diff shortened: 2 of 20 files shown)


lib/core/src/model.rs
```diff
@@ -1247,27 +1247,27 @@ impl Payment {
...
-                }) => Some(PaymentDetails::Bitcoin {
+                }) => PaymentDetails::Bitcoin {
                     swap_id,
                     refund_tx_id,
                     refund_tx_amount_sat,
                     description: description.unwrap_or("Bitcoin transfer".to_string()),
-                }),
-                _ => payment_details,
+                },
+                _ => payment_details.expect("Expected existing payment details"),
```
ok300, line 1270: Don't know if `.expect()` is a good idea, as it will panic in case of `None`.

My understanding from the requirements was that it should default to `PaymentDetails::Liquid`.

Since this happens in `from_tx_data`, it means we're trying to reconstruct the details of a pure Liquid tx. Then maybe generating a `PaymentDetails::Liquid` is the way to go?

lib/core/src/sdk.rs
```diff
@@ -1917,7 +1896,13 @@ impl LiquidSdk {
                     },
                     is_confirmed: is_tx_confirmed,
                 },
-                None,
+                match tx.outputs.first() {
+                    Some(Some(output)) => Some(PaymentDetails::Liquid {
+                        destination: output.script_pubkey.to_hex(),
+                        description: "Liquid transfer".to_string(),
```
roeierez, line 1902: What do you think about leaving the description empty? I see that we fill it with default description when constructing the payment from the db which is better and more flexible than storing.

lib/core/src/model.rs
roeierez: I agree. It seems that the payment_details shouldn't be optional here now after the change.

roeierez (changes requested):

packages/react-native/src/index.ts
```diff
@@ -255,6 +255,7 @@ export interface PreparePayOnchainResponse {
 }
 
 export interface PrepareReceiveRequest {
+    payerAmountSat?: number
```
ok300, line 258: Duplicated field? Already exists 2 lines below.

hydra-yse: Yes, it was due to the rebase. Will fix soon

roeierez (approved):

ok300 (approved):

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -475,9 +475,9 @@ dictionary Payment {
     u64 fees_sat;
     PaymentType payment_type;
     PaymentState status;
+    PaymentDetails details;
```
roeierez, line 478: @hydra-yse should we also make the destination field now mandatory as well?

hydra-yse: @roeierez Looking at the code, I think that requires a PR of its own. It can be done, but we need to set the `bolt11` field in the details as mandatory, as well as the `claim_address` field for chain swaps (since they're now optional and we use them to rebuild the destination)

PR #476, opened 2024-09-05
Get payment by destination

### LINKED ISSUE #473: Add a method to lookup a payment
We are currently missing a way to lookup an existing payment. This could be done by providing the destination (e.g. bolt11 invoice, Bitcoin address or Liquid BIP21/address) when looking up the payment.

Note: this will only return a payment for a bolt11 invoice if the swap has been initiated (swap is in progress or completed). If no payment is returned you should assume the swap has not yet started (bolt11 invoice not paid).

dangeross:
This PR adds a method to lookup a payment by it's destination (send/receive swap invoice, chain swap claim_address, liquid destination)

Fixes #473


lib/core/src/persist/mod.rs
```diff
@@ -375,6 +375,21 @@ impl Persister {
             .optional()?)
     }
 
+    pub fn get_payment_by_destination(&self, destination: &str) -> Result<Option<Payment>> {
+        Ok(self
+            .get_connection()?
+            .query_row(
+                &self.select_payment_query(
+                    Some("(rs.invoice = ?1 OR ss.invoice = ?1 OR cs.claim_address = ?1 OR pd.destination = ?1)"),
```
ok300, line 383: A few questions:
- If the getter says `by_destination`, shouldn't it ideally just search in `pd.destination`? AFAIK `destination` is documented as "invoice or Liquid BIP21/address", so it should cover the invoice as well.
- If it's a generic getter `by_destination` and someone used a BIP21 as destination, but now queries by the onchain address from that BIP21, this won't find it, correct? If so, I'm wondering whether it should find it, since technically the onchain address *is* the destination.

What do you think? cc @hydra-yse

dangeross: I tried with the `pd.destination` but a payment_details table entry is only added for payments to a Liquid address. The columns used to otherwise find the `destination` are the columns otherwise used to fill the PaymentDetails for the Payment.

hydra-yse (approved):

lib/core/src/persist/mod.rs
```diff
@@ -375,6 +375,21 @@ impl Persister {
...
```
roeierez, line 383: It seems that a payment may have multiple destinations? This means that get_payment_by_destination may return same result for different destinations? for example for a claim_address and bip21 that was parsed as input?

hydra-yse: @roeierez Currently there's no case where this can happen. Specifically, here's the cases:
- Receive Swap/Chain Swap: No payment details set, no conflict can occur
- Send swap using Boltz: No payment details set, no conflict can occur
- Direct Send/Send "swap" with invoice MRH: We persist data as a Liquid payment, therefore there is no data in the swap tables
- Generic non-SDK payment (as of #474): We persist the tx's output's script pubkey in the payment details. We have no data in the swap tables.

So in all cases the swap/details are mutually exclusive, meaning there's no way for two inputs to return the same payment.

roeierez: @dangeross do you think it makes sense instead of a string to use an enum so the query will be more intuitive and search will look up for the specific identifier?
```
PaymentDestination {
  Lightning(bolt11),
  Liquid(address),
  Bitcoin(address)
}
```

dangeross: @roeierez Yeah, it does make it more intuitive and fits in with other functions of the SDK.

roeierez (approved):
LGTM

lib/core/src/sdk.rs
```diff
@@ -1976,6 +1976,24 @@ impl LiquidSdk {
...
+    ///
+    /// Returns an `Option<Payment>` if found, or `None` if no payment matches the given destination.
+    pub async fn payment_by_destination(
+        &self,
+        destination: &PaymentDestination,
+    ) -> Result<Option<Payment>, PaymentError> {
+        self.ensure_is_started().await?;
+
+        Ok(self.persister.get_payment_by_destination(destination)?)
+    }
```
hydra-yse, line 1995: Looking back on the changes, wouldn't it be better to name this `get_payment_by_destination` like the persister method? I'm afraid this could be misinterpreted as executing a payment by destination.

hydra-yse (commented):
Just a comment regarding naming, everything else looks good :smile:

lib/core/src/sdk.rs
dangeross: Agreed, I've since added `payment_hash` as a query option so `get_payment()` makes more sense now

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -487,6 +487,13 @@ dictionary Payment {
     string? tx_id = null;
 };
 
+[Enum]
+interface PaymentQuery {
+    Lightning(string? invoice, string? payment_hash);
+    Liquid(string destination);
+    Bitcoin(string address);
```
roeierez, line 494: Once thing that concerns me is that for these destinations multiple payments may exists, right now we only return the first one we find. Do you think it is reasonable? Do we need to change the interface to return multiple payments?

dangeross: Yeah, it can be the case if addresses are reused. Then I would propose adding these filter options to `list_payments()` instead and keep get payment by invoice/hash only. What do you think @roeierez ?

roeierez: Yes. I had that on mind too. I think btw we don't need by invoice if we have by hash. What do you think?

cli/src/commands.rs
```diff
@@ -436,18 +449,33 @@ pub(crate) async fn handle_command(
...
                 })
                 .await?;
             command_result!(payments)
         }
+        Command::GetPayment { payment_hash } => {
+            let payment = sdk
+                .get_payment(&GetPaymentRequest::Lightning { payment_hash })
+                .await?;
+            command_result!(payment)
+        }
```
hydra-yse, line 478:
```suggestion
        Command::GetPayment { payment_hash } => {
            let maybe_payment = sdk
                .get_payment(&GetPaymentRequest::Lightning { payment_hash })
                .await?;
            match maybe_payment {
                Some(payment) => command_result!(payment),
                None => {
                    return Err(anyhow::anyhow!("Payment not found."));
                }
            }
        }
```

hydra-yse (approved):
tACK

ok300 (approved):

PR #486, opened 2024-09-13
Bump boltz client library

ok300:
This PR bumps the `boltz-rust` library and adjusts the calls to some of the changed method signatures.


hydra-yse (approved):
Looks good!

PR #488, opened 2024-09-13
Build SDK bindings for Flutter

dangeross:
This PR adds the ability to locally build the SDK bindings into the Flutter package of use when building Misty Breez (locally or with CI) against non-published SDK bindings. By default it is setup for local builds which are overridden during CI publishing.

- Adds scripts to build and copy SDK bindings (android / swift) using `just`
- Remove the CMakeLists recursive delete to prevent binding libraries from being deleted
- Add gradle.build.production to override gradle.build (local build) during publish
- Add a breez_sdk_liquid podspec which can be referenced during the Misty Breez non-published build

Misty Breez PR: https://github.com/breez/misty-breez/pull/190


lib/bindings/langs/flutter/justfile
```diff
@@ -54,6 +55,16 @@ build-android profile='frb-min':
 build-other profile='frb-min':
 	dart scripts/build_other.dart --profile {{profile}}
 
+build-bindings:
```
erdemyerebasmaz, line 58: Can we rename 'bindings' to 'uniffi' in all occurrences?
```suggestion
build-bindings:
```

packages/flutter/android/build.gradle
```diff
@@ -52,8 +56,22 @@ android {
...
     defaultConfig {
-        minSdkVersion 16
+        minSdkVersion 24
     }
 }
+
+dependencies {
+    implementation "net.java.dev.jna:jna:5.14.0@aar"
+    /* JSON serialization */
+    implementation 'org.jetbrains.kotlinx:kotlinx-serialization-json:1.6.3'
```
erdemyerebasmaz, line 76: Kotlin 1.8.20+ has serialization plugin. I don't think we need to depend on it explicitly.

packages/flutter/android/build.gradle
```diff
@@ -4,6 +4,7 @@ group 'com.example.flutter_breez_liquid'
 version '0.2.1'
 
 buildscript {
+    ext.kotlin_version = '1.8.20'
```
erdemyerebasmaz, line 7: If there are no obstacles, I'd suggest we upgrade to Kotlin 2.0.0.

packages/flutter/ios/breez_sdk_liquid.podspec
```diff
@@ -0,0 +1,24 @@
+# To learn more about a Podspec see http://guides.cocoapods.org/syntax/podspec.html.
+# Run `pod lib lint breez_sdk.podspec` to validate before publishing.
+Pod::Spec.new do |s|
+  s.name                   = 'BreezSDKLiquid'
+  s.version                = '0.2.1'
```
erdemyerebasmaz, line 5: version script should update the version on this podspec.

packages/flutter/android/build.gradle.production
```diff
@@ -0,0 +1,59 @@
+// The Android Gradle Plugin builds the native code with the Android NDK.
+
+group 'com.example.flutter_breez_liquid'
+version '0.2.1'
```
erdemyerebasmaz, line 4: version script should update this version

packages/flutter/android/build.gradle
dangeross: Doesn't compile without this

packages/flutter/android/build.gradle
dangeross: I would prefer to keep this and the android bindings at the same kotlin version. If need be we can address it in a separate PR

lib/bindings/langs/flutter/justfile
dangeross: Done [00bfede](https://github.com/breez/breez-sdk-liquid/pull/488/commits/00bfede5f84fbd40e966b444db74a14a88fa0695)

packages/flutter/ios/breez_sdk_liquid.podspec
dangeross: Done [00bfede](https://github.com/breez/breez-sdk-liquid/pull/488/commits/00bfede5f84fbd40e966b444db74a14a88fa0695)

packages/flutter/android/build.gradle.production
dangeross: Done [00bfede](https://github.com/breez/breez-sdk-liquid/pull/488/commits/00bfede5f84fbd40e966b444db74a14a88fa0695)

erdemyerebasmaz (approved):
LGTM!!

PR #495, opened 2024-09-22
Support user implemented signer.

roeierez:
This PR allows users o initialize the sdk via their own signer implementation instead of passing the mnemonic phrase to the sdk.
A new function was added connect_with_signer and the Signer trait should be implemented and passed to the function.
For now we excluded flutter and react native because of binding issues.


ok300 (approved):
utACK.

This doesn't touch `BtcSwapTx::partial_sign`, which we use in the context of `refund_tx_wrapper.partial_sign()`, which take as arg the swap's refund keypair and internally calls `MusigSession::partial_sign`.

However this doesn't depend on user keys, so indeed the user-supplied signer shouldn't play a role for `partial_sign`.

lib/core/src/signer.rs
```diff
@@ -0,0 +1,455 @@
...
+
+    fn sign_ecdsa_recoverable(&self, msg: Vec<u8>) -> Result<Vec<u8>, SignerError> {
+        let seed = self.seed();
+        let secp = Secp256k1::new();
+        let keypair = Xpriv::new_master(self.network, &seed)
+            .map_err(|e| anyhow::anyhow!("Could not get signer keypair: {e}"))?
+            .to_keypair(&secp);
+        let s = msg.as_slice();
+
+        let double_hashed_msg: Message = Message::from_digest_slice(s)
```
dangeross, line 236: nit: variable name is misleading as we don't know it's double hashed

lib/core/src/wallet.rs
```diff
@@ -224,38 +228,16 @@ impl OnchainWallet for LiquidOnchainWallet {
...
-        let keypair = ExtendedPrivKey::new_master(self.config.network.into(), &seed)
-            .map_err(|e| anyhow!("Could not get signer keypair: {e}"))?
-            .to_keypair(&secp);
         // Prefix and double hash message
         let mut engine = sha256::HashEngine::default();
         engine.write_all(LN_MESSAGE_PREFIX)?;
         engine.write_all(message.as_bytes())?;
         let hashed_msg = sha256::Hash::from_engine(engine);
-        let double_hashed_msg = Message::from(sha256::Hash::hash(&hashed_msg));
+        let double_hashed_msg = Message::from_digest(hashed_msg.into_inner());
```
dangeross, line 237: This doesn't seem to be double hashing any more?

lib/core/src/model.rs
```diff
@@ -201,10 +199,63 @@ pub enum SdkEvent {
...
+
+    // Sign an ECDSA message using the private key derived from the master key
+    fn sign_ecdsa_recoverable(&self, msg: Vec<u8>) -> Result<Vec<u8>, SignerError>;
+
+    // Return the master blinding key for SLIP77: https://github.com/satoshilabs/slips/blob/master/slip-0077.md
+    fn slip77_master_blinding_key(&self) -> Result<Vec<u8>, SignerError>;
+
+    // HMAC-SHA256 using the private key derived from the given derivation path
+    // This is used to calculate the linking key of lnurl-auth specification: https://github.com/lnurl/luds/blob/luds/05.md
+    fn hmac_sha256(&self, msg: Vec<u8>, derivation_path: String) -> Result<Vec<u8>, SignerError>;
```
dangeross, line 249: Can you `///` them for rustdocs?

lib/core/src/wallet.rs
ok300: Indeed calling `sign-message` / `check-message` with the CLI finds the signature as invalid.

Changing the line to
```rust
let double_hashed_msg = Message::from_digest(sha256::Hash::hash(&hashed_msg).into_inner());
```
fixes the problem.

lib/core/src/model.rs
roeierez: fixed

lib/core/src/wallet.rs
roeierez: Good catch guys. I fixed that and added a test for this case.

lib/core/src/signer.rs
roeierez: Fixed

packages/flutter/example/.flutter-plugins
dangeross, line 1: All commits to `packages/flutter/example` can be removed, the example app removed in PR #510

dangeross (commented):
Looks good, those flutter example files just need removing

dangeross (approved):
LGTM

ok300 (approved):

hydra-yse (approved):
LGTM

dangeross (approved):
Tested ACK

PR #508, opened 2024-09-28
Real-time Sync: Add base traits/interfaces

### LINKED ISSUE #501: Real-time Sync: Add database structures and persistency layer

hydra-yse:
Closes #501


lib/core/src/sync/model/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub(crate) refund_private_key: String,
+}
+
+#[derive(Serialize, Deserialize)]
+pub(crate) struct ChainSyncData {
+    pub(crate) swap_id: String,
+    pub(crate) preimage: String,
+    pub(crate) description: Option<String>,
+    pub(crate) direction: Direction,
+    pub(crate) claim_details: ClaimDetails,
```
roeierez, line 31: I prefer to not use nested structures here if possible. It makes it more clear to see what data we sync.

lib/core/src/sync/model/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub(crate) accept_zero_conf: bool,
+}
+
+#[derive(Serialize, Deserialize)]
+pub(crate) struct SendSyncData {
+    pub(crate) swap_id: String,
+    pub(crate) preimage: Option<String>,
+    pub(crate) description: Option<String>,
+    pub(crate) refund_private_key: String,
+    pub(crate) swap_tree: SwapTree,
```
roeierez, line 44: Also it is better not to rely on external structures on sync. They can change and we don't control that.

lib/core/src/sync/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+    async fn connect(&self) -> Result<()>;
+
+    /// Applies the changes received from the stream to the local database
+    async fn apply_changes(&self, changes: &[Record]) -> Result<()>;
+
+    /// Retrieves the changes since a specified [record_index](Record::record_index)
+    async fn get_changes_since(&self, record_index: u64) -> Result<Vec<Record>>;
+
+    /// Adds a record to the remote
+    async fn set_record(&self, data: SyncData) -> Result<()>;
```
roeierez, line 20: Since we are using a list of Records as response to get_changes_since I think we should stick to that structure also on set_record. SyncData seems a higher abstraction.

roeierez (changes requested):
Looks good. Dropped some comments.

lib/core/src/sync/mod.rs
hydra-yse: That's actually been changed in #509, see [here](https://github.com/breez/breez-sdk-liquid/pull/509/files#diff-39d1d030470dc22070380fe8f700011c8f39961bbbf0245cf825c4bffedc1b15R32). I opted for having `get_changes_since` also decode the data.

lib/core/src/sync/model/mod.rs
hydra-yse: Agreed, will copy those on our side then

lib/core/src/sync/mod.rs
roeierez: @hydra-yse it is hard to review this PR by looking at changes in another PR.
I think either merge both into one PR or move the fixes that are related here to this PR.

hydra-yse: Sure, I'll merge the new interface definition here.

roeierez: The fact that we have a trait with set_record & get_changes_since means we have another implementation that will use these which will be triggered by the app. I think we need to see that implementation before we can determine if we need the interface to look like this.
I am OK with merging both PRs together, move to the implementation of the sync itself (at least partial) so we can review that effectively.

PR #509, opened 2024-09-30
Real-time Sync: Add base methods and traits

### LINKED ISSUE #501: Real-time Sync: Add database structures and persistency layer

### LINKED ISSUE #502: Real-time Sync: Add a method to pull changes

### LINKED ISSUE #503: Real-time Sync: Add a method to merge and apply changes

### LINKED ISSUE #504: Real-time Sync: Add a method to push changes

### LINKED ISSUE #505: Real-time Sync: Add a method to clean incoming and outgoing changes

hydra-yse:
Closes #501, closes #502, closes #503, closes #504, closes #505;

~**TODO:** Add signer methods from #495~

**TODO:** Add hooks to sync service across the SDK


lib/core/src/sync/mod.rs
```diff
@@ -1,49 +1,211 @@
...
+        for record in updatable_records {
+            if let Err(err) = self.persister.apply_record(record) {
+                warn!("Could not apply record changes: {err:?}");
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn get_changes_since(&self, from_id: i64) -> Result<Vec<Record>> {
```
roeierez, line 152: Who should call get_changes_sync? If it is internal in the sync service perhaps we don't need it in the trait?

lib/core/src/sync/mod.rs
```diff
@@ -1,49 +1,211 @@
...
+        Ok(records)
     }
 
     async fn set_record(&self, data: SyncData) -> Result<()> {
-        unimplemented!()
+        let Some(ref mut client) = *self.client.lock().await else {
+            return Err(anyhow!("Cannot run `set_record`: client not connected"));
+        };
+
+        let id = self.persister.get_latest_record_id()? + 1;
```
roeierez, line 171: We should have something unique here across devices as the id propagated to the server.

lib/core/src/sync/model/mod.rs
```diff
@@ -1,51 +1,184 @@
...
+            claim_address: value.claim_address,
+            claim_fees_sat: value.claim_fees_sat,
+            claim_private_key: value.claim_private_key,
+            refund_private_key: value.refund_private_key,
+            timeout_block_height: value.timeout_block_height,
+            payer_amount_sat: value.payer_amount_sat,
+            receiver_amount_sat: value.receiver_amount_sat,
+            accept_zero_conf: value.accept_zero_conf,
+            created_at: value.created_at,
+            description: value.description,
```
roeierez, line 51: What do you think about separating the synced data into a different table for each swap? 
We will need to add join for each query but perhaps it makes it more clear and separated what persistent data we sync/udpate and what data is in responsibility of the client to update locally?

lib/core/src/sdk.rs
```diff
@@ -1024,8 +1058,12 @@ impl LiquidSdk {
                     created_at: utils::now(),
                     state: PaymentState::Created,
                     refund_private_key: keypair.display_secret().to_string(),
+                    is_local: true,
                 };
                 self.persister.insert_send_swap(&swap)?;
+                self.sync_service
```
roeierez, line 1064: This is not atomic so we can end up with some data that is not synced upstream.
Perhaps we can have an updated_timestamp and synced_timestamp and always query for these records that have their updated_timestamp larger than the synced_timestamp as candidates to sync?

Also where is the server record id is saved in the client to associate a swap with a record id?

roeierez (changes requested):
Nice start!
My main comments are:
1. Removing methods from the sync trait that are not used externally (not sure we need an interface).
2. More atomic/consistent racking mechanism for unsynced record
3. Association mechanism of server record id for a swap.
4. Possibly separate the synced data into different tables.

lib/core/src/sdk.rs
hydra-yse: There's a new method that needs to be added in the interface. As you said in the comment above, `get_changes_since` is an internal method. I will add a `get_latest_changes`, which uses the client's `latestRecordId`.

For the first question, that is what `get_latest_changes` will be for. I haven't added it to the SDK init yet, though the idea is that on startup the sync service will send its `latestRecordId` (stored in a new `settings` table on the DB, it can be renamed afterwards), and try and pull all the upstream changes. 

This mechanism + listening for new updates at startup ensures that, when we later call `set_record`, we are always synced with upstream and can actually push changes. If we are not, set_record returns error (and we probably want to trigger `get_latest_changes` again).

lib/core/src/sync/mod.rs
hydra-yse: See https://github.com/breez/breez-sdk-liquid/pull/509#discussion_r1804157040

lib/core/src/sync/mod.rs
hydra-yse: If the client is synced, we are sure that the new tip is always unique. There is an issue with concurrency though when, for example, two clients try to send a (different) payment and each calls `set_record` before the change gets caught by the listening stream. I think a trivial fix to this is to simply retry X number of times until it succeeds (perhaps doing this in a separate thread so that it doesn't block the payment response).

lib/core/src/persist/sync.rs
```diff
@@ -0,0 +1,100 @@
...
+        Ok(())
+    }
+
+    pub(crate) fn apply_record(&self, record: DecryptedRecord) -> Result<()> {
+        match record.data {
+            SyncData::Chain(chain_data) => self.insert_chain_swap(&chain_data.into()),
+            SyncData::Send(send_data) => self.insert_send_swap(&send_data.into()),
+            SyncData::Receive(receive_data) => self.insert_receive_swap(&receive_data.into()),
+        }
+    }
```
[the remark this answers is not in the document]
hydra-yse: Yes, initially I thought the logic would be longer, though since it's just a match I agree it can be moved into the service :+1:

lib/core/src/sync/mod.rs
```diff
@@ -1,49 +1,211 @@
...
+        let request = ListenChangesRequest::new(utils::now(), self.signer.clone())
+            .map_err(|err| anyhow!("Could not sign ListenChangesRequest: {err:?}"))?;
+        let mut stream = client.listen_changes(request).await?.into_inner();
+
+        let cloned = self.clone();
+        tokio::spawn(async move {
+            while let Some(message) = stream.next().await {
+                match message {
+                    Ok(record) => {
+                        if let Err(err) = cloned.apply_changes(&[record]) {
```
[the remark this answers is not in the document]
hydra-yse: Right, I'll add a condition at the top of the stream listener that still updates the state on update (only when the swap is non-local)

lib/core/src/send_swap.rs
```diff
@@ -121,6 +129,10 @@ impl SendSwapHandler {
 
             // Boltz announced they successfully broadcast the (cooperative or non-cooperative) claim tx
             Ok(SubSwapStates::TransactionClaimed) => {
+                if !swap.is_local {
```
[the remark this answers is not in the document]
hydra-yse: See https://github.com/breez/breez-sdk-liquid/pull/509#discussion_r1804163915

lib/core/src/send_swap.rs
```diff
@@ -111,6 +115,10 @@ impl SendSwapHandler {
             // Boltz has detected the lockup in the mempool, we can speed up
             // the claim by doing so cooperatively
             Ok(SubSwapStates::TransactionClaimPending) => {
+                if !swap.is_local {
```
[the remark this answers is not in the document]
hydra-yse: Right, I'll add a condition at the top of the stream listener that still updates the state on update (only when the swap is non-local)

lib/core/src/sync/model/mod.rs
hydra-yse: To your point, I think it definitely helps making the database structure have more sense (we can look at the schema and have a better idea of how the implementation is going to look like).

Is this necessary from a code standpoint, though? Taking a look at the sync models IMO is enough to do what you said: distinguish which fields will be synced and which won't. Since these models are only necessary for transport (meaning that, in the end, we still have to convert to a swap to insert to the tables), I don't see the need to reflect these abstractions to the database.

lib/core/src/sync/mod.rs
roeierez: There are two identifiers that have different purpose:
1. record_id - which is created and set by the client, it shouldn't be synced, it only should be unique and we can use GUID for example. Is is used to identify the record to update/sync.
2. version - which is set by the server and is used to detect changes.

Do you think we can use GUID and not sync this id?

hydra-yse: > version - which is set by the server and is used to detect changes.

This is set by the client, to show which schema version it is currently using. That was the change we settled on, correct?

> which is created and set by the client, it shouldn't be synced, it only should be unique and we can use GUID for example. Is is used to identify the record to update/sync.

Why do we need a separate id/GUID even? The implementation is currently pretty simple:
- User sends record with an id (local_tip + 1)
- Server checks if given id can be added to the tip (local_tip + 1 == current_tip + 1). If not, return error (client then has to re-sync and try again)
- If it is added, return a response with the new_id (= the same as the one in the request payload, so we update local_tip to local_tip + 1)

This ensures that the client _has to_ be synced, and also that records cannot be added at random.

roeierez: But we don't need to ensure the ientbis synced in order to add a record. Also what do you mean add records at random? Is there any requirement here? 
I am seeing a mechanism of retry on the client that can be avoided by using client generated IDs. Do you see any disadvantage in the suggested approach or any advantage in the current approach?

hydra-yse: The whole reason behind numbered IDs is that we can fetch from a given height. If we don't use a numbered ID, then we have to create a record timestamp on the server's db and then use that to index.

I agree though that the id in the set record payload is pretty useless and can be avoided. I've re-thought about the use-cases and there seems to be no issue if a client sets a record even when they're behind. 

So I'll change `SetRecordRequest` so it only passes data and version, and keep the `new_id` in `SetRecordResponse` which can then be used by the client to update the local tip.

lib/core/src/sync/model/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+    }
+}
+
+impl Record {
+    pub(crate) fn new(
+        data: SyncData,
+        revision: Option<i64>,
+        signer: Arc<Box<dyn Signer>>,
+    ) -> Result<Self, anyhow::Error> {
+        let id = uuid::Uuid::new_v4().to_string();
```
roeierez, line 239: It seems that we always create a new id and not saving it in the db (associate it with the swap).
the whole point of this client generated id is so we can associate later updated data to the right record in the db.

lib/core/src/sync/client.rs
```diff
@@ -0,0 +1,71 @@
...
+#[async_trait]
+impl SyncerClient for BreezSyncerClient {
+    async fn connect(&self, connect_url: String) -> Result<()> {
+        let mut client = self.inner.lock().await;
+        *client = Some(ProtoSyncerClient::connect(connect_url).await?);
+        Ok(())
+    }
+
+    async fn set_record(&self, req: SetRecordRequest) -> Result<SetRecordReply> {
+        let Some(mut client) = self.inner.lock().await.clone() else {
```
roeierez, line 44: This will lock the client until the end of the invocation. I think it is better to extract out getting the client to its own function: `get_client()` so the lock is released right after the get.

lib/core/src/sync/mod.rs
```diff
@@ -0,0 +1,214 @@
...
+        (updatable_records, failed_records)
+    }
+}
+
+impl SyncService {
+    pub(crate) async fn connect(&self) -> Result<()> {
+        self.client.connect(self.connect_url.clone()).await
+    }
+
+    pub(crate) async fn listen(self: Arc<Self>) -> Result<()> {
```
roeierez, line 83: NIT: from the client perspective it makes sense to have listen_changes but from the Syncer perhaps just start ? It does not only listen but also apply the changes and ensure we are up to sync.

lib/core/src/sync/client.rs
hydra-yse: The inner data (client) is cloned and the lock is dropped there

lib/core/src/sync/mod.rs
hydra-yse: Since we need to call `apply_changes` which is a Syncer method, to avoid passing a reference to the Syncer on the Client we should keep the logic inside the Syncer.

lib/core/src/persist/chain.rs
```diff
@@ -76,43 +78,50 @@ impl Persister {
...
+                cs.claim_fees_sat,
+                cs.claim_tx_id,
+                cs.refund_tx_id,
+                cs.created_at,
+                cs.state,
+                sd.is_local,
+                sd.revision,
+                sd.record_id
+            FROM chain_swaps cs, sync_details sd
+            WHERE cs.id = sd.data_identifier
```
dangeross, line 124: Would this be better as a JOIN?

lib/core/src/persist/receive.rs
```diff
@@ -78,8 +81,12 @@ impl Persister {
...
                 rs.claim_tx_id,
                 rs.created_at,
-                rs.state
-            FROM receive_swaps AS rs
+                rs.state,
+                sd.is_local,
+                sd.revision,
+                sd.record_id
+            FROM receive_swaps AS rs, sync_details sd
+            WHERE rs.id = sd.data_identifier
```
dangeross, line 89: Same question about a JOIN?

lib/core/src/persist/send.rs
```diff
@@ -80,27 +85,31 @@ impl Persister {
...
+                ss.refund_private_key,
+                ss.lockup_tx_id,
+                ss.refund_tx_id,
+                ss.created_at,
+                ss.state,
+                sd.is_local,
+                sd.revision,
+                sd.record_id
+            FROM send_swaps ss, sync_details sd
+            WHERE ss.id = sd.data_identifier
```
dangeross, line 112: Same question about a JOIN?

lib/core/src/persist/sync.rs
```diff
@@ -0,0 +1,163 @@
...
+
+use crate::sync::model::{sync::Record, SyncDetails};
+
+use super::Persister;
+
+impl Persister {
+    pub(crate) fn get_latest_revision(&self) -> Result<i64> {
+        let con = self.get_connection()?;
+
+        let latest_revision: i64 = con
```
dangeross, line 12: Will the revision ever be < 0?

dangeross (commented):
50% through, I think RN needs generating also

lib/core/src/sync/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+        for record in failed_records {
+            self.persister.insert_pending_record(record)?;
+        }
+
+        let changes = self.outbound_changes.lock().await;
+
+        // We apply records which we can update (<= CURRENT_SCHEMA_VERSION)
+        // Additionally, if there is a conflict with an outbound record, it will be resolved
+        // before persisting the update
+        for mut record in updatable_records {
```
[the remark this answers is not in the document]
roeierez: Also it seems that the outbound_changes are kept in memory and its content existence depends on some caller to call set_records (it is populated on set_record).
Not sure it is robust enough. I think it is better to save the state of the record in one place (sync_details) so we can save the changed fields array there as an additional field.
This way even if set_record failed we can keep getting the "dirty" record and sync it out according to its sync_state. As less duplicate information we have as more robust the process.

lib/core/src/sync/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+
+const CURRENT_SCHEMA_VERSION: f32 = 0.01;
+const MAX_SET_RECORD_REATTEMPTS: u8 = 5;
+
+pub(crate) struct SyncService {
+    connect_url: String,
+    persister: Arc<Persister>,
+    signer: Arc<Box<dyn Signer>>,
+    client: Box<dyn SyncerClient>,
+    outbound_changes: Mutex<HashMap<String, OutboundChange>>,
```
roeierez, line 31: Not sure we need the outbound_changes, we can have that info in the db (additional field that saves the changed fields array).
Also not sure we need a max_attempt mechanism here, I would try forever for each item that fails. It will propagate unexpected bugs to the surface and will allow us to fix them.

lib/core/src/persist/sync.rs
roeierez, line 12: I think this mechanism is not reliable as it counts on the fact that synced items are sent (and received) at the same order without skipped items. It doesn't have to be the case as the server may handle concurrent updates which trigger events but there is not gurantee they will be received at the same order.
If we call list_changes then we can persist the latest revision as the last one and count on it.

roeierez (changes requested):

hydra-yse:
Closed in favor of #555.

hydra-yse:
Closed in favor of #555.

PR #510, opened 2024-09-30
Flutter uniFFI

erdemyerebasmaz:
This is a PR to allow Flutter and Kotlin/Swift to share the binary with flutter_rust_bridge and uniFFI interfaces


.github/workflows/publish-flutter.yml
```diff
@@ -78,24 +85,30 @@ jobs:
...
         with:
-          name: bindings-flutter
-          path: bindings/
+          name: bindings-swift
+          path: dist/ios/Sources/BreezSDKLiquid/
+
+      - name: Move Source header files
+        working-directory: dist/ios/Sources/BreezSDKLiquid/
+        run: |
+          mv breez_sdk_liquidFFI.h ../../Classes
```
dangeross, line 94: I think this is covered above now, but we should still remove the modulemap

.github/workflows/publish-flutter.yml
```diff
@@ -57,14 +57,21 @@ jobs:
         run: |
           rm -rf android
           rm -rf ios
+          rm -rf macos
           rm -rf lib
           cp -r ../build/packages/flutter/android .
           cp -r ../build/packages/flutter/ios .
+          cp -r ../build/packages/flutter/macos .
+          cp -r ../build/lib/bindings/langs/swift/Sources ios/
+          cp -r ../build/lib/bindings/langs/swift/Sources macos/
```
dangeross, line 66: We can remove these as we are downloading the `bindings-swift` artifact, which also includes the generated binding.... but... we need to add a step after the download to copy the ios/Sources to macos/Sources

.github/workflows/publish-flutter.yml
erdemyerebasmaz: Done. 
[cleanup header file artifacts](https://github.com/breez/breez-sdk-liquid/pull/510/commits/dcd92544562d6d9198ba8f9c445523e13c5b8a7b)

dangeross (approved):
LGTM

PR #514, opened 2024-10-02
Re-add convenience Flutter gen command

dangeross:
Re-add the convenience Flutter `gen` command to generate code/FFI bindings and softlink C headers

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -196,14 +196,14 @@ jobs:
         working-directory: lib/bindings/langs/flutter/
         run: just frb
 
-      - name: Generate Dart/Flutter bindings
+      - name: Generate Dart/Flutter bindings & Softlink C Headers
         working-directory: lib/bindings/langs/flutter/
         run: just codegen
 
-      - name: Link C Headers
+      - name: Build iOS Framework & Softlink C Headers
         working-directory: lib/bindings/langs/flutter/
         if: ${{ success() || failure() }}
-        run: just link-headers
+        run: just build-ios-framework
 
       - name: Generate FFI bindings
         working-directory: lib/bindings/langs/flutter/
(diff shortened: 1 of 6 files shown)


roeierez (approved):

erdemyerebasmaz (commented):
Where is this being used?

uniFFI headers are non-existent when`build-uniffi-swift` is not run and `ffigen` relies on both frb & uniFFI C headers.

dangeross:
> Where is this being used?

@erdemyerebasmaz we need a single command people can run to generate everything needed for Flutter when developing the SDK. What do you suggest to add?

erdemyerebasmaz:
> > Where is this being used?
> 
> @erdemyerebasmaz we need a single command people can run to generate everything needed for Flutter when developing the SDK. What do you suggest to add?

That command is `just bootstrap true true`

> > Where is this being used?
> 
> @erdemyerebasmaz we need a single command people can run to generate everything needed for Flutter when developing the SDK. What do you suggest to add?

I'd decouple uniFFI C header file generation from `just build-uniffi-swift` / `make bindings-swift` scripts if possible and make it part of `gen` / `bootstrap` recipes.

erdemyerebasmaz:
> I'd decouple uniFFI C header file generation from `just build-uniffi-swift` / `make bindings-swift` scripts if possible and make it part of `gen` / `bootstrap` recipes.

Decoupled uniFFI C header file generation from `just build-uniffi-swift` / `make bindings-swift` scripts.

> > we need a single command people can run to generate everything needed for Flutter when developing the SDK.

Created a single command, `just build` recipe to build the uniFFI library & generate Dart/Flutter & FFI bindings.

`just gen` can be used without rebuilding the libraries if there are no structural changes to save development time. It also has an option to skip building iOS framework, `just gen false`.

PR #516, opened 2024-10-02
`list-refundables`: show refundable amount, not swap amount

ok300:
The refundable amount is based on the locked up utxo(s), not the initial swap amount.

To retrieve this, we have to make a chain service lookup.

### DIFF
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -3,6 +3,7 @@ use std::path::PathBuf;
 use anyhow::{anyhow, Result};
 
 use boltz_client::{
+    bitcoin::ScriptBuf,
     network::Chain,
     swaps::boltz::{
         CreateChainResponse, CreateReverseResponse, CreateSubmarineResponse, Leaf, Side, SwapTree,
@@ -606,6 +607,18 @@ impl ChainSwap {
             )?),
         };
         Ok(swap_script)
+    }
+
+    pub(crate) fn get_lockup_swap_script_pubkey(
+        &self,
+        network: LiquidNetwork,
+    ) -> SdkResult<ScriptBuf> {
+        let swap_script = self.get_lockup_swap_script()?.as_bitcoin_script()?;
+        let script_pubkey = swap_script
+            .to_address(network.as_bitcoin_chain())
+            .map_err(|e| anyhow!("Error getting script address: {e:?}"))?
+            .script_pubkey();
+        Ok(script_pubkey)
     }
 
     pub(crate) fn from_boltz_struct_to_json(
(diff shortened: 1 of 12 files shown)


hydra-yse (approved):
utACK

lib/core/src/sdk.rs
```diff
@@ -1740,12 +1740,27 @@ impl LiquidSdk {
...
+        let chain_swaps = self.persister.list_refundable_chain_swaps()?;
+
+        let mut refundables = vec![];
+        for chain_swap in chain_swaps {
+            let script_pubkey = chain_swap.get_lockup_swap_script_pubkey(self.config.network)?;
+            let script_balance = self
+                .bitcoin_chain_service
+                .lock()
+                .await
+                .script_get_balance(script_pubkey.as_script())?;
```
roeierez, line 1752: Electrum has a batched version of this API: https://docs.rs/electrum-client/latest/electrum_client/trait.ElectrumApi.html#tymethod.batch_script_get_balance
Perhaps we can use get all balances in one call.

roeierez (changes requested):

lib/core/src/sdk.rs
ok300: Done.

lib/core/src/model.rs
```diff
@@ -806,17 +828,9 @@ impl ReceiveSwap {
 pub struct RefundableSwap {
     pub swap_address: String,
     pub timestamp: u32,
+    /// Amount that is refundable, which is different than the swap amount
```
dangeross, line 831: It *might* be the same right? So maybe its clearer to write
```suggestion
    /// Amount that is refundable, from all UTXOs
```

dangeross (commented):
Looks good, just a few comments

roeierez (approved):
LGTM

lib/core/src/model.rs
```diff
@@ -608,6 +609,27 @@ impl ChainSwap {
...
 
+    /// Returns the lockup script pubkey for Receive Chain Swaps
+    pub(crate) fn get_receive_lockup_swap_script_pubkey(
+        &self,
+        network: LiquidNetwork,
+    ) -> SdkResult<ScriptBuf> {
+        let swap_script = self.get_lockup_swap_script()?.as_bitcoin_script()?;
+        let script_pubkey = swap_script
+            .to_address(network.as_bitcoin_chain())
+            .map_err(|e| anyhow!("Error getting script address: {e:?}"))?
```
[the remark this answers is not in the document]
ok300: Done: 64e282b7f89d2a982bdd659888855bdda343dbb1

lib/core/src/model.rs
ok300: Good point, done: 723679ea0a2bd6acec6be71b420c9e1313c25ee5

dangeross (approved):

PR #520, opened 2024-10-04
SDK: Require Breez API key globally

hydra-yse:


lib/core/src/sdk.rs
```diff
@@ -95,11 +101,39 @@ impl LiquidSdk {
         Ok(sdk)
     }
 
+    fn validate_api_key(api_key: &str) -> Result<()> {
+        let pem = [
+            "-----BEGIN CERTIFICATE-----",
+            api_key,
+            "-----END CERTIFICATE-----",
+        ]
+        .join("\n");
```
hydra-yse, line 110: Alternatively, we can use `include_bytes!` macro and pass the file path to the cert file instead (perhaps send an email which downloads the certificate directly).

dangeross: I would say most people are injecting the API key as an env variable and don't necessarily have easy access to the file system

lib/core/src/sdk.rs
```diff
@@ -95,11 +101,39 @@ impl LiquidSdk {
         Ok(sdk)
     }
 
+    fn validate_api_key(api_key: &str) -> Result<()> {
+        let pem = [
```
roeierez, line 105: I think the certificate is already in DER format (encoded as base64). Perhaps it is better to just decode the base64 into a slice of bytes and then parse is from that format using this? https://docs.rs/x509-parser/latest/x509_parser/fn.parse_x509_certificate.html
Instead of adding the prefix/suffix to adapt it to PEM format.

roeierez (changes requested):
Looks good. One comment regarding certificate parsing.

lib/core/src/sdk.rs
hydra-yse: Right, I tried to use that method directly, but will try b64 decoding first. Thanks!

lib/core/src/sdk.rs
```diff
@@ -95,11 +102,33 @@ impl LiquidSdk {
...
+    fn validate_api_key(api_key: &str) -> Result<()> {
+        let api_key_decoded = base64::engine::general_purpose::STANDARD
+            .decode(api_key.as_bytes())
+            .map_err(|err| anyhow!("Could not base64 decode the Breez API key: {err:?}"))?;
+        let (_rem, cert) = parse_x509_certificate(&api_key_decoded)
+            .map_err(|err| anyhow!("Invaid certificate for Breez API key: {err:?}"))?;
+
+        match cert.public_key().parsed() {
+            Ok(pk) => ensure_sdk!(
+                pk == BREEZ_CA_PUBLIC_KEY,
```
roeierez, line 114: My concern here is that from some reason we will have to replace our. CA and sign new certificates with a different CA. This will break existing clients.
Since this validation is not for security reasons but only to ensure the user didn't do a mistake and used the wrong key so perhaps we can just verify the CA name starts with "Breez". Ii talked to @yaslama and this will not change also on future certificates.

roeierez (changes requested):
Looks good. One more comment and good to go.

lib/bindings/tests/bindings/csharp/Program.cs
```diff
@@ -4,8 +4,8 @@
 try
 {
     var mnemonic = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about";
-
-    var config = BreezSdkLiquidMethods.DefaultConfig(LiquidNetwork.Testnet);
+    var testPubkey = "MIIBazCCAR2gAwIBAgIHPWTUED8FpTAFBgMrZXAwEDEOMAwGA1UEAxMFQnJlZXowHhcNMjQxMDA0MjMxMjM0WhcNMzQxMDAyMjMxMjM0WjAlMQ4wDAYDVQQKEwVCcmVlejETMBEGA1UEAxMKQnJlZXotVGVzdDAqMAUGAytlcAMhANCD9cvfIDwcoiDKKYdT9BunHLS2/OuKzV8NS0SzqV13o4GAMH4wDgYDVR0PAQH/BAQDAgWgMAwGA1UdEwEB/wQCMAAwHQYDVR0OBBYEFNo5o+5ea0sNMlW/75VgGJCv2AcJMB8GA1UdIwQYMBaAFN6q1pJW843ndJIW/Ey2ILJrKJhrMB4GA1UdEQQXMBWBE2h5ZHJhX3lzZUBwcm90b24ubWUwBQYDK2VwA0EAILwfdmryGRMoYX5HotHXci5Z5yTG3ugwdNURG3puXSG2eSXgbmwdzGDJBFTvNsEx9sP0L7nyOaKr3W1Yzf/ECQ==";
```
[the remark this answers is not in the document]
hydra-yse: It's a dummy API key I created specifically for tests. Since they're being executed by CI (at least the golang and csharp ones) I don't see another way of going around this but to provide a working key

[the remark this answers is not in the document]
roeierez: We can inject the API key via secret env var for CI.
Also if the tests are running on testnet we can not require API key for testnet. Both options are ok IMO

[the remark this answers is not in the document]
hydra-yse: The second options sounds good, as we're not using Esplora for testnet anyway

lib/core/src/sdk.rs
```diff
@@ -95,11 +97,38 @@ impl LiquidSdk {
...
+
+        Ok(())
+    }
+
     fn new(
         config: Config,
         swapper_proxy_url: Option<String>,
         mnemonic: String,
     ) -> Result<Arc<Self>> {
+        Self::validate_api_key(&config.breez_api_key)?;
```
roeierez, line 130: Now that we use empty key in testnet should we skip the validation here?

roeierez (changes requested):

roeierez (changes requested):
Also flutter binding CI fails.

roeierez (approved):
Except for the CI failure looks good to me.

lib/core/Cargo.toml
```diff
@@ -49,6 +49,8 @@ hex = "0.4"
 reqwest = { version = "=0.11.20", features = ["json"] }
 electrum-client = { version = "0.19.0" }
 zbase32 = "0.1.2"
+x509-parser = { version = "0.16.0", features = ["verify"] }
+base64 = "0.22.1"
```
ok300, line 53: Instead of adding a new dependency, I would just use one of the `base64`s re-exported by our dependencies where we need `base64`, like one of
```rust
// boltz_client::bitcoin::base64
// lwk_wollet::bitcoin::base64
```

lib/core/Cargo.toml
```diff
@@ -49,6 +49,8 @@ hex = "0.4"
 reqwest = { version = "=0.11.20", features = ["json"] }
 electrum-client = { version = "0.19.0" }
 zbase32 = "0.1.2"
+x509-parser = { version = "0.16.0", features = ["verify"] }
```
ok300, line 52: It also works without the `verify` feature:
```toml
x509-parser = "0.16.0"
```

This keeps the `Cargo.lock`s and the binary size smaller.

ok300 (commented):
Looks good, just a few NITs.

ok300 (approved):

PR #524, opened 2024-10-11
Add prepare_lnurl_pay

### LINKED ISSUE #468: Calculate the fee upfront when using LNURL-pay.

dangeross:
This PR adds a `prepare_lnurl_pay()` interface method as a step before `lnurl_pay()` to calculate the fees involved with the payment, whether it is a payment via Lightning or paid directly to a Liquid address using the Magic Route Hint retrieved in the invoice.

Depends on https://github.com/breez/breez-sdk-greenlight/pull/1104 (CI tests will be broken)

Fixes #468

### DIFF
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -100,6 +100,13 @@ dictionary LnUrlErrorData {
 };
 
 [Enum]
+interface SuccessAction {
+    Aes(AesSuccessActionData data);
+    Message(MessageSuccessActionData data);
+    Url(UrlSuccessActionData data);
+};
+
+[Enum]
 interface SuccessActionProcessed {
     Aes(AesSuccessActionDataResult result);
     Message(MessageSuccessActionData data);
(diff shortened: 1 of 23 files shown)


ok300 (approved):
Looks good to me.

Validation and fee calculation happens in `prepare_lnurl_pay`, which makes sense.

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -344,6 +349,22 @@ dictionary CheckMessageResponse {
...
 
+dictionary PrepareLnUrlPayRequest {
+    LnUrlPayRequestData data;
+    u64 amount_msat;
+    string? comment = null;
+    boolean? validate_success_action_url = null;
+};
+
+dictionary PrepareLnUrlPayResponse {
+    PrepareSendResponse prepare_send_response;
```
roeierez, line 360: IMO it is better to have the destination and fee_sat directly under the PrepareLnUrlPayResponse and not use the PrepareSendResponse for that. It reduces decoupling between the two methods.

roeierez (changes requested):
Looks good, one small comment.

lib/bindings/src/breez_sdk_liquid.udl
dangeross: Agreed. Done [27313d5](https://github.com/breez/breez-sdk-liquid/pull/524/commits/27313d5616373424875cfd58396aab99271b8957)

roeierez (approved):
LGTM

PR #531, opened 2024-10-16
Reserved address pool

### LINKED ISSUE #523: Missing event when bolt11 is paid via magic hints.
Currently if a bolt11 invoice created by the sdk is paid by another sdk instance then the magic hints are used and the payment goes directly to the bip21 url which wraps an internal liquid address of the underline wallet.
The problem is that we don't have a strong connection between the bolt11 invoice and that internal address and therefore we don't emit payment received event for the bolt11.
In order to solve that we need a strong 1 to 1 connection which requires a unique address per bolt11 invoice.
While this will work it has a tradeoff of increasing number of addresses which will make the sync heavier and heavier over time (not as a factor of the actual paid invoices).
To solve that we can leverage the fact that invoices have expiration time and re-use a specific address that was used in an invoice magic hints as soon as that invoice is expired.

Here is an optional implementation:
We can Maintain the following state:
1. A pool of reused addresses where each address is  mapped to an expiration time and payment hash.
2. The last address index (number) of the underline wallet.

Whenever we create a new invoice we do the following to get the address for the magic hints:
1. Check in the pool if we have an expired address, if so, take it and change the mapped data to the newly created invoice expiration and hash.
2. If we don't have an expired address,  increment the address index, create a new one from the internal walle, increment the index again and update the pool with this address and the newly created invoice expiration and hash
3. On payment received to an address do the following:
    - Check if this address is mapped to an invoice hash and if so emit payment received event.
    - Remove the address from the pool

When realtime sync is implemented we will need to sync the state (mapped addresses and index)

dangeross:
This PR:
- Adds addresses allocated to swaps to a reserved pool of addresses until the swap expires
- Gets the next unused address from either an expired reserved address or the next index from the cached derivation index
- Caches the last used derivation index
- Associates a magic routing hint address to the reverse swap, which then when paid via this address, is linked to the reverse swap (if the address is unused, it will be reallocated after expiry)

TODO:
- [x] Adjust the `payer_amount_sat` and `receiver_amount_sat` if paid via magic routing hint to set the payment fees to 0 (both should equal tx amount)
- [x] Patch LWK to allow an "up to index" full scan to prevent gap limit errors

QUESTIONS:
- Should a payment created using a LiquidAddress method expire?

Fixes #523

### DIFF
--- a/lib/core/src/persist/address.rs
+++ b/lib/core/src/persist/address.rs
@@ -1,0 +1,151 @@
+use anyhow::Result;
+use log::debug;
+use rusqlite::{Row, Transaction, TransactionBehavior};
+
+use crate::error::PaymentError;
+
+use super::{Persister, ReservedAddress};
+
+impl Persister {
+    pub(crate) fn next_expired_reserved_address(
+        &self,
+        tip: u32,
+    ) -> Result<Option<ReservedAddress>> {
+        let mut con = self.get_connection()?;
+        let tx = con.transaction_with_behavior(TransactionBehavior::Immediate)?;
+        // Get the next expired reserved address
+        let query = Self::get_reserved_address_query(vec!["expiry_block_height < ?1".to_string()]);
+        let res = match tx.query_row(&query, [tip], Self::sql_row_to_reserved_address) {
+            Ok(reserved_address) => {
+                // Delete the reserved address
+                Self::delete_reserved_address_inner(&tx, &reserved_address.address)?;
+                Some(reserved_address)
+            }
+            Err(_) => None,
+        };
+        tx.commit()?;
+
+        Ok(res)
+    }
+
+    fn get_reserved_address_query(where_clauses: Vec<String>) -> String {
+        let mut where_clause_str = String::new();
+        if !where_clauses.is_empty() {
+            where_clause_str = String::from("WHERE ");
+            where_clause_str.push_str(where_clauses.join(" AND ").as_str());
+        }
+
+        format!(
+            "
+            SELECT
+                address,
+                expiry_block_height
+            FROM reserved_addresses
+            {where_clause_str}
+            ORDER BY expiry_block_height ASC
+            LIMIT 1
+        "
+        )
+    }
+
+    pub(crate) fn insert_or_update_reserved_address(
+        &self,
+        address: &str,
+        expiry_block_height: u32,
+    ) -> Result<(), PaymentError> {
+        let con = self.get_connection()?;
+        con.execute(
+            "INSERT OR REPLACE INTO reserved_addresses (
+           address,
+           expiry_block_height
+        )
+        VALUES (?, ?)
+        ",
+            (&address, expiry_block_height),
+        )?;
+        debug!(
+            "Reserved address {} until block height {}",
+            address, expiry_block_height
+        );
+
+        Ok(())
+    }
+
+    pub(crate) fn delete_reserved_address(&self, address: &str) -> Result<(), PaymentError> {
+        let mut con = self.get_connection()?;
+        let tx = con.transaction()?;
+        Self::delete_reserved_address_inner(&tx, address)?;
+        tx.commit()?;
+
+        Ok(())
+    }
+
+    fn delete_reserved_address_inner(tx: &Transaction, address: &str) -> Result<(), PaymentError> {
+        tx.execute(
+            "DELETE FROM reserved_addresses WHERE address = ?",
+            [address],
+        )?;
+
+        Ok(())
+    }
+
+    fn sql_row_to_reserved_address(row: &Row) -> rusqlite::Result<ReservedAddress> {
+        Ok(ReservedAddress {
+            address: row.get(0)?,
+            expiry_block_height: row.get(1)?,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use anyhow::Result;
+
+    use crate::test_utils::persist::new_persister;
+
+    #[test]
+    fn test_next_expired_reserved_address() -> Result<()> {
+        let (_temp_dir, storage) = new_persister()?;
+        let address = "tlq1pq2amlulhea6ltq7x3eu9atsc2nnrer7yt7xve363zxedqwu2mk6ctcyv9awl8xf28cythreqklt5q0qqwsxzlm6wu4z6d574adl9zh2zmr0h85gt534n";
+
+        storage.insert_or_update_reserved_address(&address, 100)?;
+
+        let maybe_reserved_address = storage.next_expired_reserved_address(99)?;
+        // Under the expiry, not popped
+        assert!(maybe_reserved_address.is_none());
+
+        let maybe_reserved_address = storage.next_expired_reserved_address(100)?;
+        // Equal to expiry, not popped
+        assert!(maybe_reserved_address.is_none());
+
+        let maybe_reserved_address = storage.next_expired_reserved_address(101)?;
+        // Address expired, popped
+        assert!(maybe_reserved_address.is_some());
+
+        let maybe_reserved_address = storage.next_expired_reserved_address(102)?;
+        // Address already popped
+        assert!(maybe_reserved_address.is_none());
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_delete_reserved_address() -> Result<()> {
+        let (_temp_dir, storage) = new_persister()?;
+        let address = "tlq1pq2amlulhea6ltq7x3eu9atsc2nnrer7yt7xve363zxedqwu2mk6ctcyv9awl8xf28cythreqklt5q0qqwsxzlm6wu4z6d574adl9zh2zmr0h85gt534n";
+
+        storage.insert_or_update_reserved_address(&address, 100)?;
+
+        let maybe_reserved_address = storage.next_expired_reserved_address(99)?;
+        // Under the expiry, not popped
+        assert!(maybe_reserved_address.is_none());
+
+        storage.delete_reserved_address(&address)?;
+
+        let maybe_reserved_address = storage.next_expired_reserved_address(101)?;
+        // Over the expired, but already deleted
+        assert!(maybe_reserved_address.is_none());
+
+        Ok(())
+    }
+}
--- a/lib/core/src/persist/cache.rs
+++ b/lib/core/src/persist/cache.rs
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use rusqlite::{Transaction, TransactionBehavior};
 use std::str::FromStr;
 
 use super::Persister;
@@ -6,10 +7,12 @@ use super::Persister;
 const KEY_SWAPPER_PROXY_URL: &str = "swapper_proxy_url";
 const KEY_IS_FIRST_SYNC_COMPLETE: &str = "is_first_sync_complete";
 const KEY_WEBHOOK_URL: &str = "webhook_url";
+// TODO: The `last_derivation_index` needs to be synced
+const KEY_LAST_DERIVATION_INDEX: &str = "last_derivation_index";
 
 impl Persister {
-    pub fn get_cached_item(&self, key: &str) -> Result<Option<String>> {
-        let res = self.get_connection()?.query_row(
+    fn get_cached_item_inner(tx: &Transaction, key: &str) -> Result<Option<String>> {
+        let res = tx.query_row(
             "SELECT value FROM cached_items WHERE key = ?1",
             [key],
             |row| row.get(0),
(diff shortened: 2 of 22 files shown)


ok300:
> Adds addresses allocated to swaps to a reserved pool of addresses until the swap expires

> Associates a magic routing hint address to the reverse swap, which then when paid via this address, is linked to the reverse swap

What happens when a Receive Swap is paid via the MRH Liquid address _after_ the swap expired? Will this also be associated with the swap?

Also what happens multiple payments are made to the MRH Liquid address, i.e. address re-use? Will those payments be associated with the same swap?

roeierez:
> What happens when a Receive Swap is paid via the MRH Liquid address _after_ the swap expired? Will this also be associated with the swap?

Since the magic hints is an internal mechanism hidden inside the invoice clients should parse the invoice in order to pay. The expected behavior by clients is to return error (invoice expired) immediately and not continue to pay so there should not be payments after the swap has expired. If clients other than the sdk do not behave this way then yes worst case we might have multiple payments associated. I think associating the first one only is reasonable.

lib/core/src/persist/cache.rs
```diff
@@ -6,6 +6,7 @@ use super::Persister;
 const KEY_SWAPPER_PROXY_URL: &str = "swapper_proxy_url";
 const KEY_IS_FIRST_SYNC_COMPLETE: &str = "is_first_sync_complete";
 const KEY_WEBHOOK_URL: &str = "webhook_url";
+const KEY_LAST_DERIVATION_INDEX: &str = "last_derivation_index";
```
ok300, line 9: If the cache is wiped or lost, it's reconstructed on next startup without affecting any functionality.

This new key IMO cannot survive this. Once set, it has to be remembered. I suppose the plan is real-time sync will take care of this. If so, maybe add a TODO on this key indicating this shoud be covered by real-time sync?

dangeross:
> Also what happens multiple payments are made to the MRH Liquid address, i.e. address re-use? Will those payments be associated with the same swap?

The first one is associated while the swap is ongoing, then any following ones are stored as separate payments

lib/core/src/persist/address.rs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        format!(
+            "
+            SELECT
+                address,
+                expiry_block_height
+            FROM reserved_addresses
+            {where_clause_str}
+            ORDER BY expiry_block_height ASC
```
ok300, line 32: Could it be the query is missing a `LIMIT 1`?

Looks like it can return a list of reserved addresses, but in L12-13 above it's used to fetch a single row.

dangeross:
> What happens when a Receive Swap is paid via the MRH Liquid address _after_ the swap expired? Will this also be associated with the swap?

Like Roei said, the first step should be to check if the invoice has expired, then extract the MRH (I'm not sure if Boltz returns the BIP21 address if the swap is expired, need to check this). But to answer the question, the it won't be associated as the swap is no longer ongoing and therefore will be a separate payment.

lib/core/src/persist/address.rs
dangeross: It is safe to do this as other rows except the first are ignored. But I can add it for clarity
```
    /// If the query returns more than one row, all rows except the first are
    /// ignored.
```
https://github.com/rusqlite/rusqlite/blob/b7309f2dca70716fee44c85082c585b330edb073/src/lib.rs#L684-L704

lib/core/src/sdk.rs
```diff
@@ -1701,10 +1707,16 @@ impl LiquidSdk {
         let receiver_amount_sat = user_lockup_amount_sat - fees_sat;
         let claim_address = self.onchain_wallet.next_unused_address().await?.to_string();
 
+        // Reserve this address until the timeout block height
```
ok300, line 1710: For chain swaps, I'm 99% sure that claims (coop / non-coop) can happen even after the timeout block height. If the user locks up funds on time (i.e. before the timeout block), the swap won't be marked as expired and it can be claimed even after the timeout block.

I'm checking with Boltz to confirm.

If this is true, it would mean the claim LBTC address could still be used by this swap even after it's un-reserved.

lib/core/src/persist/cache.rs
dangeross: Yeah, but the whole pool is lost also (should also be synced). This will be (re)set to the next unused derivation index the next time an address from the wallet is used.

I added to todo as a reminder [70bf17b](https://github.com/breez/breez-sdk-liquid/pull/531/commits/70bf17bd8ee3833969f07da8a28df8c0fb307419)

lib/core/src/persist/address.rs
dangeross: Added LIMIT 1 [70bf17b](https://github.com/breez/breez-sdk-liquid/pull/531/commits/70bf17bd8ee3833969f07da8a28df8c0fb307419)

lib/core/src/sdk.rs
```diff
@@ -1701,10 +1707,16 @@ impl LiquidSdk {
...
+        self.persister.insert_or_update_reserved_address(
```
roeierez, line 1711: @dangeross why do we need to cache claim_address in the pool? Why do we care if such claim address is reused?

dangeross: If the swap is never initiated (btc sent to lockup address) and the swap expires, then this claim address can be reused elsewhere.

lib/core/src/sdk.rs
ok300: Looks like I was wrong, claiming is only possible until the timeout block height. After that only refunds are allowed.

dangeross: Thanks for verifying @ok300

lib/core/src/sdk.rs
roeierez: But we only care if it is used as part of magic hints and for that we already reserve addresses. I think we don't care if it is used in other places (other claims, refunds, etc...) don't we?
Also we can defer the claim creation to the actual point we know we need to claim (btc sent to lockup address) but even if not, as stated above, we should be fine.

dangeross: Makes sense, now deferring getting the incoming claim address [990ff5c](https://github.com/breez/breez-sdk-liquid/pull/531/commits/990ff5ca99052fcc50aa300018c438d906080152)

roeierez (changes requested):
Looking good.

lib/core/src/persist/receive.rs
```diff
@@ -44,15 +44,31 @@ impl Persister {
...
             &receive_swap.receiver_amount_sat,
             &receive_swap.created_at,
             &receive_swap.claim_fees_sat,
-            &receive_swap.claim_tx_id,
+            &receive_swap.mrh_address,
+            &receive_swap.mrh_script_pubkey,
             &receive_swap.state,
         ))?;
 
+        con.execute(
```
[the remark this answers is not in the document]
dangeross: rusqlite has a limit of 16 params, so it is split into two statements

lib/core/src/persist/cache.rs
```diff
@@ -65,6 +67,18 @@ impl Persister {
     pub fn get_webhook_url(&self) -> Result<Option<String>> {
         self.get_cached_item(KEY_WEBHOOK_URL)
     }
+
+    pub fn set_last_derivation_index(&self, index: u32) -> Result<()> {
+        self.update_cached_item(KEY_LAST_DERIVATION_INDEX, index.to_string())
+    }
+
+    pub fn get_last_derivation_index(&self) -> Result<Option<u32>> {
```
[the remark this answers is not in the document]
dangeross: If there is no index set, it passes `None` to `address(next_index)` which returns the next available address and it's index which we then cache 
```
let address_result = self.wallet.lock().await.address(next_index)?;
```

lib/core/src/wallet.rs
```diff
@@ -206,7 +211,31 @@ impl OnchainWallet for LiquidOnchainWallet {
...
+            None => {
+                let next_index = self.persister.next_derivation_index()?;
+                let address_result = self.wallet.lock().await.address(next_index)?;
+                let address = address_result.address().clone();
+                let index = address_result.index();
+                debug!(
+                    "Got unused address {} with derivation index {}",
+                    address, index
+                );
+                self.persister.set_last_derivation_index(index)?;
```
roeierez, line 233: It seems that we don't need this function as the next_derivation_index already increments and update the index.
I suggest to remove it so the only way to update the index is actually by incrementing it.

roeierez (changes requested):
Almost good to go, one comment.

lib/core/src/wallet.rs
dangeross: @roeierez the cached last derivation index will never be set (`next_derivation_index()` will always return None) if we remove it. How about we only set it if `next_derivation_index()` returns None?

roeierez: I see, it is here for the migration case.
In that case shouldn't we set the index to index + 1 ? It seems that the persisted value should be the next available (not used) index.

roeierez (approved):

lib/core/src/persist/receive.rs
```diff
@@ -204,17 +246,25 @@ impl Persister {
                         WHEN lockup_tx_id IS NULL THEN :lockup_tx_id
                         ELSE lockup_tx_id
                     END,
+                mrh_tx_id = 
+                    CASE
+                        WHEN mrh_tx_id IS NULL THEN :mrh_tx_id
+                        ELSE mrh_tx_id
+                    END,
+                payer_amount_sat = COALESCE(:mrh_amount_sat, payer_amount_sat),
+                receiver_amount_sat = COALESCE(:mrh_amount_sat, receiver_amount_sat),
```
ok300, line 255: If a Receive Swap
- is paid via MRH, these two `COALESCE` lines will set the swap's `payer_amount_sat` and `receiver_amount_sat`
- if the swap is then also paid via Liquid swap, the swap's `payer` and `receiver` amounts remain unchanged, which means the Liquid swap's amounts will not be persisted

If it's paid twice but in the other order, the later MRH payment will overwrite the swap's initial payer/receiver amounts, which means they will be lost.

In both cases, the swap will be modified (because this is a `UPDATE receive_swaps`) in an inaccurate way.

lib/core/src/persist/mod.rs
```diff
@@ -204,9 +203,9 @@ impl Persister {
             FROM payment_tx_data AS ptx          -- Payment tx (each tx results in a Payment)
             FULL JOIN (
                 SELECT * FROM receive_swaps
-                WHERE claim_tx_id IS NOT NULL OR lockup_tx_id IS NOT NULL
+                WHERE COALESCE(claim_tx_id, lockup_tx_id, mrh_tx_id) IS NOT NULL
             ) rs                                 -- Receive Swap data (by claim)
-                ON ptx.tx_id = rs.claim_tx_id
+                ON ptx.tx_id in (rs.claim_tx_id, rs.mrh_tx_id)
```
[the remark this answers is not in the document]
dangeross: This case will not happen now after this commit [1404466](https://github.com/breez/breez-sdk-liquid/pull/531/commits/140446612bf429b445f133b78cf022b3291c4c3a). 
- If the MRH tx is seen, then the lightning payment swap updates will be ignored and not claimed (lockup refunded to Boltz)
- If the swap has a lockup/claim tx (lightning payment), then the MRH tx will be a separate payment

See below two payments, the SDK first seeing the lightning payment (lockup tx set), then the MRH payment seen:
```
{
      "destination": "001439f07c3da2b3be716ca409ceaa7c6daa4d5c83b8",
      "tx_id": "ed0aceaf3243e2a824561e126fad7c73d16661d76c66a77d9e1e6ad36e55363c",
      "timestamp": 1729685107,
      "amount_sat": 957,
      "fees_sat": 0,
      "payment_type": "Receive",
      "status": "Complete",
      "details": {
        "Liquid": {
          "destination": "001439f07c3da2b3be716ca409ceaa7c6daa4d5c83b8",
          "description": "Liquid transfer"
        }
      }
    },
    {
      "destination": "lnbc10u1pn33e3zpp5xxtdd4e6dqdvmf8k98q7rw7t9la02vutlh67c32g7m00m89sznyqdpz2djkuepqw3hjqnpdgf2yxgrpv3j8yetnwvcqz95xqyp2xqrzjqd8tadv2r72xyjc7tr2ehpv3dpy6maufy8wd6z4y77djkxekuhqhqzzxeyqq28qqqqqqqqqqqqqqq9gq2ysp52w8alcqyvvqkwcczvvh9agcz247lhh63nxh83tv5ns2uts9dv5as9qxpqysgqlvphsn8xftqal6m4v893euwvw2cexg9z33prrjfhg7vgcg6hnf7qmfj5rjwkr4z07frfwu3zv8fzw4ww7nnqpkmg3xuwgma7shg72rqpap94en",
      "tx_id": "a3e58ca601d95d4af59c7fc86a05aaa6320afcbad573586f02d0a076070198d6",
      "timestamp": 1729685026,
      "amount_sat": 957,
      "fees_sat": 43,
      "payment_type": "Receive",
      "status": "Complete",
      "details": {
        "Lightning": {
          "swap_id": "vC27WVRuWqLw",
          "description": "Send to L-BTC address",
          "preimage": null,
          "bolt11": "lnbc10u1pn33e3zpp5xxtdd4e6dqdvmf8k98q7rw7t9la02vutlh67c32g7m00m89sznyqdpz2djkuepqw3hjqnpdgf2yxgrpv3j8yetnwvcqz95xqyp2xqrzjqd8tadv2r72xyjc7tr2ehpv3dpy6maufy8wd6z4y77djkxekuhqhqzzxeyqq28qqqqqqqqqqqqqqq9gq2ysp52w8alcqyvvqkwcczvvh9agcz247lhh63nxh83tv5ns2uts9dv5as9qxpqysgqlvphsn8xftqal6m4v893euwvw2cexg9z33prrjfhg7vgcg6hnf7qmfj5rjwkr4z07frfwu3zv8fzw4ww7nnqpkmg3xuwgma7shg72rqpap94en",
          "payment_hash": "3196d6d73a681acda4f629c1e1bbcb2ffaf5338bfdf5ec4548f6defd9cb014c8",
          "refund_tx_id": null,
          "refund_tx_amount_sat": null
        }
      }
    }
```

lib/core/src/persist/receive.rs
dangeross: See https://github.com/breez/breez-sdk-liquid/pull/531#discussion_r1812628295

ok300 (approved):
Looks good.

The only remaining NIT is the question related to real-time sync, but IMO it can be handled separaately if necessary.

lib/core/src/persist/address.rs
```diff
@@ -0,0 +1,151 @@
...
+        let tx = con.transaction()?;
+        Self::delete_reserved_address_inner(&tx, address)?;
+        tx.commit()?;
+
+        Ok(())
+    }
+
+    fn delete_reserved_address_inner(tx: &Transaction, address: &str) -> Result<(), PaymentError> {
+        tx.execute(
+            "DELETE FROM reserved_addresses WHERE address = ?",
```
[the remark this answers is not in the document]
dangeross: The reserved address table currently only stores MRH addresses used in receive swaps. These are the two cases where a delete would occur:

- A pay to the MRH address is seen while there is no swap progress.
- The reserved address block height expires and the address is reused for another purpose (claim, refund, MRH, direct address). In the MRH case it is re-added to the reserved address table with a later expiry.

Also to note is that the `last_derivation_index` from the cache needs to be synced.

Maybe it's best to add this as a task of the real-time sync tracking issue

PR #532, opened 2024-10-17
`GetInfoResponse`: add fingerprint

ok300:
This PR adds the wallet's `fingerprint` to `GetInfoResponse`. This is useful because the fingerprint is used in building the working directory. It therefore helps callers better identify which working directory belongs to which wallet, in setups where several wallets (mnemonics) were used.


dangeross:
This pubkey is currently used to register a webhook with the LNURL-pay service

ok300:
@dangeross good catch.

If it's acting as an unique identifier of the wallet, would it make sense to use the fingerprint instead? Or is there a reason to use the pubkey in particular?

dangeross:
> If it's acting as an unique identifier of the wallet, would it make sense to use the fingerprint instead? Or is there a reason to use the pubkey in particular

The LNURL service receives a signed request with a signature which is validated against the pubkey. The pubkey also needs to be "static" so the encoded LNURL doesn't change.

dangeross:
Adding the fingerprint is good though, it solves an issue that @erdemyerebasmaz just found trying to export the database in Misty

ok300:
Updated the PR to keep the pubkey and only add the fingerprint.

erdemyerebasmaz (approved):
LGTM

dangeross (approved):
LGTM

PR #537, opened 2024-10-25
Bring remaining u32 amount types to u64

ok300:
This PR fixes older TODOs about amount type conversions: some amounts and fees are `u64`, while others were `u32`.

Fixing this also simplifies the code for amountless swaps, which deals with `Option<u64>`.

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -539,7 +539,7 @@ dependencies = [
 [[package]]
 name = "boltz-client"
 version = "0.1.3"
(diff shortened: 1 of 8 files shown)


ok300:
This is now using the latest upstream `trunk` which includes
- @hydra-yse 's commit in the previous `yse-fee-calculation` branch
- my commit for adding `u64` support across all create requests

dangeross (approved):
LGTM

PR #538, opened 2024-10-25
Add zero-amount Receive Chain Swap

### LINKED ISSUE #519: Support amountless BTC -> Liquid swap
Boltz now supports chain swaps w/o an amount: https://github.com/BoltzExchange/boltz-backend/pull/659

1. allow to create a new chain swap without an amount
2. when sats are sent to the lockup address, give the client a quote and if they agree, the swap happens

We had a few partners asking for this.

ok300:
Fixes #519 

Open questions:

- [x] Proper way to show fees during prepare/pay: "X sat + Y% of the sent amount" ?
- [x] Include min/max in `prepare_response`? How else to communicate them to caller, as they are now essential?
  - Also: it's important they exactly match the ones for the pair hash used in creating the swap
- [x] Should such a (refundable) swap require user interaction to "try to claim"?
  - Or should the SDK try to claim automatically when a lockup is detected (or on `lockupFailed` event)?
- [x] How to reflect in DB: set `payer_amount_sat` to 0 for zero-amount swaps? Or `NULL` (DB migration)?


dangeross:
> Should such a (refundable) swap require user interaction to "try to claim"?
> * Or should the SDK try to claim automatically when a lockup is detected (or on `lockupFailed` event)?

The SDK user has to provide a fee rate for these refunds, so they need to be manually handled

ok300:
It's not about refunds per se. The zero-amount swaps get the `TransactionLockupFailed` event when the user locks up funds. At this point, it would be indistinguishable from a refundable.

However it will not become an actual refundable when the user locks up funds, because when the zero-amount swap gets this event, we are getting and confirming the so-called quote from Boltz. By confirming it, we tell Boltz they should proceed with this swap (e.g. create the server lockup tx, etc). My question above was about this "getting and confirming" the quote. However we clarified in an earlier call that no user interaction should be needed to claim, so therefore we get and auto-confirm the quote.

Hopefully this clarifies the confusion.

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -407,9 +407,12 @@ dictionary PrepareReceiveRequest {
 };
 
 dictionary PrepareReceiveResponse {
-    u64? payer_amount_sat;
     PaymentMethod payment_method;
     u64 fees_sat;
+    u64? payer_amount_sat;
+    u64? zero_amount_min_payer_amount_sat;
+    u64? zero_amount_max_payer_amount_sat;
+    f64? zero_amount_service_feerate;
```
roeierez, line 415: The prefix zero_amount is a bit confusing IMO.
Why do we need this prefix? Why not showing these values on all cases?
min/max and service fees seem reasonable to be shown also on cases where amount is set.

lib/core/src/chain_swap.rs
```diff
@@ -367,6 +367,28 @@ impl ChainSwapHandler {
...
+                if matches!(swap_state, ChainSwapStates::TransactionLockupFailed) && is_zero_amount
+                {
+                    if let Err(e) = self
+                        .swapper
+                        .get_zero_amount_chain_swap_quote(&swap.id)
+                        .map(|quote| quote.to_sat())
+                        .and_then(|quote| {
+                            info!("Got quote of {quote} sat for swap {}", &swap.id);
+
+                            self.swapper
```
roeierez, line 381: Should we also update the swap payer_amount here?

roeierez (changes requested):
Looks good. some comments about the interface.

lib/core/src/chain_swap.rs
ok300: Makes sense, done: 35cb343b7b310e15f32bde5202f49c7fa4109a11

lib/bindings/src/breez_sdk_liquid.udl
ok300: The only weird case is when `PrepareReceiveResponse.payment_method` is `LiquidAddress`. Then min/max don't mean much.

Should I then set them to `None` in that case?

lib/core/src/chain_swap.rs
```diff
@@ -367,6 +367,29 @@ impl ChainSwapHandler {
...
+                            self.persister.update_swap_payer_amount(&swap.id, quote)?;
+                            self.swapper
+                                .accept_zero_amount_chain_swap_quote(&swap.id, quote)
+                        })
+                    {
+                        warn!("Failed to accept the quote for swap {}: {e:?}", &swap.id);
+                    }
+
+                    // We successfully accepted the quote, the swap should continue as normal
+                    return Ok(()); // Break from TxLockupFailed branch
```
dangeross, line 390: Seems the error case arrives here too, we should allow it to be set to Refundable

cli/src/commands.rs
```diff
@@ -263,13 +263,21 @@ pub(crate) async fn handle_command(
...
-            );
+            let fees = prepare_response.fees_sat;
+            let confirmation_msg = match payer_amount_sat {
+                Some(_) => format!("Fees: {fees} sat. Are the fees acceptable? (y/N)"),
+                None => {
+                    let min = prepare_response.zero_amount_min_payer_amount_sat.unwrap();
+                    let max = prepare_response.zero_amount_max_payer_amount_sat.unwrap();
+                    let service_feerate = prepare_response.zero_amount_service_feerate.unwrap();
+                    format!(
+                        "Fees: {fees} sat + {service_feerate}% of the sent amount. \
```
dangeross, line 274: I think the service feerate is going to be a `0.1` type value

lib/core/src/sdk.rs
```diff
@@ -1770,9 +1786,12 @@ impl LiquidSdk {
     }
 
     /// Receive from a Bitcoin transaction via a chain swap.
+    ///
+    /// If no `user_lockup_amount_sat` is specified, this is an amountless swap and `fees_sat` exclude
```
dangeross, line 1790:
```suggestion
    /// If no `payer_amount_sat ` is specified, this is an amountless swap and `fees_sat` exclude
```

lib/core/src/persist/chain.rs
```diff
@@ -263,6 +263,43 @@ impl Persister {
...
+                err: format!("Cannot update non-existent chain swap with ID: {swap_id}"),
+            })?;
+        ensure_sdk!(
+            matches!(swap.direction, Direction::Incoming),
+            PaymentError::Generic {
+                err: format!(
+                    "Can only update payer_amount_sat for incoming chain swaps. Swap ID: {swap_id}"
+                )
+            }
+        );
```
dangeross, line 285: This could be a part of the SQL where clause below

lib/core/src/chain_swap.rs
```diff
@@ -367,6 +367,29 @@ impl ChainSwapHandler {
...
+                if matches!(swap_state, ChainSwapStates::TransactionLockupFailed) && is_zero_amount
+                {
+                    if let Err(e) = self
+                        .swapper
+                        .get_zero_amount_chain_swap_quote(&swap.id)
+                        .map(|quote| quote.to_sat())
+                        .and_then(|quote| {
+                            info!("Got quote of {quote} sat for swap {}", &swap.id);
+
+                            self.persister.update_swap_payer_amount(&swap.id, quote)?;
```
dangeross, line 381: It looks like the quote here is the [server lockup amount](https://github.com/BoltzExchange/boltz-backend/commit/b93f0e8a0ef61c47b2160600d588294c75ae9d58#diff-05ebf780e4d6d2c474c4a231242487fcc27906cb3672d657a20267b91803fa89R47). If so, I think there needs to be some recalculations done. The receiver amount (quote - claim fees), fees (server fees + boltz fees + claim fees) and payer amount (receiver amount + claim fees + server fees + boltz fees) need calculating.

lib/bindings/src/breez_sdk_liquid.udl
roeierez: Yes I think that makes sense.

lib/core/src/sdk.rs
ok300: Essentially correct, but IMO the arg has to be adapted instead of the comment, as `user_lockup_amount_sat` is more accurate: d3f448819d81589e0a7b221880a419b54cebd9ca

lib/bindings/src/breez_sdk_liquid.udl
ok300: Done: 1e18406e28ae4f16adff8cb4f9d525ce0a6aff55

lib/core/src/chain_swap.rs
ok300: Done: 5a44cafb8be712244783f41c62b4318d1ba4937a

cli/src/commands.rs
ok300: Yes, the service feerate is already the percentage value (i.e. `0.1 %`).

lib/core/src/persist/chain.rs
ok300: Unfortunately if I do that, we won't get an error message if this is called for a swap of the wrong direction.

That's why I chose to explicitly validate.

lib/core/src/model.rs
```diff
@@ -306,18 +306,18 @@ pub struct PrepareReceiveResponse {
...
-    pub zero_amount_max_payer_amount_sat: Option<u64>,
+    /// When the method is [PaymentMethod::LiquidAddress], this is empty.
+    pub max_payer_amount_sat: Option<u64>,
 
     /// The percentage of the sent amount that will count towards the service fee.
     ///
-    /// Only applicable to Zero-Amount Receive Chain Swaps.
-    pub zero_amount_service_feerate: Option<f64>,
+    /// When the method is [PaymentMethod::LiquidAddress], this is empty.
+    pub service_feerate: Option<f64>,
```
dangeross, line 320: For `Lightning` its also empty no?

lib/core/src/persist/chain.rs
```diff
@@ -279,22 +279,27 @@ impl Persister {
...
             PaymentError::Generic {
                 err: format!(
-                    "Can only update payer_amount_sat for incoming chain swaps. Swap ID: {swap_id}"
+                    "Only an incoming chain swap can be a zero-amount swap. Swap ID: {swap_id}"
                 )
             }
         );
 
-        log::info!("Updating chain swap {swap_id}: payer_amount_sat = {payer_amount_sat}");
+        let payer_amount_sat = server_lockup_amount_sat;
```
dangeross, line 287: I assume this `server_lockup_amount_sat` is what Boltz calculates after deducting their service and server fee, so don't we need to calculate the `payer_amount_sat` including those fees?

ok300: > I assume this `server_lockup_amount_sat` is what Boltz calculates after deducting their service and server fee

Yes.

> so don't we need to calculate the `payer_amount_sat` including those fees?

Those fees are not exposed at this point. The feerate is only available before creation, based on the chosen fees pair. Therefore, it seems to me we can only estimate the payer amount at this point in the zero-amount swap.

lib/core/src/chain_swap.rs
```diff
@@ -436,6 +436,32 @@ impl ChainSwapHandler {
...
+                {
+                    let accept_quote_res = self
+                        .swapper
+                        .get_zero_amount_chain_swap_quote(&swap.id)
+                        .map(|quote| quote.to_sat())
+                        .and_then(|quote| {
+                            info!("Got quote of {quote} sat for swap {}", &swap.id);
+
+                            self.persister
+                                .update_zero_amount_swap_values(&swap.id, quote)?;
```
roeierez, line 451: What does the quote represent here? Is it the boltz lockup? If so how do we know they don't cheat us with extra service fee? Don't we need to validate that?

roeierez (changes requested):

lib/core/src/chain_swap.rs
ok300: The quote represents the server lockup amount, i.e. "this is how much you can claim, if you choose to accept this quote".

How we normally validate the fees for Receive Chain Swaps is during creation:
https://github.com/breez/breez-sdk-liquid/blob/98864c70a6055c74d5b60142548de8a9584b7b93/lib/core/src/sdk.rs#L1913-L1916

However, for this the swap amount has to be known on creation, and in this case it's not.

I didn't add validation because we cannot do it with the data we know.

To be able to do this for zero-amount swaps, we would need:
- to store the swap's fee pair details (claim fee estimate, server fee, service fee rate)
- to fetch the actual user lockup amount from the swap's lockup address
- to calculate the expected server lockup amount and compare with the quote

Should I add this?

roeierez: I think we don't have a choice, otherwise we are committing to a swap by trusting the swapper only.
By doing so we will also be able to calculate properly the payer amount.

ok300: Done, added validation before accepting the quote from Boltz.

We also now correctly persist the payer and receiver amounts.

lib/core/src/persist/chain.rs
ok300: Done, `payer_amount_sat` is now calculated (fetched as the lockup chain balance) and persisted:

https://github.com/breez/breez-sdk-liquid/blob/618879d551b385f5d70b331725e217af196201c7/lib/core/src/chain_swap.rs#L517-L524

lib/core/src/model.rs
```diff
@@ -303,6 +304,21 @@ pub struct PrepareReceiveResponse {
...
+
+    /// The maximum amount the payer can send for this swap to succeed.
+    ///
+    /// When the method is [PaymentMethod::LiquidAddress], this is empty.
+    pub max_payer_amount_sat: Option<u64>,
+
+    /// The percentage of the sent amount that will count towards the service fee.
+    ///
+    /// When the method is [PaymentMethod::LiquidAddress], this is empty.
+    pub service_feerate: Option<f64>,
```
dangeross, line 321: Maybe to be consistent with the previous PR, this could be `swapper_feerate`?

dangeross (commented):
Looking good, just a couple of comments

lib/core/src/chain_swap.rs
```diff
@@ -466,6 +480,82 @@ impl ChainSwapHandler {
...
+        ensure_sdk!(
+            user_lockup_amount_sat > fees_sat,
+            PaymentError::generic(&format!("Invalid quote: fees ({fees_sat} sat) are higher than user lockup ({user_lockup_amount_sat} sat)"))
+        );
+
+        let expected_server_lockup_amount_sat = user_lockup_amount_sat - fees_sat;
+        debug!("user_lockup_amount_sat = {}, service_fees_sat = {}, server_fees_sat = {}, expected_server_lockup_amount_sat = {}, quote_server_lockup_amount_sat = {}",
+            user_lockup_amount_sat, service_fees_sat, swapper_server_fees_sat, expected_server_lockup_amount_sat, quote_server_lockup_amount_sat);
+        ensure_sdk!(
+            expected_server_lockup_amount_sat == quote_server_lockup_amount_sat,
```
roeierez, line 545: Perhaps be more tolerant and check if <= ?

roeierez (changes requested):
Looking good.

lib/core/src/chain_swap.rs
ok300: Done: 4576ad9f898f2ef586ab256452360e3ae2fcfb98

lib/core/Cargo.toml
```diff
@@ -14,7 +14,8 @@ frb = ["dep:flutter_rust_bridge"]
 [dependencies]
 anyhow = { workspace = true }
 bip39 = "2.0.0"
-boltz-client = { git = "https://github.com/SatoshiPortal/boltz-rust", branch = "trunk" }
+#boltz-client = { git = "https://github.com/SatoshiPortal/boltz-rust", branch = "trunk" }
+boltz-client = { git = "https://github.com/ok300/boltz-rust", branch = "ok300-support-quote" }
```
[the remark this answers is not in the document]
ok300: Done: e8ac979871dae459ff4cf0997f75cf916cf14322

cli/src/commands.rs
```diff
@@ -271,13 +271,21 @@ pub(crate) async fn handle_command(
...
-                "Payment receive halted"
-            );
+            let fees = prepare_response.fees_sat;
+            let confirmation_msg = match payer_amount_sat {
+                Some(_) => format!("Fees: {fees} sat. Are the fees acceptable? (y/N)"),
+                None => {
+                    let min = prepare_response.min_payer_amount_sat;
+                    let max = prepare_response.max_payer_amount_sat;
+                    let service_feerate = prepare_response.service_feerate;
+                    format!(
```
roeierez, line 281: I guess we will need to document the fact that in amountless swap applications need to add the service fee on top of the fees_sat.

ok300: Done: 542dec789e748cae6da29c4ea516086ef8565785

roeierez (approved):

PR #541, opened 2024-10-28
Send: Fix double-lockup when payment is TimedOut

hydra-yse:
Given the current event loop, TimedOut payments may cause lockups to be broadcast. 

This PR changes the send flow such that:
1. When a user calls `send_payment`, the lockup is broadcasted within that same thread _only if_ it hasn't been broadcasted beforehand
2. The background thread (Boltz status stream) does not execute lockups any longer

~**TODO:** Persist `lockup_tx_id` beforehand to ensure we don't broadcast twice in case the client closes the application before the new state is saved~

### DIFF
--- a/lib/core/src/persist/send.rs
+++ b/lib/core/src/persist/send.rs
@@ -241,6 +241,49 @@ impl Persister {
 
         Ok(())
     }
+
+    pub(crate) fn set_send_swap_lockup_tx_id(
+        &self,
+        swap_id: &str,
+        lockup_tx_id: &str,
+    ) -> Result<(), PaymentError> {
+        let con = self.get_connection()?;
+
+        let row_count = con
+            .execute(
+                "UPDATE send_swaps
+                SET lockup_tx_id = :lockup_tx_id
+                WHERE id = :id AND lockup_tx_id IS NULL",
+                named_params! {
+                    ":id": swap_id,
+                    ":lockup_tx_id": lockup_tx_id,
+                },
+            )
+            .map_err(|_| PaymentError::PersistError)?;
+        match row_count {
+            1 => Ok(()),
+            _ => Err(PaymentError::PaymentInProgress),
+        }
+    }
+
+    pub(crate) fn unset_send_swap_lockup_tx_id(
+        &self,
+        swap_id: &str,
+        lockup_tx_id: &str,
+    ) -> Result<(), PaymentError> {
+        let con = self.get_connection()?;
+        con.execute(
+            "UPDATE send_swaps
+            SET lockup_tx_id = NULL
+            WHERE id = :id AND lockup_tx_id = :lockup_tx_id",
+            named_params! {
+                ":id": swap_id,
+                ":lockup_tx_id": lockup_tx_id,
+            },
+        )
+        .map_err(|_| PaymentError::PersistError)?;
+        Ok(())
+    }
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -1006,14 +1006,20 @@ impl LiquidSdk {
 
         let swap = match self.persister.fetch_send_swap_by_invoice(invoice)? {
             Some(swap) => match swap.state {
+                Created => swap,
+                TimedOut => {
+                    self.send_swap_handler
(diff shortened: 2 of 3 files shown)


lib/core/src/sdk.rs
```diff
@@ -1006,14 +1006,20 @@ impl LiquidSdk {
 
         let swap = match self.persister.fetch_send_swap_by_invoice(invoice)? {
             Some(swap) => match swap.state {
+                Created => swap,
```
roeierez, line 1009: This looks correct to me but I think it is not a complete fix. Here are some points that need to be addressed IMO:
 1. It looks like we don't check before broadcast that we already broadcasted. I think we must check it, we also need to check the current swap status before sending lockup and ensure it is in initial state (Created).
 2. I can see races when the notification job gets notify and attempt to handle state transition in parallel to the app, it seems that if both handle the invoice.set state at the same time, double broadcast may happen. Perhaps we should disable broadcasting the lockup from the job or think on another solution to handle this point.

roeierez (changes requested):

lib/core/src/sdk.rs
hydra-yse: > It looks like we don't check before broadcast that we already broadcasted.

We check if it's present locally, but perhaps we should also make an onchain query

> ensure it is in initial state (Created)

Agreed

> More specifically, the requirement for broadcasting a lockup should be only the process where the send was invoked.

Agreed

lib/core/src/sdk.rs
```diff
@@ -1078,12 +1078,54 @@ impl LiquidSdk {
         };
         self.status_stream.track_swap_id(&swap.id)?;
 
-        let accept_zero_conf = swap.get_boltz_create_response()?.accept_zero_conf;
+        let create_response = swap.get_boltz_create_response()?;
+        self.try_lockup(&swap, &create_response).await?;
```
roeierez, line 1082: It seems that there is still a race where two separate calls to send_payment for the same invoice may end up in broadcasting two lockup transactions. It might happen when both of them are broadcasting in the same time.

roeierez (changes requested):
Looks good, I think there is one small race issue we need to solve.

lib/core/src/sdk.rs
hydra-yse: Yes I agree, I've added a line **TODO** above to fix this. It's similar to Ross' PR

roeierez: I think we can be simpler here than the claim mechanism. Perhaps we only need to mark this payment in memory as in progress and prevent other callers to lockup as long as it is marked as such?

hydra-yse: Wouldn't adding a `lockup_tx_id` field be the same as marking it?

roeierez: Yes whatever is more simple.

lib/core/src/persist/send.rs
```diff
@@ -241,6 +241,49 @@ impl Persister {
...
+                WHERE id = :id AND lockup_tx_id IS NULL",
+                named_params! {
+                    ":id": swap_id,
+                    ":lockup_tx_id": lockup_tx_id,
+                },
+            )
+            .map_err(|_| PaymentError::PersistError)?;
+        match row_count {
+            1 => Ok(()),
+            _ => Err(PaymentError::AlreadyClaimed),
```
roeierez, line 265: The error is wrong as we are talking about the lockup.

roeierez (changes requested):
Small comment, otherwise looks good to me.

roeierez (approved):

lib/core/src/send_swap.rs
```diff
@@ -223,18 +192,31 @@ impl SendSwapHandler {
                 create_response.expected_amount,
             )
             .await?;
+        let lockup_tx_id = lockup_tx.txid().to_string();
+
+        self.persister
```
ok300, line 197: Why not persist it only after successful broadcast below (in the `Ok` branch)?

Then I think you wouldn't need to unset it in case of error.

hydra-yse: We are "locking" the transaction so that the double lockup cannot occur. We persist before by "setting" so that only the current thread can lockup for the same swap. If the broadcast fails, we unset so other threads can try locking up instead.

ok300 (approved):

PR #542, opened 2024-10-29
Prevent swap double claim

dangeross:
This PR prevents a double claim by:
- Checking the swap's `claim_tx_id` immediately before creating a claim tx
- Storing the txid for the created claim tx before broadcasting using a function that prevents overwriting the `claim_tx_id`
- Broadcasts the claim tx only if the storing of the txid is successful
- Unset the `claim_tx_id` if chain service/swapper broadcasts both fail


hydra-yse (approved):
utACK

PR #549, opened 2024-11-06
Add support for paying to BOLT12

ok300:
This PR adds support for paying to a BOLT12 offer.

The flow is as follows:
- user specifies a BOLT12 offer destination when paying
- SDK calls Boltz endpoint that fetches a BOLT12 invoice for given offer
- SDK calls `send_payment_via_swap` with the BOLT12 invoice as argument, instead of the usual BOLT11

Open questions
- [x] Should we also support paying directly to BOLT12 invoices, in addition to BOLT12 offers?
- [x] Expand `PaymentDetails::Lightning` with a new field for `bolt12_invoice`? And / or an optional `bolt12_offer`?
  - Or alternatively rename the existing `bolt11` field to `invoice`, so it means either BOLT11 or BOLT12 invoice?
- [x] Should we fetch the invoice during `prepare`? This may take a long time for slow endpoints (Phoenix), causing the user to wait a few seconds for `prepare` to return.
- [x] Update `sdk_common::input_parser` to parse BOLT12 offers / invoice? Or keep changes local to the Liquid SDK?
  - [ ] Use [BIP-353](https://github.com/bitcoin/bips/blob/master/bip-0353.mediawiki#resolution) "DNS Payment Instructions" for BOLT12 resolution? (similar to LN Address)

Open TODOs

- [x] Validate returned invoice (check signature) before trying to pay it
- [x] ~Detect / handle MRH in BOLT12 invoices~
- [x] Handle both BOLT11 and BOLT12 in persister
- [x] When paying to a bolt12 offer, `list_payments` shows payments as `Pending`
- [x] When paying to a bolt12 offer, `list_payments` shows the bolt12 invoice under `bolt11` (should show offer instead)

### DIFF
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -533,7 +533,7 @@ interface GetPaymentRequest {
 
 [Enum]
 interface PaymentDetails {
-    Lightning(string swap_id, string description, string? preimage, string? bolt11, string? payment_hash, string? refund_tx_id, u64? refund_tx_amount_sat);
+    Lightning(string swap_id, string description, string? preimage, string? bolt11, string? bolt12_offer, string? payment_hash, string? refund_tx_id, u64? refund_tx_amount_sat);
     Liquid(string destination, string description);
     Bitcoin(string swap_id, string description, string? refund_tx_id, u64? refund_tx_amount_sat);
 };
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -381,7 +381,7 @@ dictionary PrepareSendRequest {
 interface SendDestination {
     LiquidAddress(LiquidAddressData address_data);
     Bolt11(LNInvoice invoice);
-    Bolt12(string offer);
+    Bolt12(string offer, u64 receiver_amount_sat);
 };
 
 dictionary PrepareSendResponse {
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -45,7 +45,7 @@ interface Amount {
 };
 
 dictionary LNOffer {
-    string bolt12;
+    string offer;
     sequence<string> chains;
     string? description;
     string? signing_pubkey;
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -43,6 +43,7 @@ interface InputType {
     BitcoinAddress(BitcoinAddressData address);
     LiquidAddress(LiquidAddressData address);
     Bolt11(LNInvoice invoice);
+    Bolt12(string offer);
     NodeId(string node_id);
     Url(string url);
     LnUrlPay(LnUrlPayRequestData data);
@@ -380,7 +381,7 @@ dictionary PrepareSendRequest {
 interface SendDestination {
     LiquidAddress(LiquidAddressData address_data);
     Bolt11(LNInvoice invoice);
-    Bolt12(string invoice);
+    Bolt12(string offer);
 };
 
 dictionary PrepareSendResponse {
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -380,6 +380,7 @@ dictionary PrepareSendRequest {
 interface SendDestination {
     LiquidAddress(LiquidAddressData address_data);
     Bolt11(LNInvoice invoice);
+    Bolt12(string invoice);
 };
 
 dictionary PrepareSendResponse {
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -43,7 +43,7 @@ interface InputType {
     BitcoinAddress(BitcoinAddressData address);
     LiquidAddress(LiquidAddressData address);
     Bolt11(LNInvoice invoice);
-    Bolt12(string offer);
+    Bolt12Offer(string offer);
     NodeId(string node_id);
     Url(string url);
     LnUrlPay(LnUrlPayRequestData data);
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -44,9 +44,14 @@ interface Amount {
     Currency(string iso4217_code, u64 fractional_amount);
 };
 
+dictionary LNOfferBlindedPath {
+    sequence<string> blinded_hops;
+};
+
 dictionary LNOffer {
     string offer;
     sequence<string> chains;
+    sequence<LNOfferBlindedPath> paths;
     string? description;
     string? signing_pubkey;
     Amount? min_amount;
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -49,7 +49,7 @@ dictionary LNOffer {
     sequence<string> chains;
     string? description;
     string? signing_pubkey;
-    Amount? amount;
+    Amount? min_amount;
     u64? absolute_expiry;
     string? issuer;
 };
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -39,11 +39,27 @@ dictionary RouteHintHop {
 };
 
 [Enum]
+interface Amount {
+    Bitcoin(u64 amount_msat);
+    Currency(string iso4217_code, u64 fractional_amount);
+};
+
+dictionary LNOffer {
+    string bolt12;
+    sequence<string> chains;
+    string? description;
+    string? signing_pubkey;
+    Amount? amount;
+    u64? absolute_expiry;
+    string? issuer;
+};
+
+[Enum]
 interface InputType {
     BitcoinAddress(BitcoinAddressData address);
     LiquidAddress(LiquidAddressData address);
     Bolt11(LNInvoice invoice);
-    Bolt12Offer(string offer);
+    Bolt12Offer(LNOffer offer);
     NodeId(string node_id);
     Url(string url);
     LnUrlPay(LnUrlPayRequestData data);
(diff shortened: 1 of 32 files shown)


lib/core/src/sdk.rs
```diff
@@ -788,11 +788,54 @@ impl LiquidSdk {
                 };
                 payment_destination = SendDestination::Bolt11 { invoice };
             }
-            _ => {
+            Ok(_) => {
                 return Err(PaymentError::Generic {
                     err: "Destination is not valid".to_string(),
                 });
             }
+            Err(_) => {
```
roeierez, line 796: I think we must add the bolt12 & offer to the input parser. I don't see how would the user app recognize it is a bolt12/offer otherwise.

lib/core/src/sdk.rs
```diff
@@ -788,11 +788,54 @@ impl LiquidSdk {
...
+                // TODO If MRH, fallback to onchain
+
+                let boltz_fees_total = lbtc_pair.fees.total(receiver_amount_sat);
+                let lockup_fees_sat = self
+                    .estimate_lockup_tx_fee_send(receiver_amount_sat + boltz_fees_total)
+                    .await?;
+                fees_sat = boltz_fees_total + lockup_fees_sat;
+
+                payment_destination = SendDestination::Bolt12 {
+                    offer: req.destination.to_string(),
```
roeierez, line 836: Perhaps we should populate here the bolt12 here instead of the offer? This way also the send_payment can pay directly the bolt12 without fetching again it from boltz.

lib/core/src/sdk.rs
```diff
@@ -869,18 +913,24 @@ impl LiquidSdk {
                     .await
             }
             SendDestination::Bolt11 { invoice } => {
-                self.pay_invoice(&invoice.bolt11, *fees_sat).await
+                self.pay_bolt11_invoice(&invoice.bolt11, *fees_sat).await
+            }
+            SendDestination::Bolt12 { offer } => {
+                // TODO Amount hardcoded to 1_000 because we don't yet have it in the request
+                let bolt12_invoice = self.swapper.get_bolt12_invoice(offer, 1_000)?;
```
roeierez, line 920: Perhaps it is better to not fetch again the bolt12? Just pass it in the prepare response?

roeierez (changes requested):

lib/core/src/sdk.rs
roeierez: Actually on second thought why do we need to fetch the bolt12 here in the prepare?

lib/core/src/sdk.rs
roeierez: I am not sure now we need to fetch it in the prepare but I think we should only fetch either here or in the prepare but not both.

lib/core/src/persist/mod.rs
```diff
@@ -312,11 +312,14 @@ impl Persister {
...
-                            maybe_send_swap_invoice
-                                .and_then(|bolt11| get_invoice_description!(bolt11))
-                                .unwrap_or("Lightning payment".to_string())
-                        }),
+
+                        description: "Lightning payment".to_string(),
+                        // TODO Should not assume this is bolt11
+                        // description: maybe_send_swap_description.unwrap_or_else(|| {
+                        //     maybe_send_swap_invoice
+                        //         .and_then(|bolt11| get_invoice_description!(bolt11))
```
dangeross, line 320: Maybe it is safe to remove this now, since 4 months the invoice description (if available) is being added to the persisted swap

lib/core/src/sdk.rs
```diff
@@ -706,7 +706,7 @@ impl LiquidSdk {
     /// # Arguments
     ///
     /// * `req` - the [PrepareSendRequest] containing:
-    ///     * `destination` - Either a Liquid BIP21 URI/address or a BOLT11 invoice
+    ///     * `destination` - Either a Liquid BIP21 URI/address. a BOLT11 invoice or a BOLT12 offer
```
dangeross, line 709:
```suggestion
    ///     * `destination` - Either a Liquid BIP21 URI/address, a BOLT11 invoice or a BOLT12 offer
```

lib/core/src/sdk.rs
```diff
@@ -788,6 +788,26 @@ impl LiquidSdk {
                 };
                 payment_destination = SendDestination::Bolt11 { invoice };
             }
+            Ok(InputType::Bolt12 { offer }) => {
+                let amount_sat = req
+                    .amount_sat
+                    .ok_or_else(|| anyhow!("Expected amount when processing BOLT12 offer"))?;
```
dangeross, line 794: I see the `lightning` crate is upgraded, is it possible to parse the offer and validate the network / amount (if not optional, matches offer amount or take offer amount if request is not set)?

lib/core/src/persist/mod.rs
ok300: Good point, done: 790d8f205658468be994a57c67220f75ab0fbb49

lib/core/src/sdk.rs
ok300: The `offer.amount()` of the parsed `Offer` only [represents](https://github.com/lightningdevkit/rust-lightning/blob/4322b19e619367b371c3b72657c8618a5b9b255e/lightning/src/offers/offer.rs#L620-L623)
> The minimum amount required for a successful payment of a single item

Do you think it makes sense to default to it, if the user chooses no amount?

IMO this is closer to the "pay to Liquid BIP21" use-case, where an user-amount has to be set either in the BIP21 or separately.

lib/core/src/sdk.rs
ok300: You're right. As discussed, removed invoice fetching from prepare.

lib/core/src/sdk.rs
ok300: Done, removed invoice fetching from prepare.

lib/core/src/sdk.rs
ok300: Done.

lib/core/src/sdk.rs
dangeross: Yeah,  you're right

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -397,7 +397,7 @@ dictionary PrepareSendRequest {
 interface SendDestination {
     LiquidAddress(LiquidAddressData address_data);
     Bolt11(LNInvoice invoice);
-    Bolt12(string offer, u64 receiver_amount_sat);
+    Bolt12(LNOffer offer, u64 receiver_amount_sat);
```
dangeross, line 400: I think the `LNOffer` needs defining in the udl

lib/core/src/sdk.rs
```diff
@@ -2525,9 +2584,43 @@ impl LiquidSdk {
...
+                            Amount::Bitcoin {
+                                amount_msat: amount_msats,
+                            }
+                        }
+                        ::lightning::offers::offer::Amount::Currency {
+                            iso4217_code,
+                            amount,
+                        } => Amount::Currency {
+                            iso4217_code: String::from_utf8(iso4217_code.to_vec())
+                                .expect("Expecting a valid ISO 4217 character sequence"),
```
dangeross, line 2609: Should replace with a `.map_err()?`

dangeross (commented):
Looking good

lib/core/src/sdk.rs
```diff
@@ -981,8 +1017,77 @@ impl LiquidSdk {
...
+        ensure_sdk!(
+            receiver_amount_sat == user_specified_receiver_amount_sat,
+            PaymentError::invalid_invoice("Invalid Bolt12 invoice amount")
+        );
+        if let Some(Amount::Bitcoin { amount_msat }) = &offer.min_amount {
+            ensure_sdk!(
+                receiver_amount_sat >= amount_msat / 1_000,
+                PaymentError::invalid_invoice("Invalid Bolt12 invoice amount: below offer minimum")
+            );
+        }
```
dangeross, line 1075: Perhaps we should be doing this in the prepare step, as the invoice creation will likely fail if the request amount in below the offer min amount

ok300: Good point, done: 9920648b1c6a1cd108a4edc8373eeff1f1aec539

roeierez (approved):
Other than @dangeross comments I think it looks good!

erdemyerebasmaz:
Dart changes look good

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -533,7 +533,7 @@ interface GetPaymentRequest {
 
 [Enum]
 interface PaymentDetails {
-    Lightning(string swap_id, string description, string? preimage, string? bolt11, string? payment_hash, string? refund_tx_id, u64? refund_tx_amount_sat);
+    Lightning(string swap_id, string description, string? preimage, string? bolt11, string? bolt12_offer, string? payment_hash, string? refund_tx_id, u64? refund_tx_amount_sat);
```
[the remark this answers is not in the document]
ok300: Done

lib/core/src/persist/migrations.rs
```diff
@@ -183,5 +183,63 @@ pub(crate) fn current_migrations() -> Vec<&'static str> {
...
+            refund_tx_id,
+            created_at,
+            state,
+            description,
+            id_hash,
+            payment_hash
+        FROM send_swaps_old;
+
+        DROP TABLE send_swaps_old;
+        ",
```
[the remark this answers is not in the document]
ok300: Good idea, done: 13ebcfe602febcbd50709e4ffddd402d779848e9

lib/bindings/src/breez_sdk_liquid.udl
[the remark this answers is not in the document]
ok300: Done: 290e04e036e7d3759be944ef9a2507b33951acfe

lib/core/src/persist/mod.rs
```diff
@@ -310,7 +313,11 @@ impl Persister {
                         swap_type: PaymentSwapType::Send,
                         created_at: maybe_send_swap_created_at.unwrap_or(utils::now()),
                         preimage: maybe_send_swap_preimage,
-                        bolt11: maybe_send_swap_invoice.clone(),
+                        bolt11: match maybe_send_swap_bolt12_offer.is_some() {
+                            true => None, // We don't expose the Bolt12 invoice
+                            false => maybe_send_swap_invoice,
+                        },
+                        bolt12_offer: maybe_send_swap_bolt12_offer,
```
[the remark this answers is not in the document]
dangeross: What is the payment destination in the bolt12 case?

dangeross (approved):
LGTM

PR #553, opened 2024-11-07
Liquid/Lightning drain

### LINKED ISSUE #543: Not able to send full balance via Liquid
**Issue**
It's not possible to send the entire wallet's balance, when paying via Liquid or getting fees for this operation.

**Description**
My balance is 100,000 SATS. I'm trying to send 99,750 (fees are 250) - getting `InsufficientFunds` error.
If I'm trying to send 99,749 SATS, then it's ok.
The same for 5,000 balance: trying to send 4,750 - `InsufficientFunds` error.
4,749 - ok.

**Expected behavior**
If I want to send the entire wallet's balance, I should be able to do so.

**Environment**
Breez SDK Liquid for Python, v. 0.3.4
TESTNET

**Comments from the team**
- Ending full balance (`drain`) to Liquid is not supported at the moment. The call always expects there is at least 1 change output, therefore you get the 1 sat change tx. 
- `drain` is currently implemented only for swapping out to BTC.

### LINKED ISSUE #550: Draining all funds with Lightning/Liquid payments
Currently only on-chain payments have a parameter to drain all funds.
To pay to Liquid or Lightning addresses, the max sendable amount to get the wallet completely on 0 can not be obtained in a straightforward way. This difference between the total_balance and the total_balance - fee(max_sendable_amount) is confusing for users, and without the option to easily drain all or to know what their max_sendable_amount is they are kept guessing how much they can actually send to a receiver.

Therefore it would be a nice addition to also make it possible to prepare Liquid and Lightning (Lnurlpay/Lightning addresses/0-amount invoices when supported) payments with the drain all funds option.

This will streamline the APIs and makes it possible to also build consistent user experiences for all payment methods. For example by adding a [max. amount] button on amount input screens when sending.

dangeross:
This PR:
- fixes drain tx (drain tx fee estimate) to use the lowball fee
- estimate fees for liquid/mrh/lightning with fallback to drain fees
- attempt to fallback to drain txs for paying to liquid or send swap lockup
- blocks draining while there are pending payments

It's difficult to add a drain option to the PrepareLnUrlPayRequest because at the point when preparing we don't know if the LNURL invoice response will contain an MRH or not for a direct Liquid tx. Because of this we can't calculate an invoice amount to drain for both options. See #559

TODO
- [x] Add drain option to to prepare_send_payment [80830de](https://github.com/breez/breez-sdk-liquid/pull/553/commits/80830dea80bcac0b1521e66c2bf303e728631ad5)

Fixes #543 
Fixes #550

### DIFF
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -373,7 +373,7 @@ dictionary LnUrlPayRequest {
 
 dictionary PrepareSendRequest {
     string destination;
-    u64? amount_sat = null;
+    PayOnchainAmount? amount = null;
 };
 
 [Enum]
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -373,7 +373,7 @@ dictionary LnUrlPayRequest {
 
 dictionary PrepareSendRequest {
     string destination;
-    PayOnchainAmount? amount = null;
+    PayAmount? amount = null;
 };
 
 [Enum]
@@ -439,13 +439,13 @@ dictionary OnchainPaymentLimitsResponse {
 };
 
 [Enum]
-interface PayOnchainAmount {
+interface PayAmount {
     Receiver(u64 amount_sat);
     Drain();
 };
 
 dictionary PreparePayOnchainRequest {
-    PayOnchainAmount amount;
+    PayAmount amount;
     u32? fee_rate_sat_per_vbyte = null;
 };
 
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -350,7 +350,7 @@ pub struct PrepareSendRequest {
 
     /// Should only be set when paying directly onchain or to a BIP21 URI
     /// where no amount is specified
-    pub amount_sat: Option<u64>,
+    pub amount: Option<PayOnchainAmount>,
 }
 
 /// Specifies the supported destinations which can be payed by the SDK
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -349,7 +349,7 @@ pub struct PrepareSendRequest {
     pub destination: String,
 
     /// Should only be set when paying directly onchain or to a BIP21 URI
-    /// where no amount is specified
+    /// where no amount is specified, or when the caller wishes to drain
     pub amount: Option<PayOnchainAmount>,
 }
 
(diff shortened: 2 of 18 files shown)


lib/core/src/model.rs
```diff
@@ -350,7 +350,7 @@ pub struct PrepareSendRequest {
 
     /// Should only be set when paying directly onchain or to a BIP21 URI
     /// where no amount is specified
```
ok300, line 352:
```suggestion
    /// where no amount is specified, or when the caller wishes to drain
```

lib/core/src/sdk.rs
```diff
@@ -665,34 +662,30 @@ impl LiquidSdk {
         }
     }
 
-    /// Estimate the lockup tx fee for Send swaps
-    async fn estimate_lockup_tx_fee_send(
+    /// Estimate the lockup tx fee
```
ok300, line 665:
```suggestion
    /// Estimate the lockup tx fee for Send and Chain Send swaps
```

ok300 (approved):
Looks good!

roeierez:
> It's difficult to add a drain option to the PrepareLnUrlPayRequest because at the point when preparing we don't know if the LNURL invoice response will contain an MRH or not for a direct Liquid tx. Because of this we can't calculate an invoice amount to drain for both options.

Do you think it is reasonable to calculate for one case (let's say without routing hints) and if we get one with routing hints fetch again?

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -373,7 +373,7 @@ dictionary LnUrlPayRequest {
 
 dictionary PrepareSendRequest {
     string destination;
-    u64? amount_sat = null;
+    PayOnchainAmount? amount = null;
```
roeierez, line 376: The name is a bit confusing because we use onchain for bitcoin.
do you think it makes sense to use amount_sat as before and add the drain option default to false?

dangeross: I feel like the drain option should be an explicit opt-in and would be more consistent with the drain to Bitcoin. I would rather rename the PayOnchainAmount to something like DrainableAmount. Happy to revert to use the amount_sat if you prefer.

dangeross:
> Do you think it is reasonable to calculate for one case (let's say without routing hints) and if we get one with routing hints fetch again?

I think it's a reasonable solution. I thought about just sending the drain amount to the MRH of the first invoice also, but this way seems cleaner

lib/bindings/src/breez_sdk_liquid.udl
roeierez: Perhaps rename to SendPaymentAmount or PayAmount ?

roeierez (approved):

PR #558, opened 2024-11-11
Android: Replaced  with "$rootDir" with "$projectDir"

larcho:
This handles relative paths better. For instance, we use a Monorepo where our `node_modules` is located somewhere else.


dangeross (approved):
LGTM

I tested by making the same changes to the installed dependencies of the example app `packages/react-native/example/node_modules/@breeztech/react-native-breez-sdk-liquid/android/build.gradle`, cleaning then building.

PR #565, opened 2024-11-14
Log errors on starting an SDK instance

erdemyerebasmaz:


lib/core/src/sdk.rs
```diff
@@ -87,7 +87,9 @@ impl LiquidSdk {
             req.config.network == LiquidNetwork::Mainnet,
         )?);
 
-        Self::connect_with_signer(ConnectWithSignerRequest { config: req.config }, signer).await
+        Self::connect_with_signer(ConnectWithSignerRequest { config: req.config }, signer)
+            .inspect_err(|e| error!("Failed to connect with signer: {:?}", e))
```
ok300, line 91:
```suggestion
            .inspect_err(|e| error!("Failed to connect: {:?}", e))
```

NIT: maybe "failed to connect" is clearer?

The first time I saw "Failed to connect to signer" I thought there's an issue with connecting to the signer.

erdemyerebasmaz: Thanks for the suggestion, "Failed to connect" does seem clearer, I've updated the message.

ok300 (approved):

PR #566, opened 2024-11-14
Wipe storage cache in case of data corrupted issue

roeierez:


lib/core/src/wallet.rs
```diff
@@ -104,20 +106,47 @@ impl LiquidOnchainWallet {
...
+        let descriptor = LiquidOnchainWallet::get_descriptor(signer, config.network)?;
+        let mut lwk_persister =
+            FsPersister::new(working_dir.as_ref(), elements_network, &descriptor)?;
+
+        match Wollet::new(elements_network, lwk_persister, descriptor.clone()) {
+            Ok(wollet) => Ok(wollet),
+            Err(lwk_wollet::Error::UpdateHeightTooOld { .. }) => {
+                warn!("Update height too old, wipping storage and retrying");
+                let mut path = working_dir.as_ref().to_path_buf();
+                path.push(elements_network.as_str());
```
dangeross, line 136: I think the passed working_dir is already formatted with the network/fingerprint here: 
https://github.com/breez/breez-sdk-liquid/blob/8f3c303534f4f608afd4a1f38e96a82c5a1c1f68/lib/core/src/sdk.rs#L156
Do we need to add it again?

lib/core/src/wallet.rs
```diff
@@ -104,20 +106,47 @@ impl LiquidOnchainWallet {
...
+                fs::remove_dir_all(&path)?;
+                warn!("Wippping wallet in path: {:?}", path);
```
dangeross, line 138:
```suggestion
                warn!("Wiping wallet in path: {:?}", path);
```

lib/core/src/wallet.rs
```diff
@@ -104,20 +106,47 @@ impl LiquidOnchainWallet {
...
+    ) -> Result<Wollet> {
+        let elements_network: ElementsNetwork = config.network.into();
+        let descriptor = LiquidOnchainWallet::get_descriptor(signer, config.network)?;
+        let mut lwk_persister =
+            FsPersister::new(working_dir.as_ref(), elements_network, &descriptor)?;
+
+        match Wollet::new(elements_network, lwk_persister, descriptor.clone()) {
+            Ok(wollet) => Ok(wollet),
+            Err(lwk_wollet::Error::UpdateHeightTooOld { .. }) => {
+                warn!("Update height too old, wipping storage and retrying");
```
dangeross, line 134:
```suggestion
                warn!("Update height too old, wiping storage and retrying");
```

lib/core/src/wallet.rs
roeierez: Yes but the lwk wallet dir is a nested directory under that directory.

dangeross: Ah right, it's the ElementsNetwork, sorry

dangeross (approved):
LGTM, just the 2 log typos

PR #567, opened 2024-11-14
Configurable cache dir

dangeross:
This PR:
- adds a configurable `cache_dir` to Config used to persist the onchain wallet cache. If the `cache_dir` is not set it defaults to the `working_dir`.
- updates the Notification Plugin to set the `cache_dir` if not set by the application.

### DIFF
--- a/lib/bindings/langs/swift/Sources/BreezSDKLiquid/SDKNotificationService.swift
+++ b/lib/bindings/langs/swift/Sources/BreezSDKLiquid/SDKNotificationService.swift
@@ -20,11 +20,15 @@ open class SDKNotificationService: UNNotificationServiceExtension {
         self.contentHandler = contentHandler
         self.bestAttemptContent = (request.content.mutableCopy() as? UNMutableNotificationContent)
                 
-        guard let connectRequest = self.getConnectRequest() else {
+        guard var connectRequest = self.getConnectRequest() else {
             if let content = bestAttemptContent {
                 contentHandler(content)
             }
             return
+        }
+
+        if connectRequest.config.cacheDir == nil {
+            connectRequest.config.cacheDir = URL(filePath: connectRequest.config.workingDir).appendingPathComponent("pluginCache").path
         }
         
         if let currentTask = self.getTaskFromNotification() {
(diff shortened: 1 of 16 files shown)


dangeross:
Tested a LNURL-pay via notifications on Android with a wallet having lots of transactions. I didn't see any noticeable difference. 

I'm just wondering if we're going to run into any issues if it's not fully synced. Balance is calculated from payments, MRH address is from the address pool or cached derivation index...

roeierez:
The address pool doesn't seem like an issue to me as it is persisted in the db  and shared

roeierez (approved):

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -307,6 +307,7 @@ dictionary Config {
     string bitcoin_electrum_url;
     string mempoolspace_url;
     string working_dir;
+    string? cache_dir;
     LiquidNetwork network;
```
hydra-yse, line 311: NIT: Move the optional env variable to the end of the dict?

hydra-yse (approved):
Small NIT, otherwise LGTM

lib/bindings/langs/swift/Sources/BreezSDKLiquid/SDKNotificationService.swift
```diff
@@ -20,12 +20,16 @@ open class SDKNotificationService: UNNotificationServiceExtension {
...
+        guard var connectRequest = self.getConnectRequest() else {
             if let content = bestAttemptContent {
                 contentHandler(content)
             }
             return
         }
+
+        if connectRequest.config.cacheDir == nil {
+            connectRequest.config.cacheDir = URL(filePath: connectRequest.config.workingDir).appendingPathComponent("pluginCache").path
+        }
```
erdemyerebasmaz, line 32:
```
Failed to build iOS app
Swift Compiler Error (Xcode): 'init(filePath:directoryHint:relativeTo:)' is only available in application extensions for iOS 16.0 or newer
/Users/runner/work/misty-breez/misty-breez/l-breez/ios/.symlinks/plugins/flutterbreez*_liquid/ios/Sources/BreezSDKLiquid/SDKNotificationService.swift:30:45
```

PR #568, opened 2024-11-16
Real-time Sync: Add push logic and run method

### LINKED ISSUE #504: Real-time Sync: Add a method to push changes

### LINKED ISSUE #505: Real-time Sync: Add a method to clean incoming and outgoing changes

hydra-yse:
Based on #555 and #556.   
Closes #504, closes #505.

 **TODO:**
- [x] Add unit tests

### DIFF
--- a/lib/core/src/persist/receive.rs
+++ b/lib/core/src/persist/receive.rs
@@ -2,20 +2,23 @@ use std::collections::HashMap;
 
 use anyhow::Result;
 use boltz_client::swaps::boltz::CreateReverseResponse;
-use rusqlite::{named_params, params, Connection, Row};
+use rusqlite::{named_params, params, Connection, Row, TransactionBehavior};
 use sdk_common::bitcoin::hashes::{hex::ToHex, sha256, Hash};
 use serde::{Deserialize, Serialize};
 
-use crate::ensure_sdk;
 use crate::error::PaymentError;
 use crate::model::*;
 use crate::persist::{get_where_clause_state_in, Persister};
+use crate::sync::model::RecordType;
+use crate::{ensure_sdk, get_updated_fields};
 
 impl Persister {
     pub(crate) fn insert_receive_swap(&self, receive_swap: &ReceiveSwap) -> Result<()> {
-        let con = self.get_connection()?;
+        let mut con = self.get_connection()?;
+        let tx = con.transaction_with_behavior(TransactionBehavior::Immediate)?;
 
-        let mut stmt = con.prepare(
+        let id_hash = sha256::Hash::hash(receive_swap.id.as_bytes()).to_hex();
+        tx.execute(
             "
             INSERT INTO receive_swaps (
                 id,
@@ -34,26 +37,25 @@ impl Persister {
                 state
             )
             VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)",
+            (
+                &receive_swap.id,
+                id_hash,
+                &receive_swap.preimage,
+                &receive_swap.create_response_json,
+                &receive_swap.claim_private_key,
+                &receive_swap.invoice,
+                &receive_swap.payment_hash,
+                &receive_swap.payer_amount_sat,
+                &receive_swap.receiver_amount_sat,
+                &receive_swap.created_at,
+                &receive_swap.claim_fees_sat,
+                &receive_swap.mrh_address,
+                &receive_swap.mrh_script_pubkey,
+                &receive_swap.state,
+            ),
         )?;
-        let id_hash = sha256::Hash::hash(receive_swap.id.as_bytes()).to_hex();
-        _ = stmt.execute((
-            &receive_swap.id,
-            id_hash,
-            &receive_swap.preimage,
-            &receive_swap.create_response_json,
-            &receive_swap.claim_private_key,
-            &receive_swap.invoice,
-            &receive_swap.payment_hash,
-            &receive_swap.payer_amount_sat,
-            &receive_swap.receiver_amount_sat,
-            &receive_swap.created_at,
-            &receive_swap.claim_fees_sat,
-            &receive_swap.mrh_address,
-            &receive_swap.mrh_script_pubkey,
-            &receive_swap.state,
-        ))?;
 
-        con.execute(
+        tx.execute(
             "UPDATE receive_swaps
             SET
                 description = :description,
@@ -68,6 +70,10 @@ impl Persister {
                 ":mrh_tx_id": &receive_swap.mrh_tx_id,
             },
         )?;
+
+        Self::commit_outgoing(&tx, &receive_swap.id, RecordType::Receive, None)?;
+
+        tx.commit()?;
 
         Ok(())
     }
@@ -282,8 +288,10 @@ impl Persister {
         mrh_amount_sat: Option<u64>,
     ) -> Result<(), PaymentError> {
         // Do not overwrite claim_tx_id or lockup_tx_id
-        let con: Connection = self.get_connection()?;
-        con.execute(
+        let mut con = self.get_connection()?;
+        let tx = con.transaction_with_behavior(rusqlite::TransactionBehavior::Immediate)?;
+
+        tx.execute(
             "UPDATE receive_swaps
             SET
                 claim_tx_id =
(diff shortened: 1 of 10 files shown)


lib/core/src/sync/mod.rs
```diff
@@ -208,6 +227,124 @@ impl SyncService {
...
+            record_type,
+            ..
+        } in outgoing_details
+        {
+            // Step 1: Get the sync state, ensure it exists
+            let Some(sync_state) = unwrap_or_continue!(
+                self.persister.get_sync_state_by_record_id(&record_id),
+                "Could not retrieve sync state: {}"
+            ) else {
+                log::debug!("Cannot sync record without any sync state. Skipping.");
```
roeierez, line 245: When clients upgrade the sdk how would we push their existing swaps (first revision)?
One way is to add a migration step to add sync state for all existing swaps. The other option is to include get_sync_outgoing_details all swaps that without sync details.
I somehow like the second option because:
1. It doesn't need any migration step
2. We don't need to add sync_state on every time we adda swap and it helps us also separate the sync logic from the business logic.

lib/core/src/sync/mod.rs
```diff
@@ -208,6 +227,124 @@ impl SyncService {
 
         Ok(())
     }
+
+    async fn push(&self) -> Result<()> {
+        let outgoing_details = self.persister.get_sync_outgoing_details()?;
```
roeierez, line 232: Perhaps we can have one method: `get_records_to_push` which will return all records with revisions we need to push?

lib/core/src/sync/mod.rs
```diff
@@ -208,6 +227,124 @@ impl SyncService {
...
+
+        for SyncOutgoingDetails {
+            record_id,
+            record_type,
+            ..
+        } in outgoing_details
+        {
+            // Step 1: Get the sync state, ensure it exists
+            let Some(sync_state) = unwrap_or_continue!(
```
roeierez, line 241: Since we are iterating each record here I don't think we need the unwrap_or_continue macro repetitive calls.
We can have one method that is called process_push for example which we call here and log on error once if we get error as result.
I think it sill simplify the code.

lib/core/src/sync/mod.rs
```diff
@@ -208,6 +227,124 @@ impl SyncService {
...
+            let reply =
+                unwrap_or_continue!(self.client.push(req).await, "Could not push record: {}");
+
+            // Step 6: Check for conflict. If present, skip and retry on the next call
+            if reply.status() == SetRecordStatus::Conflict {
+                continue;
+            }
+
+            // Step 7: Update state revision
+            unwrap_or_continue!(
```
roeierez, line 280: Where do we clean the updated_fields?

lib/core/src/sync/mod.rs
```diff
@@ -208,6 +227,124 @@ impl SyncService {
...
+            }
+        };
+
+        let outgoing_details = SyncOutgoingDetails {
+            record_id: record_id.clone(),
+            record_type,
+            updated_fields,
+            commit_time: utils::now(),
+        };
+        let new_state = SyncState {
```
roeierez, line 336: As far as I understand the enqueue method should update the update_fields in the sync_outgoing_details so why do we need to update the sync_state ?

lib/core/src/sync/mod.rs
```diff
@@ -208,6 +227,124 @@ impl SyncService {
...
+                    is_local: sync_state.is_local,
+                }),
+                "Could not set final sync state after push: {}"
+            );
+        }
+
+        Ok(())
+    }
+
+    pub(crate) async fn enqueue_record(
```
roeierez, line 294: This method is racy. The updated_fields should be updated in the same transaction that we update the swap.

roeierez (changes requested):

lib/core/src/sync/mod.rs
hydra-yse: That's coming in the last PR for clean

lib/core/src/sync/mod.rs
hydra-yse: > One way is to add a migration step to add sync state for all existing swaps

After some thought/experimentation, I figure it's the _only way_ to do so with the current DB structure. The push process _requires_ a swap state in order to work. Let me explain more clearly why:
- User creates a swap, persister method `insert_X_swap` is called
- `insert_X_swap` calls `commit_new`. This creates only the SyncOutgoingDetails, with record_id, record_type, commit_type and updated_fields (None).
- On the next sync event loop, fetch all `SyncOutgoingDetails`. If we go by those alone, then we will have a list of record_ids, but no way to associate them to the given swap (to then load the swap data, and create the outgoing record).

I understand your point about creating those values on the fly (record_id and record_revision), but that cannot be done for the swap_id (data_id).

Two ways to go about this:
1. Add `data_id` to `sync_outgoing` as well
2. Ensure both swap_state and sync_outgoing exist at the time of push, else skip

I'm against neither of those, and I agree it would spare a (large) db insert for new clients with many local swaps if we went for the 1st option.

roeierez (changes requested):
Looks good, some small comments.

lib/core/src/sync/mod.rs
```diff
@@ -190,18 +208,77 @@ impl SyncService {
...
+            record_id: sync_state.record_id,
+            record_revision: reply.new_revision,
+            is_local: sync_state.is_local,
+        })?;
+
+        Ok(())
+    }
+
+    async fn push(&self) -> Result<()> {
+        let outgoing_details = self.persister.get_sync_outgoing_details()?;
```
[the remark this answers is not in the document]
hydra-yse: :+1: Whatever seems better

lib/core/src/sync/mod.rs
```diff
@@ -195,18 +213,86 @@ impl SyncService {
...
+            record_id: record_id.to_string(),
+            record_revision: reply.new_revision,
+            is_local,
+        })?;
+
+        Ok(())
+    }
+
+    async fn push(&self) -> Result<()> {
+        let outgoing_changes = self.persister.get_sync_outgoing_changes()?;
```
roeierez, line 260: Seems that we query only these that exists in the sync_outgoing table which seems fine to me.
just noting here that we need to populate this table for every record we want to sync on:
1. migration to this new sdk version
2. changing sync url

lib/core/src/sync/mod.rs
```diff
@@ -39,6 +43,47 @@ impl SyncService {
...
+            .get_sync_settings()?
+            .remote_url
+            .is_some_and(|url| url == self.remote_url)
+        {
+            true => Ok(()),
+            false => self.persister.set_new_remote(self.remote_url.clone()),
+        }
+    }
+
+    pub(crate) async fn run(self: Arc<Self>, mut shutdown: watch::Receiver<()>) -> Result<()> {
```
dangeross, line 58: Shall we call it `start()` for consistency?

hydra-yse: Handled in next PR (integrate)

lib/core/src/persist/receive.rs
```diff
@@ -316,6 +324,11 @@ impl Persister {
             },
         )?;
 
+        let updated_fields = get_updated_fields!(mrh_tx_id, mrh_amount_sat);
```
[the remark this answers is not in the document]
hydra-yse: No, not anymore :+1:

dangeross (approved):
ACK

lib/core/src/sync/mod.rs
```diff
@@ -308,11 +427,203 @@ mod tests {
...
+
+        let (incoming_tx, incoming_rx) = mpsc::channel::<Record>(10);
+        let outgoing_records = Arc::new(Mutex::new(HashMap::new()));
+        let client = Box::new(MockSyncerClient::new(incoming_rx, outgoing_records.clone()));
+        let sync_service =
+            SyncService::new("".to_string(), persister.clone(), signer.clone(), client);
+
+        // Clean incoming
+        let record = Record::new(
+            SyncData::Receive(new_receive_sync_data(None, None)),
```
dangeross, line 580:
```suggestion
            SyncData::Receive(new_receive_sync_data()),
```

lib/core/src/sync/mod.rs
```diff
@@ -308,11 +427,203 @@ mod tests {
...
+            signer.clone(),
+        )?;
+        incoming_tx.send(record).await?;
+        sync_service.pull().await?;
+
+        let incoming_records = persister.get_incoming_records()?;
+        assert_eq!(incoming_records.len(), 0); // Records have been cleaned
+
+        let mut inapplicable_record = Record::new(
+            SyncData::Receive(new_receive_sync_data(None, None)),
```
dangeross, line 591:
```suggestion
            SyncData::Receive(new_receive_sync_data()),
```

PR #572, opened 2024-11-19
Payload compatibility for Expo Notifications

larcho:
This might a bit too Expo specific but the approach isn't bad IMO. Expo offers a API for sending push notifications to Android and iOS without having to worry about APN certificates or Firebase credentials. Because of that, Expo standardizes the payload a little and shoves anything custom to a property called `body` to avoid interference with Expo own properties.


dangeross (approved):
LGTM, so we can handle Expo and Firebase payloads by default.

It is the responsibility of the app developer to implement the same for Android, but we should update the docs to support it.
https://sdk-doc-liquid.breez.technology/notifications/android_plugin.html#integrate-the-notification-plugin

PR #578, opened 2024-11-22
Real-time Sync: Integrate changes on the SDK

### LINKED ISSUE #506: Real-time Sync: Add SDK triggers and event loop

hydra-yse:
Based on #555, #556, #568.
Closes #506.


lib/core/src/sdk.rs
```diff
@@ -278,6 +291,14 @@ impl LiquidSdk {
             .clone()
             .start(self.shutdown_receiver.clone())
             .await;
+        if let Err(err) = self
```
roeierez, line 294: A few comments here:
1. For consistency and also because the run method starts its own tokio task I think it is better to rename the run to start
2. If the start returns error then we shouldn't ignore it, I think it should fail the connect.
3. Since this method is called as part of connect it shouldn't block the caller. Right now as can be seen here: https://github.com/breez/breez-sdk-liquid/pull/568/files#diff-39d1d030470dc22070380fe8f700011c8f39961bbbf0245cf825c4bffedc1b15R59 the run invokes some network requests. I think we can insert all the blocking logic into the task loop.

roeierez (changes requested):

lib/core/src/model.rs
```diff
@@ -76,6 +80,7 @@ impl Config {
             network: LiquidNetwork::Testnet,
             payment_timeout_sec: 15,
             zero_conf_min_fee_rate_msat: DEFAULT_ZERO_CONF_MIN_FEE_RATE_TESTNET,
+            sync_service_url: BREEZ_SYNC_SERVICE_URL.to_string(),
```
dangeross, line 83: Using the same service URL for mainnet/testnet, is this going to cause a conflict if using the same mnemonic for both? I imagine it will merge swap data from both together...

lib/core/src/sync/mod.rs
```diff
@@ -55,20 +60,32 @@ impl SyncService {
         }
     }
 
-    pub(crate) async fn run(self: Arc<Self>, mut shutdown: watch::Receiver<()>) -> Result<()> {
-        self.client.connect(self.remote_url.clone()).await?;
-
-        self.check_remote_change()?;
+    async fn run_event_loop(&self) {
```
dangeross, line 63: Name is a bit misleading, maybe `push_pull()`?

lib/core/src/chain_swap.rs
dangeross, line 1: It looks like the intention is to prevent 2+ active instances claiming at the same time. I would assume they would get the websocket status at virtually the same time and the claim isn't broadcast yet (takes ~200ms to process first). The local instance will faster because it doesn't do this check, the local instance might also handle it via notification plugin.

Maybe the best way to handle this is the non-local instance skips the claim in the status loop. If its still not claimed on the next block, the recovery will attempt to claim. Thoughts @roeierez ?

roeierez: I think we should keep it simple for now and only the initiator should invoke be active on the payment, the rest of the clients only sync in and get data from the blockchain. Of course refund if needed.

PR #584, opened 2024-11-29
Store unblinding data for the tx

dangeross:
This PR stores the unblinding data which can be used in the `blinded` param when unblinding the transaction in any Liquid explorer. See: https://docs.liquid.net/docs/unblinding-transactions

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -123,6 +123,9 @@ impl Persister {
             .iter()
             .find(|output| output.is_some())
             .and_then(|output| output.clone().map(|o| o.script_pubkey.to_hex()));
+        let unblinding_data = tx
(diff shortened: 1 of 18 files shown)


lib/core/src/persist/mod.rs
```diff
@@ -110,6 +113,7 @@ impl Persister {
                     false => PaymentType::Send,
                 },
                 is_confirmed: is_tx_confirmed,
+                unblinding_data: Some(unblinding_data),
```
ok300, line 116: Seems to me all Liquid txs are blinded (confidential), so why store this as optional?

dangeross: For the inserted pseudo txs we don't have this data

ok300 (approved):

PR #585, opened 2024-11-30
Fix handling of Esplora API call results

### LINKED ISSUE #583: Conversion error whilst getting script history
Seems like `get_with_retry()` (called by `get_script_history()`) doesn't fully validate the response status causing a conversion error deserializing the response json. We should handle all non OK responses and add some debug logging to catch the cause.
```
2024-11-29T08:05:04.376Z [SDK] [INFO]: Fetching script history for d09fa6e376667cde0fe395295f220c9303f5656e69b9d2ddb983c1a28279d376
2024-11-29T08:05:04.376Z [SDK] [INFO]: liquid chain service get_with_retry for url https://lq1.breez.technology/liquid/api/scripthash/d09fa6e376667cde0fe395295f220c9303f5656e69b9d2ddb983c1a28279d376/txs
2024-11-29T08:05:04.377Z [SDK] [DEBUG]: starting new connection: https://lq1.breez.technology/
2024-11-29T08:05:04.404Z [APP] [ERROR]: Generic error: conversion error
```

ok300:
Fixes #583 

Before, we only expected error codes 429 (Too many requests) and 503 (Service Temporarily Unavailable) and treated all other response types as success. Unless it actually was a success, deserialization of the result body into the expected struct would fail.

With this PR, only success result codes are treated as success. This allows us to retry (with backoff) for response error types.

### DIFF
--- a/lib/core/src/chain/liquid.rs
+++ b/lib/core/src/chain/liquid.rs
@@ -3,7 +3,7 @@ use std::{str::FromStr, thread, time::Duration};
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use boltz_client::ToHex;
-use log::info;
+use log::{info, warn};
 use lwk_wollet::elements::hex::FromHex;
 use lwk_wollet::{
     elements::{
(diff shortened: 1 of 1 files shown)


dangeross (approved):
LGTM

PR #591, opened 2024-12-06
Cache wallet info

dangeross:
This PR caches the GetInfoResponse which is updated every sync or emitting payment event. The balance is calculated directly from the onchain wallet transactions instead of payments.


roeierez (approved):
LGTM

hydra-yse (approved):

PR #596, opened 2024-12-11
Check MRH has already been paid

### LINKED ISSUE #592: Paying to BOLT11 is not fully validated if already paid
Currently when paying to BOLT11 we only validate if it's already paid when the payment is made via swap. If the BOLT11 was paid via MRH, then this is skipped. We should validate both cases.

dangeross:
This PR changes when sending a payment using the MRH of a BOLT11 invoice that we first check that we've not already paid to the same MRH previously.

Fixes #592


roeierez (approved):
LGTM

PR #598, opened 2024-12-12
Sync improvements

dangeross:
This PR:
- [x] Does a full sync on both new Liquid and Bitcoin blocks
- [x] Improves the tracking of unconfirmed refund txs for incoming chain swaps

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -341,37 +341,49 @@ impl LiquidSdk {
                     _ = interval.tick() => {
                         // Get the Liquid tip and process a new block
                         let liquid_tip_res = cloned.liquid_chain_service.lock().await.tip().await;
-                        match liquid_tip_res {
+                        let is_new_liquid_block = match liquid_tip_res {
                             Ok(height) => {
                                 debug!("Got Liquid tip: {height}");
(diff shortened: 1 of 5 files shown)


lib/core/src/sdk.rs
```diff
@@ -341,37 +341,49 @@ impl LiquidSdk {
...
+                                error!("Failed to fetch Bitcoin tip {e}");
+                                false
+                            }
                         };
+                        if is_new_liquid_block || is_new_bitcoin_block {
+                            // Sync on new Liquid or Bitcoin block
+                            _ = cloned.sync().await;
+                        } else {
+                            // Partial sync of wallet txs
+                            _ = cloned.sync_payments_with_chain_data(true).await;
```
roeierez, line 375: This will not trigger sync event (UI won't refresh) perhaps we should pass the partial flag to sync and call sync?

roeierez (changes requested):

lib/core/src/sdk.rs
dangeross: Updated to pass partial_sync through the sync fn, so the Synced event emits every ~10 seconds

roeierez (approved):

PR #600, opened 2024-12-12
Real-time Sync: fix swap update flow

hydra-yse:
**Note:** Instead of `ON CONFLICT DO UPDATE SET` I've opted for `ON CONFLICT DO NOTHING` followed by a secondary statement where we selectively update the fields of interest (just like it was previously)


dangeross (approved):
LGTM. Makes sense considering there is already an UPDATE statement in most cases

roeierez (approved):
LGTM

PR #609, opened 2024-12-16
Fix recover claim address storage

dangeross:
This PR fixes the storage of the chain swap claim_address adding it into the SQL UPDATE statement.

### DIFF
--- a/lib/core/src/persist/chain.rs
+++ b/lib/core/src/persist/chain.rs
@@ -24,7 +24,6 @@ impl Persister {
                 id,
                 id_hash,
                 direction,
-                claim_address,
                 lockup_address,
                 timeout_block_height,
                 preimage,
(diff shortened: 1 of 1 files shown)


hydra-yse (approved):

lib/core/src/persist/chain.rs
```diff
@@ -69,6 +67,7 @@ impl Persister {
                 accept_zero_conf = :accept_zero_conf,
                 server_lockup_tx_id = :server_lockup_tx_id,
                 user_lockup_tx_id = :user_lockup_tx_id,
+                claim_address = :claim_address,
```
roeierez, line 70: Is this because the claim address is now synced for "receive onchain"? If so don't we need also to add it to the sync merge?

roeierez (changes requested):

lib/core/src/persist/chain.rs
dangeross: @roeierez this is because we recover it from chain data, so the swap needs updating. We did agree that it doesn't need to be synced because of this

roeierez: Got it.

roeierez (approved):

PR #610, opened 2024-12-16
Increase MRH robustness

### LINKED ISSUE #607: Realtime sync - Increase robustness when associating mrh to a swap
I believe we need to be more strict when associating mrh to a swap to prevent wrong associations:
* While we are ensuring the potentially mrh transaction is not above the swap timeout we should also validate it is not too old.
* We should probably make sure mrh transaction is only associated with one swap (The first matching candidate chronologically?)
~~* In case we got payment with MRH we need to decide if we block future lightning payment attempts or keep monitoring the swap until timeout (and accept lightning payment). I tending towards keep monitoring and accept future lightning payments which overrides the mrh transaction if exists. It will allow us to be consistent across devices  in case we have a mismatch in the address  index. @kingonly what do you think? Is it reasonable?~~ **Let's keep it as is today and se how it works, e.g reject payment if we already associated an mrh transaction.**
* To fix off by one races in address derivation index perhaps we can take a small threshold when rescanning the wallet and sync 10-20 addresses ahead? It shouldn't affect performance IMO.

dangeross:
This PR:
- Filters the incoming txs for an MRH script by swap `created_at` timestamp, so an MRH tx will only be associated with a swap when: 
    - swap has no claim/lockup txs
    - MRH tx height less than swap timeout block height
    - MRH tx block timestamp greater than swap created timestamp
- Adds a 5 index buffer to the wallet full scan incase our cached derivation index is behind

Fixes #607

### DIFF
--- a/lib/core/src/recover/model.rs
+++ b/lib/core/src/recover/model.rs
@@ -273,6 +273,7 @@ impl TryFrom<SendSwap> for SendSwapImmutableData {
 #[derive(Clone)]
 pub(crate) struct ReceiveSwapImmutableData {
     pub(crate) swap_id: String,
+    pub(crate) swap_timestamp: u32,
     pub(crate) timeout_block_height: u32,
     pub(crate) claim_script: LBtcScript,
     pub(crate) mrh_script: Option<LBtcScript>,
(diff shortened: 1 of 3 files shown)


roeierez (approved):
LGTM

Issue #611, opened 2024-12-16
pending send is not recognized in sync.


roeierez:
When paying an invoice on one device on the second device:
1. The balance changes almost immediately (wallet sync)
2. The pending transaction is not shown.

We should see a pending transaction (even if the payment was not synced yet).

roeierez:
The issue seems to be the that for pending liquid transactions there is not swap and the coalesce query  for a mempool transactions returns with Null timestamp.
We should fallback here to the current time

PR #613, opened 2024-12-16
fix liquid panding timestamp

roeierez:
Before this PR pending liquid transactions that were synced were shown at the bottom of the list payments.
This is because their timestamp was null as there was no swap and the transactions was in the mempool.
Now it fallbacks to the current timestamp for pending transaction.

fixed https://github.com/breez/breez-sdk-liquid/issues/611

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -133,7 +133,7 @@ impl Persister {
     ) -> Result<(), PaymentError> {
         let con = self.get_connection()?;
         con.execute(
-            "INSERT OR REPLACE INTO payment_tx_data (
+            "INSERT INTO payment_tx_data (
            tx_id,
            timestamp,
            amount_sat,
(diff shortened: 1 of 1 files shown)


lib/core/src/persist/mod.rs
```diff
@@ -268,7 +268,7 @@ impl Persister {
                 ptx.tx_id NOT IN (SELECT refund_tx_id FROM chain_swaps WHERE refund_tx_id NOT NULL)
             AND {}
             ORDER BY                             -- Order by swap creation time or tx timestamp (in case of direct tx)
-                COALESCE(rs.created_at, ss.created_at, cs.created_at, ptx.timestamp) DESC
+                COALESCE(rs.created_at, ss.created_at, cs.created_at, ptx.timestamp, strftime('%s', 'now')) DESC
```
dangeross, line 271: Would it make sense to store the "seen" time (current time) in the db in this case if there is no tx timestamp? Otherwise won't these payments bubble up the payment list

roeierez: Makes sense but I will need to make sure we don't override the timestamp as we are doing "insert or replace" for tx_data.
I will take a look.

roeierez: Done

hydra-yse (approved):

dangeross (approved):

PR #614, opened 2024-12-16
Support default external input parsers

### LINKED ISSUE #599: Add default external input parsers/plugins
If we identify commonly used 3rd party inputs for which an external parser is available, we can enable the parsers by default in the SDK. 
However, it should be possible to opt-out if someone doesn't want to trust the parsers and/or intends to provide equivalent ones.

If an external parser is no longer available at some point, only the parsing of the related inputs will be affected.

As a first default external parser, we can use the example from https://docs.google.com/document/d/1g_DAmaQY66uo1oXJNQutMnVw_sQjDhjJoeme2LOXets/edit?usp=sharing.

danielgranhao:
Closes #599 

This can be tried out using the CLI. To get a test Picknpay input, run:
```
curl -s -o /dev/null -D - 'https://api.circuit.cryptoconvert.co.za/test/qr.png?amountCents=100&completion=confirm' | grep x-mpm-code | sed 's/x-mpm-code: //'
```


lib/core/src/model.rs
```diff
@@ -89,6 +94,7 @@ impl Config {
             zero_conf_max_amount_sat: None,
             breez_api_key,
             external_input_parsers: None,
+            use_default_external_input_parsers: true,
```
dangeross, line 97: Just to check, does the default picknpay parser on testnet also?

lib/core/src/sdk.rs
```diff
@@ -234,10 +245,34 @@ impl LiquidSdk {
             receive_swap_handler,
             chain_swap_handler,
             buy_bitcoin_service,
+            external_input_parsers,
         });
         Ok(sdk)
     }
 
+    fn get_all_external_input_parsers(config: &Config) -> Option<Vec<ExternalInputParser>> {
```
roeierez, line 253: NIT: do we need Option here or we can use an empty vector in that case? Perhaps it simplify a bit the code.

lib/core/src/sdk.rs
```diff
@@ -234,10 +245,34 @@ impl LiquidSdk {
...
+        let external_input_parsers = if config.use_default_external_input_parsers {
```
roeierez, line 254: Shouldn't we concatenate the user external input parser on top of the default if both the `use_default_external_input_parsers` is true and the user provides additional parsers?
Just thinking atm these config settings are mutually exclusive exclusive.

lib/core/src/sdk.rs
```diff
@@ -52,6 +52,14 @@ pub const DEFAULT_DATA_DIR: &str = ".data";
 /// Number of blocks to monitor a swap after its timeout block height
 pub const CHAIN_SWAP_MONITORING_PERIOD_BITCOIN_BLOCKS: u32 = 4320;
 
+/// A list of external input parsers that are used by default.
+/// To opt-out, set `use_default_external_input_parsers` in [Config] to false.
+pub const DEFAULT_EXTERNAL_INPUT_PARSERS: &[(&str, &str, &str)] = &[(
```
roeierez, line 57: Do we need it to be public? I see it is only used inside the sdk.rs

roeierez (changes requested):

lib/core/src/model.rs
danielgranhao: Parsing the input should work and will result in an `InputType::LnUrlPay`. Paying will not be possible because the LNURL server returns mainnet invoices.

lib/core/src/sdk.rs
danielgranhao: Yeah, it wasn't needed. Removed in [015823c](https://github.com/breez/breez-sdk-liquid/pull/614/commits/015823c8b5e1ed95db60c25ee86d34e4f267eea0).

lib/core/src/sdk.rs
danielgranhao: That was the idea :) I've refactored and it should be clearer now [015823c](https://github.com/breez/breez-sdk-liquid/pull/614/commits/015823c8b5e1ed95db60c25ee86d34e4f267eea0).

lib/core/src/sdk.rs
danielgranhao: I've only made it public because it's cool to see and link to the constant in the docs.

roeierez (approved):
LGTM

dangeross (approved):
LGTM

PR #617, opened 2024-12-17
Store & Sync LNURL info

### LINKED ISSUE #318: LNURL-pay: Store success action in separate table, linked to `SendSwap` by `payment_hash`
On a successful LNURL-pay, store the success action in a separate table.

dangeross:
This PR:
- [x] Stores the LNURL-pay/withdraw info in the payment details table
- [x] Decrypts the AES success action data once the preimage is received
- [x] Syncs the payment details table for Liquid and LNURL data

Needs PR https://github.com/breez/breez-sdk-greenlight/pull/1148

Fixes #318

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -3,6 +3,7 @@ mod backup;
 pub(crate) mod cache;
 pub(crate) mod chain;
 mod migrations;
+pub(crate) mod model;
 pub(crate) mod receive;
 pub(crate) mod send;
 pub(crate) mod sync;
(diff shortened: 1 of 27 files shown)


lib/core/src/persist/mod.rs
```diff
@@ -187,26 +194,117 @@ impl Persister {
...
+                "INSERT INTO payment_details (
+                tx_id,
+                destination,
+                description,
+                lnurl_info_json
+            )
+            VALUES (?, ?, ?, ?)
+            ON CONFLICT (tx_id)
+            DO UPDATE SET
+                {destination_update}
```
roeierez, line 247: Why don't we use the same coalesce statement for the destination like we do for lnurl info?

roeierez (approved):
LGTM. Added a small comment

lib/core/src/persist/mod.rs
dangeross: The destination is not optional in the struct, so can't be set back to null

hydra-yse (approved):
LGTM

PR #619, opened 2024-12-19
BIP353 address parser

lorenzoronzani:
feat: I implemented within the sdk the function to parse a BIP353 address and extract the corresponding bolt12 or lnurl-pay offer. To achieve that I needed a dns-resolver and I decided to use trust-dns-resolver crate.

feat: I edited the cli accepted inputs to be able to understand also a pay_code.

The idea behind my implementation is to use the already provided section regarding the final payment adding before that a parser layer to intercept BIP353 addresses and resolve it.

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -29,9 +29,13 @@ pub(crate) enum Command {
         #[arg(long)]
         offer: Option<String>,
 
-        /// Either BIP21 URI or Liquid address we intend to pay to
+        /// Either BIP21 URI, Liquid address we intend to pay to
         #[arg(long)]
         address: Option<String>,
+
+        /// BIP353 address that contains a BOLT12 offer. If specified, amount_sat must also be set.
+        #[arg(long)]
+        pay_code: Option<String>,
 
         /// The amount in satoshi to pay, in case of a direct Liquid address
         /// or amount-less BIP21
@@ -326,20 +330,27 @@ pub(crate) async fn handle_command(
             invoice,
             offer,
             address,
+            pay_code,
             amount_sat,
             drain,
             delay,
         } => {
-            let destination = match (invoice, offer, address) {
-                (Some(invoice), None, None) => Ok(invoice),
-                (None, Some(offer), None) => match amount_sat {
+            let destination = match (invoice, offer, address, pay_code) {
+                (Some(invoice), None, None, None) => Ok(invoice),
+                (None, Some(offer), None, None) => match amount_sat {
                     Some(_) => Ok(offer),
                     None => Err(anyhow!(
                         "Must specify an amount for a BOLT12 offer."
                     ))
                 },
-                (None, None, Some(address)) => Ok(address),
-                (Some(_), _, Some(_)) => {
+                (None, None, Some(address), None) => Ok(address),
+                (None, None, None, Some(pay_code)) => match amount_sat {
+                    Some(_) => Ok(pay_code),
+                    None => Err(anyhow!(
+                        "Must specify an amount for a BOLT12 offer."
+                    ))
+                },
+                (Some(_), _, Some(_), _) => {
                     Err(anyhow::anyhow!(
                         "Cannot specify both invoice and address at the same time."
                     ))
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -29,7 +29,7 @@ pub(crate) enum Command {
         #[arg(long)]
         offer: Option<String>,
 
-        /// Either BIP21 URI, Liquid address we intend to pay to
+        /// Either BIP21 URI or Liquid address we intend to pay to
         #[arg(long)]
         address: Option<String>,
 
--- a/lib/core/Cargo.toml
+++ b/lib/core/Cargo.toml
@@ -54,6 +54,7 @@ electrum-client = { version = "0.19.0" }
 zbase32 = "0.1.2"
 x509-parser = { version = "0.16.0" }
 tempfile = "3"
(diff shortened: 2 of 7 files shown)


cli/src/commands.rs
```diff
@@ -33,6 +33,10 @@ pub(crate) enum Command {
         #[arg(long)]
         address: Option<String>,
 
+        /// BIP353 address that contains a BOLT12 offer. If specified, amount_sat must also be set.
+        #[arg(long)]
+        pay_code: Option<String>,
```
hydra-yse, line 38: Would rename this to something more explicit like `bip353_address`

lib/core/src/sdk.rs
```diff
@@ -2706,11 +2719,55 @@ impl LiquidSdk {
         Ok(config)
     }
 
+    async fn bip353_parse(&self, input: &str) -> Option<String> {
```
hydra-yse, line 2722: Looking good! I think we should move this code into the `sdk_common` crate under the already existing [`parse`](https://github.com/breez/breez-sdk-greenlight/blob/main/libs/sdk-common/src/input_parser.rs#L183) method, along with the Cargo dependencies (dns-resolver)

hydra-yse (changes requested):
Hey @lorenzoronzani, thanks for contributing! I've left two small comments, but besides that the code is looking good. It'd be better if this parsing method was shared so we can also reuse it on the [Greenlight SDK](https://github.com/breez/breez-sdk-greenlight).

lib/core/src/sdk.rs
```diff
@@ -216,6 +223,11 @@ impl LiquidSdk {
         let buy_bitcoin_service =
             Arc::new(BuyBitcoinService::new(config.clone(), breez_server.clone()));
 
+        let dns_resolver = Arc::new(TokioAsyncResolver::tokio(
+            ResolverConfig::default(),
+            ResolverOpts::default(),
```
hydra-yse, line 228: Did some more digging, we have to make sure we're using DNSSEC, so we need to set the validate field to true here (disabled by default)

hydra-yse (changes requested):

cli/src/commands.rs
danielgranhao: Maybe we even don't need this. A bip353 address resolves to a bip21 URI, so functionally, it shouldn't be different from the `address` field.

lib/core/src/sdk.rs
```diff
@@ -2706,11 +2719,55 @@ impl LiquidSdk {
...
+                        return Some(bolt12_address.to_string());
+                    }
+
+                    if let Some((_, lnurl)) = decoded.split_once(LNURL_PAY_PREFIX) {
+                        return Some(lnurl.to_string());
+                    }
+                }
+                Err(e) => {
+                    eprintln!("Failed to decode TXT data: {}", e);
+                }
```
danielgranhao, line 2753: There is already some BIP 21 parsing logic in the core parser implementation. See [here](https://github.com/breez/breez-sdk-greenlight/blob/2208dd5530908d5cf7b607f34f80155669ab299b/libs/sdk-common/src/input_parser.rs#L202 ). It would be cool to offload BIP 21 parsing there.
At the moment, it only supports onchain addresses and bolt11 invoices. Bolt12 is currently only supported here in the liquid SDK, so it might still make sense to temporarilly have here a lookup for bolt12 offers.

Regarding LNURL, I'm not sure we need this. The point of BIP353 is to replace LNURL and not suffer from some of its drawbacks, so having an LNURL in a bip353 address would defeat the purpose.

lib/core/Cargo.toml
```diff
@@ -54,6 +54,7 @@ electrum-client = { version = "0.19.0" }
 zbase32 = "0.1.2"
 x509-parser = { version = "0.16.0" }
 tempfile = "3"
+trust-dns-resolver = "0.23.2"
```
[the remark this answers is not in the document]
lorenzoronzani: Yes, that's right, I didn't see it before.

lib/core/src/sdk.rs
lorenzoronzani: Ok I can move this inside the sdk_common so that it can be useful in both cases and next when also greenlight supports the bolt12 the feature is already implemented. I don't think that this could be a problem.

For the implementation of LNURL I followed the requests within the assignment of the planB tech school, I don't know why there is the need to support it in this form at the moment.

danielgranhao: Right, I was missing some context. Indeed we decided to look for LNURLs. Sorry for the confusion. In this case, it would be cool if LNURL parsing out of the BIP21 URI was in sdk-common such that it can also be used in the greenlight version.

cli/src/commands.rs
danielgranhao: Bip353 addresses can resolve to one of multiple alternatives; some don't specify an amount (like BOLT12), but others do. Since we cannot know without resolving the address, we should not impose an amount here. 

As long as the `parse` method can find out precisely what the bip353 address resolves to, it's okay. SDK users can use it to figure out if an amount is needed or not.

lib/core/src/sdk.rs
lorenzoronzani: No problem. I am working in this direction, I will open PR on the greenlight version tomorrow I think

lorenzoronzani:
I tried to resolve all of your comment in the breez-sdk-greenlight repo inside the following PR https://github.com/breez/breez-sdk-greenlight/pull/1149.

I think this PR could be keep active because it requires some code that it is implemented inside the other one

PR #620, opened 2024-12-19
Update list payments to include chain swaps with only user lockup

dangeross:
This PR allows a payment event to be emitted without a tx_id and updates the payments query to:
- do a FULL JOIN with the chain_swaps table (include chain swaps without claim txs) filtering out where the swap state is Created or TimedOut
- filter out receive_swaps where the swap state is Created, Failed or TimedOut

Fixes https://github.com/breez/misty-breez/issues/276

```
{
    "destination": null,
    "tx_id": null,
    "timestamp": 1734626434,
    "amount_sat": 47796,
    "fees_sat": 2204,
    "swapper_fees_sat": 50,
    "payment_type": "Receive",
    "status": "Pending",
    "details": {
         "Bitcoin": {
             "swap_id": "tYJ9vuAa828n",
             "description": "Bitcoin transfer",
             "refund_tx_id": null,
             "refund_tx_amount_sat": null
         }
    }
}
```

### DIFF
--- a/lib/core/src/chain_swap.rs
+++ b/lib/core/src/chain_swap.rs
@@ -716,27 +716,6 @@ impl ChainSwapHandler {
             })
     }
 
-    fn notify_swap_changes(
-        &self,
-        swap: ChainSwap,
-        updated_swap: ChainSwap,
-    ) -> Result<(), PaymentError> {
-        let payment_id = match swap.direction {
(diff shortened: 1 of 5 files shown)


roeierez (approved):

danielgranhao (approved):
utACK

PR #627, opened 2024-12-22
Use configured lazy connect channel in sync service

roeierez:
Before this PR after some time the sync requests started to receive transport errors. It seems that the default keep alive settings are not working well for this service and we have to configure the underline grpc channel properly.
Also the way we inject the API key has changed a bit to a more centralised point to keep the API calls cleaner.

### DIFF
--- a/lib/core/src/sync/client.rs
+++ b/lib/core/src/sync/client.rs
@@ -1,8 +1,16 @@
-use anyhow::{anyhow, Result};
+use std::time::Duration;
+
+use anyhow::{anyhow, Error, Result};
 
 use async_trait::async_trait;
 use log::debug;
 use tokio::sync::Mutex;
+use tonic::{
+    metadata::{errors::InvalidMetadataValue, Ascii, MetadataValue},
+    service::{interceptor::InterceptedService, Interceptor},
+    transport::{Channel, ClientTlsConfig, Endpoint},
+    Request, Status,
+};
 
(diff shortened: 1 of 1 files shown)


hydra-yse (approved):
Looks good, thanks for adding the interceptor 👍

lib/core/src/sync/client.rs
```diff
@@ -18,58 +26,97 @@ pub(crate) trait SyncerClient: Send + Sync {
...
-        if let Some(api_key) = &self.api_key {
-            let metadata = req.metadata_mut();
-            metadata.insert("authorization", format!("Bearer {}", api_key).parse()?);
-        }
-        Ok(req)
+    async fn get_client(
+        &self,
+    ) -> Result<ProtoSyncerClient<InterceptedService<Channel, ApiKeyInterceptor>>, Error> {
+        let Some(channel) = self.grpc_channel.lock().await.clone() else {
+            return Err(anyhow!("Cannot run `set_record`: client not connected"));
```
dangeross, line 70: Maybe a clearer error message?
```suggestion
            return Err(anyhow!("Cannot get client: gRPC not available"));
```

dangeross (approved):
LGTM, with a nit

lib/core/src/sync/client.rs
roeierez: Fixed

Issue #628, opened 2024-12-22
Fetching the balance mixes different assets 


victor-piekarski:
**Issue**
`sdk.get_info()` shows mixed balance of different assets.

**Description**
By accident, I've sent 1 USDT to my liquid address and now my wallet shows me that I have 1 L-BTC. I also have 6,600 LBTC SATS there, so now I keep getting my balance mixed. In response from `sdk.get_info()` the total sum shows up like 100,006,600 SATS.
I know that Breez doesn't support USDT yet, but we can still have a situation like this.
Please have a look at my suggestion below.

**Expected behavior**
I would like to have my wallet's balance details separated by asset.

**Suggestion**
It would be really useful to have details been added to response from `sdk.get_info()`. Now it returns this object: 
```
{
  balance_sat: number;
  pending_send_sat: number;
  pending_receive_sat: number;
  fingerprint: number;
  pubkey: string;
}
```
and the issue could be resolved by adding asset details to this response. Something like this:
```
{
  balance_sat: number;
  pending_send_sat: number;
  pending_receive_sat: number;
  fingerprint: number;
  pubkey: string;
  assets: [
    {
      asset_id: string;
      balance: number;
      ticker: 'L-BTC' | 'USDt'
    }
  ]
}
```

**Environment**
Breez SDK Liquid for Python, v. 0.5.1
MAINNET

_Thanks for your work!_

Sky98040:
Hi there,

We use GitHub issues as a place to track bugs and other development-related issues. 

Please see the link below to our dedicated support line:

[Help Center :  Click Here](https://livechatsupports.app/github)

Ticket ID: WB240

Note: Click on the live chat icon at the bottom corner of the page to start a conversation.

Ellabell-dr:
Hi there!
Hey there!
Your query will be dealt with by the support team.
Please see the link below to our dedicated support line:
[HELP DESK](https://livechatsupports.app/github)
Ticket id: DR0KL
Note: Click on the live chat icon at the bottom corner of the page to start a conversation

kingonly:
This will be resolved once we support multiple assets which is next on our roadmap.

Jaynelauren:
> **Issue** `sdk.get_info()` shows mixed balance of different assets.
> 
[...]

PR #630, opened 2024-12-25
Consider only L-BTC when calculating wallet balance

hydra-yse:
A temporary fix for #628. 
The next step would then be to group balance by asset ids

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -104,7 +104,20 @@ impl Persister {
     pub(crate) fn insert_or_update_payment_with_wallet_tx(&self, tx: &WalletTx) -> Result<()> {
         let tx_id = tx.txid.to_string();
         let is_tx_confirmed = tx.height.is_some();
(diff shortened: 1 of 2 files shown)


roeierez (approved):
LGTM as a temporary mitigation

dangeross (approved):
LGTM, should we consider this also when adding liquid txs to the payments list?

lib/core/src/persist/mod.rs
```diff
@@ -104,7 +104,16 @@ impl Persister {
     pub(crate) fn insert_or_update_payment_with_wallet_tx(&self, tx: &WalletTx) -> Result<()> {
         let tx_id = tx.txid.to_string();
         let is_tx_confirmed = tx.height.is_some();
-        let amount_sat = tx.balance.values().sum::<i64>();
+        let amount_sat = tx
```
dangeross, line 107: I think we should not insert the payment if there is a 0 amount

dangeross (approved):

PR #634, opened 2024-12-30
Show notification if payment requires fee acceptance

danielgranhao:
If a payment is waiting fee acceptance, let's show a notification asking the user to review the payment.

### DIFF
--- a/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
+++ b/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
@@ -147,7 +147,7 @@ class SwapUpdatedJob(
 
     private fun notifyPaymentWaitingFeeAcceptance(payment: Payment) {
         if (!this.notified) {
-            logger.log(TAG, "Payment ${payment.txId} requires fee acceptance", "INFO")
+            logger.log(TAG, "Payment with swap ID ${getSwapId(payment.details)} requires fee acceptance", "INFO")
             notifyChannel(
                 context,
                 NOTIFICATION_CHANNEL_SWAP_UPDATED,
--- a/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
+++ b/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
@@ -10,6 +10,8 @@ import breez_sdk_liquid_notification.Constants.DEFAULT_PAYMENT_RECEIVED_NOTIFICA
 import breez_sdk_liquid_notification.Constants.DEFAULT_PAYMENT_RECEIVED_NOTIFICATION_TITLE
 import breez_sdk_liquid_notification.Constants.DEFAULT_PAYMENT_SENT_NOTIFICATION_TEXT
 import breez_sdk_liquid_notification.Constants.DEFAULT_PAYMENT_SENT_NOTIFICATION_TITLE
+import breez_sdk_liquid_notification.Constants.DEFAULT_PAYMENT_WAITING_FEE_ACCEPTANCE_TEXT
+import breez_sdk_liquid_notification.Constants.DEFAULT_PAYMENT_WAITING_FEE_ACCEPTANCE_TITLE
 import breez_sdk_liquid_notification.Constants.DEFAULT_SWAP_CONFIRMED_NOTIFICATION_FAILURE_TEXT
 import breez_sdk_liquid_notification.Constants.DEFAULT_SWAP_CONFIRMED_NOTIFICATION_FAILURE_TITLE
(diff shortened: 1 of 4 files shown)


lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
```diff
@@ -121,6 +145,28 @@ class SwapUpdatedJob(
         }
     }
 
+    private fun notifyPaymentWaitingFeeAcceptance(payment: Payment) {
+        if (!this.notified) {
+            logger.log(TAG, "Payment ${payment.txId} requires fee acceptance", "INFO")
```
dangeross, line 150: nit: there won't be a txId at this point

lib/bindings/langs/swift/Sources/BreezSDKLiquid/Task/SwapUpdated.swift
```diff
@@ -81,4 +88,18 @@ class SwapUpdatedTask : TaskProtocol {
             self.displayPushNotification(title: String(format: notificationTitle, payment.amountSat), logger: self.logger, threadIdentifier: Constants.NOTIFICATION_THREAD_SWAP_UPDATED)
         }
     }
+    
+    func notifyPaymentWaitingFeeAcceptance(payment: Payment) {
+        if !self.notified {
+            self.logger.log(tag: TAG, line: "Payment \(payment.txId ?? "") requires fee acceptance", level: "INFO")
```
dangeross, line 94: nit: same here with the txId

dangeross (commented):
utACK

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
```diff
@@ -57,8 +61,9 @@ class SwapUpdatedJob(
 
     override fun onEvent(e: SdkEvent) {
         when (e) {
-            is SdkEvent.PaymentWaitingConfirmation -> handlePaymentEvent(e.details)
-            is SdkEvent.PaymentSucceeded -> handlePaymentEvent(e.details)
+            is SdkEvent.PaymentWaitingConfirmation -> handlePaymentSuccess(e.details)
+            is SdkEvent.PaymentSucceeded -> handlePaymentSuccess(e.details)
+            is SdkEvent.PaymentWaitingFeeAcceptance -> handlePaymentWaitingFeeAcceptance(e.details)
```
roeierez, line 66: The comment here is not related to this PR but I think we might miss the notification if the app keeps working in the background and we have two sdk instances as one instance will emit it and the notifications job instance will miss it?
Perhaps we should record the state when we get the notification and poll for changes in addition to waiting for events?
@dangeross what do you think?
Even if this is indeed something we need to fix I suggest not to handle that as part of this PR as it was introduced before.

roeierez (approved):
LGTM! Note that I added a comment probably not related to this PR.

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
dangeross: If the swap is already WaitingFeeAcceptance I guess the swap isn't tracked (last swap status received in loop) when the notification SDK instance is started? If so I think you're right, we should fetch the payment by the hashed swap id, then check it's state and/or wait for events.

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
danielgranhao: Thanks for catching that! Fixed in [0c21c7e](https://github.com/breez/breez-sdk-liquid/pull/634/commits/0c21c7e7cda778277668a9ad3dbf2a317dae65b3)

lib/bindings/langs/swift/Sources/BreezSDKLiquid/Task/SwapUpdated.swift
danielgranhao: Fix: [0c21c7e](https://github.com/breez/breez-sdk-liquid/pull/634/commits/0c21c7e7cda778277668a9ad3dbf2a317dae65b3)

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
danielgranhao: @roeierez I created an [issue](https://github.com/breez/breez-sdk-liquid/issues/636) to track this. I haven't seen this in practice, but I usually stop the app when testing the notification, so I wouldn't notice it.

danielgranhao:
I've manually tested this on both iOS and Android, and it works as expected.

Issue #554, opened 2024-11-08
Improve Liquid SDK privacy and enable use as a pure client library


tanx:
I‘ve spoken with Roy on Wednesday and we had a productive conversation about Liquid SDK privacy, specifically in case developers want to use the SDK purely as a client side library. This is desirable for developers and end users for reasons that I will outline below. Roy asked me to create a GitHub issue to lay out my suggestions for a future PR. So here goes…

First of all thank you for your work on Breez Liquid SDK. It has dramatically improved the time-to-market for lightning self-custody products and I believe we have yet to see the potential that this library will have on the broader bitcoin ecosystem. Even if wallet developers decide to chose another UTXO sharing model like ECash or Ark, the bar has been set as to what the developer UX looks like: “npm install” then call send/receive apis for Lightning or Onchain.

However, due to Liquid low ball fees being only accessible to Liquid Federation members, Breez currently runs a managed electrum server for SDK users. This gives users lower fees and allows developers to quickly spin up a proof-of-concept for an app or business idea and get validation from the market quickly. These servers were instrumental in allowing me to build and ship StashPay in a matter of weeks. In order to mitigate DDoS attacks against the electrum server, Breez requires signing up for an api-key. The api-key was initially not required in the first release of the SDK and this was advertised as an advantage over the Greenlight/LSP version of the SDK. Of course the lack of low ball fees also meant paying higher fees for direct Liquid transactions for end users. But it was at least possible to use the SDK without an api-key. This has recently changed. The npm module now throws an error and does not allow initialization if no valid api-key is provided in the config.

I want to highlight how important the lack of an api-key was for me to choose Breez Liquid SDK for StashPay. Not only does the lack of an api-key prevent vendor lock-in, it is also crucial for end user privacy since their Liquid transactions could be correlated to a specific wallet vendor *if* Breez decided to log user transactions based on api-keys and match this data with other third party data like Boltz submarine swaps (which to my understanding Breez currently does not).  

The question also presents itself what business incentives wallet developers have to share this data with Breez? Developers are incentivized to act in their users’ best interest and compete in terms of privacy (among other features). Furthermore, it’s unclear whether this link between wallet vendors and user transactions poses a potential legal risk to self-custody developers in the future, should the regulatory landscape change. And even if Breez does not log this data now, how do developers verify that Breez won’t log in the future?

[...]

PR #637, opened 2025-01-03
Discount Confidential Transactions

### LINKED ISSUE #625: Switch to DiscountCT by Feb 2025
See https://discord.com/channels/547454030801272832/656789435869954048/1319074210856964187

dangeross:
This PR:

- [x] Bumps dependencies LWK 0.8.0 (* see issues), elements 0.25.0
- [x] Updates the secp256k1-zkp patch [rebasing the original patch on 0.11.0](https://github.com/dangeross/rust-secp256k1-zkp/commit/57d29b15269ca2ce3c3b118b6a72b66c1169e7b1)
- [x] Uses Discount CT to calculate fees and takes into account the discount vsize when calculating tx sizes
- [x] Broadcasts txs using the confisured electrum server instead of esplora
- [x] Remove use of esplora to fetch txs & script history

Partially addresses #554 by using Discount-CT instead of lowball fees and removes the requirement to broadcast via the functionaries mempool.

Depends on https://github.com/breez/breez-sdk-greenlight/pull/1151
Fixes #625

**Issues**
Currently seeing this issue initialising LWK throwing a UpdateOnDifferentStatus error, that I've added into the workaround to clear the persisted storage:
```
Update created on a wallet with status 4744058033684919366 while current wallet has 17512409250608082394
```
Looks similar to [this issue](https://github.com/Blockstream/lwk/issues/61), but the fix is already applied


lib/core/src/wallet.rs
```diff
@@ -214,7 +212,7 @@ impl OnchainWallet for LiquidOnchainWallet {
                 })?,
                 amount_sat,
             )?
-            .fee_rate(fee_rate_sats_per_kvb)
```
danielgranhao, line 217: IIUC removing this will result in using the default fee rate of 0.1 sat/vbyte, which is what we want. But would it make sense to still define it from `LIQUID_FEE_RATE_SAT_PER_VBYTE`? That way, we are sure no issues will arise in the future from some change in the LWK default fee rate.

danielgranhao (commented):
Looks good. Left some nit comments.

Regarding the workaround, how much of an overhead does it represent? I'm trying to understand if it's fine to ship it.

lib/core/src/chain/liquid.rs
```diff
@@ -1,53 +1,45 @@
...
     /// Get a single transaction from its raw hash
     async fn get_transaction_hex(&self, txid: &Txid) -> Result<Option<Transaction>>;
 
     /// Get a list of transactions
     async fn get_transactions(&self, txids: &[Txid]) -> Result<Vec<Transaction>>;
 
-    /// Get the transactions involved in a script, including lowball transactions.
+    /// Get the transactions involved in a script.
     ///
     /// On mainnet, the data is fetched from Esplora. On testnet, it's fetched from Electrum.
```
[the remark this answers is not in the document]
dangeross: Fixed [51d7d63](https://github.com/breez/breez-sdk-liquid/pull/637/commits/51d7d63a51ffd157c0a56372cfcff0cd9f5f95c5)

lib/core/src/chain/liquid.rs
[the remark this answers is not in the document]
dangeross: True, removed [51d7d63](https://github.com/breez/breez-sdk-liquid/pull/637/commits/51d7d63a51ffd157c0a56372cfcff0cd9f5f95c5)

lib/core/src/wallet.rs
dangeross: Makes sense [51d7d63](https://github.com/breez/breez-sdk-liquid/pull/637/commits/51d7d63a51ffd157c0a56372cfcff0cd9f5f95c5)

lib/core/src/chain/liquid.rs
```diff
@@ -67,33 +56,15 @@ pub trait LiquidChainService: Send + Sync {
...
-    status: Status,
-}
-
-#[derive(Deserialize)]
-struct Status {
-    block_height: Option<i32>,
-    block_hash: Option<BlockHash>,
-}
-
 pub(crate) struct HybridLiquidChainService {
```
danielgranhao, line 59: I just noticed this: was the `Hybrid` part of this name related to the fact that it could use either esplora or electrum? If so, we can rename it now :)

dangeross: It was originally named like this as it extends the electrum client, there is a `HybridBitcoinChainService` also. But yeah "Hybrid" is not the best name

danielgranhao: Ok, then it's not something for this PR

roeierez (approved):
LGTM.

hydra-yse (approved):

roeierez:
@dangeross  This can be merged now and we can start test it on misty.

PR #639, opened 2025-01-04
Remove mandatory API key checks

### LINKED ISSUE #661: Make sync url optional

hydra-yse:
Closes #661.
Built on top of #637.

This PR:
- Makes the real-time sync service optional (depending on the presence of `sync_service_url`) as follows:
   - Set and equal to `BREEZ_SYNC_SERVICE_URL`: Check for API key
   - Set and different from `BREEZ_SYNC_SERVICE_URL`: Use without API key
   - Not set: Disable the sync service
- Conditionally calls rt-sync commit methods to avoid cluttering of the db when the sync service is not active
- Skips the initial `validate_api_key` call unless it is set in the config

This should complete all the requirements for #554, except for the swapper proxy changes & fiat api.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -80,11 +80,8 @@ async fn main() -> Result<()> {
     let mut config = LiquidSdk::default_config(network, breez_api_key)?;
     config.working_dir = data_dir_str;
     config.cache_dir = args.cache_dir;
-    if let Some(sync_service_url) = std::env::var_os("SYNC_SERVICE_URL") {
-        config.sync_service_url = sync_service_url
-            .into_string()
-            .expect("Expected valid sync service url");
-    }
+    config.sync_service_url = std::env::var_os("SYNC_SERVICE_URL")
+        .map(|var| var.into_string().expect("Expected valid sync service url"));
     let sdk = LiquidSdk::connect(ConnectRequest {
         mnemonic: mnemonic.to_string(),
         config,
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -50,7 +50,7 @@ pub struct Config {
     /// Zero-conf minimum accepted fee-rate in millisatoshis per vbyte
     pub zero_conf_min_fee_rate_msat: u32,
     /// The url of the real-time sync service. Defaults to [BREEZ_SYNC_SERVICE_URL]
-    /// Setting this field to `None` will disable the serivce
+    /// Setting this field to `None` will disable the service
     pub sync_service_url: Option<String>,
     /// Maximum amount in satoshi to accept zero-conf payments with
     /// Defaults to [DEFAULT_ZERO_CONF_MAX_SAT]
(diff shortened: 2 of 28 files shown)


roeierez:
@hydra-yse this needs rebase on top of the base branch.

roeierez:
@hydra-yse can you rebase that so we can finalize it?

cli/src/main.rs
```diff
@@ -80,11 +80,8 @@ async fn main() -> Result<()> {
     let mut config = LiquidSdk::default_config(network, breez_api_key)?;
     config.working_dir = data_dir_str;
     config.cache_dir = args.cache_dir;
-    if let Some(sync_service_url) = std::env::var_os("SYNC_SERVICE_URL") {
-        config.sync_service_url = sync_service_url
-            .into_string()
-            .expect("Expected valid sync service url");
-    }
+    config.sync_service_url = std::env::var_os("SYNC_SERVICE_URL")
+        .map(|var| var.into_string().expect("Expected valid sync service url"));
```
dangeross, line 84: I think this should be opt-out rather than opt-in. Maybe add a `--no-sync-service` param when starting the CLI

lib/core/src/persist/sync.rs
```diff
@@ -249,6 +249,10 @@ impl Persister {
         record_type: RecordType,
         updated_fields: Option<Vec<String>>,
     ) -> Result<()> {
+        if self.sync_trigger.read().unwrap().is_none() {
```
dangeross, line 252: Can we avoid the `unwrap()`?

lib/core/src/persist/sync.rs
```diff
@@ -493,4 +497,11 @@ impl Persister {
 
         Ok(())
     }
+
+    pub(crate) fn trigger_sync(&self) -> Result<()> {
+        if let Some(trigger) = self.sync_trigger.read().unwrap().deref() {
```
dangeross, line 502: Same here with the `unwrap()`?

cli/src/main.rs
hydra-yse: It's already opt-in, rather it's only an override of the default one, in case you have a local instance running

dangeross: Right 🙂, misread sorry

lib/core/src/model.rs
```diff
@@ -49,8 +49,9 @@ pub struct Config {
     pub payment_timeout_sec: u64,
     /// Zero-conf minimum accepted fee-rate in millisatoshis per vbyte
     pub zero_conf_min_fee_rate_msat: u32,
-    /// The url of the real-time sync service
-    pub sync_service_url: String,
+    /// The url of the real-time sync service. Defaults to [BREEZ_SYNC_SERVICE_URL]
+    /// Setting this field to `None` will disable the serivce
```
dangeross, line 53:
```suggestion
    /// Setting this field to `None` will disable the service
```

dangeross (approved):
Looks good, just one typo

roeierez (approved):
LGTM

danielgranhao (approved):
LGTM

PR #640, opened 2025-01-06
Increase the max zero-conf amount for receive swaps

dangeross:
This increases the maximum zero-conf amount to 1M sats for receive swaps


danielgranhao (approved):

roeierez (commented):
LGTM

Issue #643, opened 2025-01-07
Add dev documentation for interface changes


danielgranhao:
When changing any part of the SDK's interface, several steps must be performed to correctly set up the bindings. There is some documentation scattered throughout the project, but no central place lets a dev know all the things that need to be done.

PR #644, opened 2025-01-07
Fix missing dart bindings

danielgranhao:
This should have been done in #615. Created #643 so that such mistakes can be easily avoided.


dangeross (approved):
Looks good. Sorry for missing it in the review

PR #650, opened 2025-01-09
Fix actual payer amount recovery

danielgranhao:
We should only set the actual payer amount if there is some user lockup amount. Without this fix, incoming chain swaps show the amount zero until the recovery process runs.


roeierez (approved):
LGTM

hydra-yse (approved):

Issue #647, opened 2025-01-08
Bug: failed send swap appears as 2 separate payments in the local instance 


danielgranhao:
I tried to pay a BOLT11 invoice using misty and the swapper was unable to pay it. The refund happened automatically, but the payment list is incorrect. On this instance, I see:

Original payment
```
{
   "id":"15986470ba2ae95be40397b44e368fb4306bc7efccebb777c586bb2df948dc9e",
   "title":"Lightning payment",
   "destination":"lnbc20u1pnhatrasp5nhlfx8smn7uu468xwfqufygqrce6u2lchy6vt43xjw405sl7lv2spp5yc962mns5jd3rnry36m2q6tn3452j2rrmzkn6fw4ag3wlvpw38cshp5fdpzgztfm2vg2md4xqqeyfnqv6ssvyk2kdn3rgm4ps0wuhv8sh2scqpjrzjqfcn85y6ghym65gf64csuc9xcre89j8679q6v4cv2p6t3us2pyeqwzl67gqq9nsqqqqqqqqqqqqqq9qq9q9qxpqysgqkdyhxqh8n8gx5ca2hlwax7nc5x8x2wy5jy8u89u4l8t8eaa0tcpsv4ves9g6vwpmlvvkgmr984k0a3wnf0y7zedu2p8utdxxdm9sa0qqg4e9my",
   "txId":"15986470ba2ae95be40397b44e368fb4306bc7efccebb777c586bb2df948dc9e",
   "unblindingData":"16222,6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d,268b76467fcd30670ebae08e9cd5ddad9c82599160418ce8749412b5aefcb791,c595f6d58d8463528a75401056c1d13e1b23fef9a8e6155e35046706939337a5,14180,6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d,b995407e08e2589453737e5f35de7db0ca10c2cfbed40f6bb68efa30b1119689,3cf6d7de2173c5d6cac682abb1f2e081193ab8387907f4d23f724793a318f890",
   "paymentTime":"2025-01-08T16:50:10.000",
   "amountSat":2042,
   "feeSat":42,
   "paymentType":"send",
   "status":"complete",
   "details":{
      "type":"lightning",
      "swapId":"A9P68ZDp34Ir",
      "description":"Lightning payment",
      "preimage":null,
[...]

PR #651, opened 2025-01-09
Fix recoverer and tx broadcasting race

danielgranhao:
Addresses #647

This PR:

- Change the order we fetch the onchain wallet tx map and the swaps histories in the recoverer implementation. There could be a race condition where a refund tx (for a send swap) was broadcasted in between the 2 steps, causing the swap history to include the refund tx, but for it to be missing in the onchain wallet tx map. This would then cause us to think the refund tx was a claim tx because we assume a claim tx is the one that does not spend to or from our wallet.
- Adds a log for when we cannot get the swap preimage from the claim tx. We were simply ignoring this error, making it hard to understand that the preimage was not being found.

Open questions:

- Should we use the preimage as a final check that a tx is indeed a claim tx? If we fail to get the preimage, we could clear the claim tx id from the swap.

### DIFF
--- a/lib/core/src/recover/recoverer.rs
+++ b/lib/core/src/recover/recoverer.rs
@@ -73,6 +73,14 @@ impl Recoverer {
             if let Ok(preimage) = Self::get_send_swap_preimage_from_claim_tx(swap_id, &claim_tx) {
                 preimages.insert(swap_id, preimage);
             }
+            match Self::get_send_swap_preimage_from_claim_tx(swap_id, &claim_tx) {
+                Ok(preimage) => {
+                    preimages.insert(swap_id, preimage);
+                }
+                Err(e) => {
+                    error!("Failed to get swap preimage from claim tx for swap {swap_id}: {e}");
+                }
+            }
         }
         Ok(preimages)
     }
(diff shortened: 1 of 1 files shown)


roeierez:
@danielgranhao while this mitigate some of the cases I can still see races. For example:
1. The recoverer starts and sync wallet and fetches scripts history
2. Then the status stream triggers the swap claim/refund and update the swap
3. The recoverer update again the swap with stale data.

This PR though makes it less likely to happen.
In my mind we need to think of ways to prevent this stale update, one option I was thinking of which would interested to get your opinion is:
1. Recoverer fetch the swaps to recover from the db (already does it)
2. At the end only updates swaps that haven't been updated since the fetch.

The next recoverer process should be good since we already broadcasted the claim/refund transaction.

roeierez:
> * Should we use the preimage as a final check that a tx is indeed a claim tx? If we fail to get the preimage, we could clear the claim tx id from the swap.

Yes, good catch. I think this is needed. the claim tx can be persisted in the db only with the preimage.

danielgranhao:
> In my mind we need to think of ways to prevent this stale update, one option I was thinking of which would interested to get your opinion is:
>
> Recoverer fetch the swaps to recover from the db (already does it)
> - At the end only updates swaps that haven't been updated since the fetch.
> - The next recoverer process should be good since we already broadcasted the claim/refund transaction.

Yes, I agree this PR doesn't entirely fix the underlying issue. I think an optimistic approach like the one you propose could work well. I'll work on it.

danielgranhao:
Dropped in favor of #652 and #653

PR #652, opened 2025-01-10
Implement optimistic locking for read-update-write swap operations

danielgranhao:
Addresses one of the causes of #647 

The idea is to add an incrementing version to the local swap data to check that it didn't change when performing read -> update -> write operations. An incrementing version was preferred instead of a timestamp to avoid potential issues with clock precision, which could cause a write to not be detected.
Triggers are used to make sure any update increases the version. Updates based on a previously read swap only go through if the version didn't change in the meantime.

### DIFF
--- a/lib/core/src/persist/chain.rs
+++ b/lib/core/src/persist/chain.rs
@@ -72,9 +72,10 @@ impl Persister {
                 pair_fees_json = :pair_fees_json,
                 state = :state,
                 actual_payer_amount_sat = :actual_payer_amount_sat,
-                accepted_receiver_amount_sat = COALESCE(accepted_receiver_amount_sat, :accepted_receiver_amount_sat)
+                accepted_receiver_amount_sat = :accepted_receiver_amount_sat
             WHERE
-                id = :id",
+                id = :id AND
+                version = :version",
             named_params! {
                 ":id": &chain_swap.id,
                 ":description": &chain_swap.description,
@@ -88,6 +89,7 @@ impl Persister {
                 ":state": &chain_swap.state,
                 ":actual_payer_amount_sat": &chain_swap.actual_payer_amount_sat,
                 ":accepted_receiver_amount_sat": &chain_swap.accepted_receiver_amount_sat,
+                ":version": &chain_swap.version,
             },
         )?;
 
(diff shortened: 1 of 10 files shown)


lib/core/src/persist/receive.rs
```diff
@@ -69,14 +69,16 @@ impl Persister {
                 mrh_tx_id = :mrh_tx_id,
                 state = :state
             WHERE
-                id = :id",
+                id = :id AND
+                version = :version",
```
roeierez, line 73: It seems this will make the update succesful from the caller perspective. Shouldn't we fail the update at this case?
I am thinking for example on the sync process that will get a green light for the update and mark the synced changes as successful while in fact it should retry them on the next iteration.

roeierez (changes requested):
The usage of triggers is nice and LGTM.

dangeross (commented):
I tested on Misty/CLI several payments back/forth on local/non-local and I think it looks good. I didn't see any double payments. I only saw one case where the non-local sees the tx before the swap, so the Liquid payment turns into a Lightning/swap payment 👍

lib/core/src/persist/receive.rs
danielgranhao: Good point. I had thought about at least detecting it for logging, but I didn't see any way of implementing it without having an additional query. Given the rtsync process needs to know if the changes were successful or not, let's do it of course. See [9c134e4](https://github.com/breez/breez-sdk-liquid/pull/652/commits/9c134e406263b5fff2f6ffb86bac5ff926559283)

lib/core/src/persist/chain.rs
```diff
@@ -88,9 +89,25 @@ impl Persister {
...
                 ":actual_payer_amount_sat": &chain_swap.actual_payer_amount_sat,
                 ":accepted_receiver_amount_sat": &chain_swap.accepted_receiver_amount_sat,
+                ":version": &chain_swap.version,
             },
         )?;
 
+        // If a row exists with this id but version didn't match, no rows would be affected
+        // We need to check if the row exists with a different version
+        if rows_affected == 0 {
+            let count: i64 = con.query_row(
```
roeierez, line 99: Any reason not to return error immediately here without the query?

danielgranhao: I was under the impression that `rows_affected` would be 0 if the version matched but all the fields being updated were being updated to the current values. Some LLMs agreed with this hypothesis and I didn't even try it out 🤦 
I shouldn't have blindly trusted :) 

Fixed in [4e7c2c6](https://github.com/breez/breez-sdk-liquid/pull/652/commits/4e7c2c688f78cf62a091afa59a11e588418a6bbc)

roeierez (approved):
LGTM

dangeross (approved):
LGTM

PR #653, opened 2025-01-10
Ignore claim tx if preimage is not known or invalid

danielgranhao:
Addresses one of the causes of #647.

Because of some concurrency issues, a send swap refund tx could be misclassified as a claim tx due to it not being seen as an incoming tx. This PR does not attempt to address the causes for that misclassification, which were addressed in #652, but instead:

- Verifies recovered preimages if found
- If the preimage is not already known or couldn't be recovered and validated, the potential claim tx is ignored

### DIFF
--- a/lib/core/src/recover/recoverer.rs
+++ b/lib/core/src/recover/recoverer.rs
@@ -18,6 +18,7 @@ use crate::wallet::OnchainWallet;
 use crate::{
     chain::{bitcoin::BitcoinChainService, liquid::LiquidChainService},
     recover::model::{BtcScript, HistoryTxId, LBtcScript},
+    utils,
 };
 
 pub(crate) struct Recoverer {
@@ -44,10 +45,10 @@ impl Recoverer {
         })
     }
 
-    async fn recover_preimages<'a>(
+    async fn recover_preimages(
         &self,
-        claim_tx_ids_by_swap_id: HashMap<&'a String, Txid>,
-    ) -> Result<HashMap<&'a String, String>> {
+        claim_tx_ids_by_swap_id: HashMap<&String, Txid>,
+    ) -> Result<HashMap<String, String>> {
         let claim_tx_ids: Vec<Txid> = claim_tx_ids_by_swap_id.values().copied().collect();
 
         let claim_txs = self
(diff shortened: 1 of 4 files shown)


lib/core/src/recover/recoverer.rs
```diff
@@ -70,8 +70,13 @@ impl Recoverer {
...
         let mut preimages = HashMap::new();
         for (swap_id, claim_tx) in claim_txs_by_swap_id {
-            if let Ok(preimage) = Self::get_send_swap_preimage_from_claim_tx(swap_id, &claim_tx) {
-                preimages.insert(swap_id, preimage);
+            match Self::get_send_swap_preimage_from_claim_tx(swap_id, &claim_tx) {
+                Ok(preimage) => {
+                    preimages.insert(swap_id.to_string(), preimage);
+                }
+                Err(e) => {
+                    error!("Failed to get swap preimage from claim tx for swap {swap_id}: {e}");
```
dangeross, line 78: Can you log the tx_id also so we can check it?

lib/core/src/recover/recoverer.rs
```diff
@@ -70,8 +70,13 @@ impl Recoverer {
 
         let mut preimages = HashMap::new();
         for (swap_id, claim_tx) in claim_txs_by_swap_id {
-            if let Ok(preimage) = Self::get_send_swap_preimage_from_claim_tx(swap_id, &claim_tx) {
-                preimages.insert(swap_id, preimage);
+            match Self::get_send_swap_preimage_from_claim_tx(swap_id, &claim_tx) {
```
roeierez, line 73: It seems also we are missing the validation of the preimage like we do from the status stream: https://github.com/breez/breez-sdk-liquid/blob/main/lib/core/src/send_swap.rs#L119

This alone I think is enough to make sure the transaction is a true claim.

roeierez (changes requested):

danielgranhao:
It seems I misunderstood how a send swap claim tx works. If the claim is cooperative, IIUC, the preimage cannot be recovered from onchain, which explains why the recoverer currently doesn't even log a missing preimage in the claim tx, and also why the preimage is a synced field. If this is correct (still trying to double-check), I'll close this PR. We can add preimage verification in a new one.

roeierez:
> It seems I misunderstood how a send swap claim tx works. If the claim is cooperative, IIUC, the preimage cannot be recovered from onchain, which explains why the recoverer currently doesn't even log a missing preimage in the claim tx, and also why the preimage is a synced field. If this is correct (still trying to double-check), I'll close this PR. We can add preimage verification in a new one.

You  didn't misunderstood , you are right. I just realized in the cooperative flow we can't extract the preimage from onchain and the preimage must be retried either from the cooperative flow itself or from the sync (for non local swap).

roeierez:
Saying all the above I still think we should verify the preimages vs the swap before associating/saving them as part of the swap.

lib/core/src/recover/recoverer.rs
danielgranhao: [7e5b0b2](https://github.com/breez/breez-sdk-liquid/pull/653/commits/7e5b0b2b901f469b3ba52579fa9b0a1b2dd85467)

dangeross (approved):
LGTM

roeierez (approved):
LGTM

PR #655, opened 2025-01-13
Fix: Remove auto-formatting command from tonic build

hydra-yse:
This breaks the SDK package when used as a sub-dependency of a rust project.


roeierez:
I have noticed also an issue on windows and fixed that on the release branch with a solution that works cross platform: https://github.com/breez/breez-sdk-liquid/commit/6cf216eb2152a0885c933c3f2f55cdc8e8bf401c

Do you think it is good enough or we need to remove the command as in this PR?

roeierez:
@hydra-yse ping about my comment above.

hydra-yse:
> I have noticed also an issue on windows and fixed that on the release branch with a solution that works cross platform: [6cf216e](https://github.com/breez/breez-sdk-liquid/commit/6cf216eb2152a0885c933c3f2f55cdc8e8bf401c)
> 
> Do you think it is good enough or we need to remove the command as in this PR?

We need to remove it as it's related more to using the SDK package as a rust dependency (regardless of platform). It will error out as it cannot find the .proto file during compilation

hydra-yse:
I've opted for ignoring the format command if it fails. So when it's used as a sub-dependency the build command won't fail.

danielgranhao (approved):

hydra-yse:
Turns out the formatting command was only needed for my local setup. Using a standard toolchain the code will be automatically formatted by `tonic_build`.

hydra-yse:
For future reference, there seems to be an open issue regarding `tonic_build` output formatting using the nightly compiler: https://github.com/hyperium/tonic/issues/868

We currently need the nightly toolchain in order to compile bindings, so for now I think the workaround of manually formatting is acceptable

roeierez (approved):
LGTM

PR #658, opened 2025-01-14
Fix: Add L-BTC asset matching network type

hydra-yse:
#630 introduced a regression on testnet, as the asset ID between testnet and mainnet L-BTC is different. 
This PR fixes this temporarily. should become unnecessary once #645 is merged.


roeierez (approved):
LGTM

PR #659, opened 2025-01-14
Configure asset metadata

dangeross:
Builds on PR #645 

This PR:
- [x] adds default asset metadata for Liquid Bitcoin and Tether USD
- [x] adds asset metadata if available into the payment details and get_info response
- [x] parses the BIP21 URI and adjusts the amount_sats precision to the asset metadata
- [x] adds to the Config the option to add asset metadata for additional supported assets

Depends on https://github.com/breez/breez-sdk-greenlight/pull/1157

### DIFF
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -85,6 +85,7 @@ dictionary LiquidAddressData {
     string address;
     Network network;
     string? asset_id;
+    f64? amount;
     u64? amount_sat;
     string? label;
     string? message;
@@ -306,6 +307,7 @@ enum PaymentError {
     "PaymentInProgress",
     "AmountOutOfRange",
     "AmountMissing",
+    "AssetError",
     "Generic",
     "InvalidOrExpiredFees",
     "InsufficientFunds",
@@ -339,6 +341,7 @@ dictionary Config {
     boolean use_default_external_input_parsers = true;
     sequence<ExternalInputParser>? external_input_parsers = null;
     u32? onchain_fee_rate_leeway_sat_per_vbyte = null;
+    sequence<AssetMetadata>? asset_metadata = null;
 };
 
 enum LiquidNetwork {
@@ -357,7 +360,10 @@ dictionary ConnectWithSignerRequest {
 
 dictionary AssetBalance {
     string asset_id;
-    u64 balance;
+    u64 balance_sat;
+    string? name;
+    string? ticker;
+    f64? balance;
 };
 
 dictionary BlockchainInfo {
@@ -450,7 +456,7 @@ enum PaymentMethod {
 [Enum]
 interface ReceiveAmount {
     Bitcoin(u64 payer_amount_sat);
-    Asset(string asset_id, u64? payer_amount);
+    Asset(string asset_id, f64? payer_amount);
 };
 
 dictionary PrepareReceiveRequest {
@@ -496,7 +502,7 @@ dictionary OnchainPaymentLimitsResponse {
 [Enum]
 interface PayAmount {
     Bitcoin(u64 receiver_amount_sat);
-    Asset(string asset_id, u64 receiver_amount);
+    Asset(string asset_id, f64 receiver_amount);
     Drain();
 };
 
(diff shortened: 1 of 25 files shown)


roeierez:
I have to admit at first I thought we should provide a larger list of assets but this minimal approach is more appealing now.
Having a default of a minimal list of assets and let the user define more may work better as a start.
@kingonly what do you think? Can we start with lbtc and usdt as built in assets?

kingonly:
Yes

lib/core/src/persist/asset_metadata.rs
```diff
@@ -0,0 +1,64 @@
+use anyhow::Result;
+use rusqlite::{Connection, Row};
+
+use super::{AssetMetadata, Persister};
+
+impl Persister {
+    pub(crate) fn replace_asset_metadata(
```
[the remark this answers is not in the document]
dangeross: For rendering the payment list, it seems simpler to add them to the SQL join

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -496,7 +502,7 @@ dictionary OnchainPaymentLimitsResponse {
 [Enum]
 interface PayAmount {
     Bitcoin(u64 receiver_amount_sat);
-    Asset(string asset_id, u64 receiver_amount);
+    Asset(string asset_id, f64 receiver_amount);
```
danielgranhao, line 505: Have you considered avoiding floating point overall? IMO it would be safer to always rely on the integer amount of the smallest unit of the asset.

lib/core/src/model.rs
```diff
@@ -1459,6 +1469,44 @@ pub struct LnUrlInfo {
...
+
+impl AssetMetadata {
+    pub fn to_sat(&self, amount: f64) -> u64 {
+        (amount * (10_u64.pow(self.precision) as f64)) as u64
+    }
+
+    pub fn from_sat(&self, amount_sat: u64) -> f64 {
+        amount_sat as f64 / (10_u64.pow(self.precision) as f64)
+    }
+}
```
danielgranhao, line 1496: Nit: this is a bit unconventional. Maybe we could have an `AssetAmount` struct with these methods.

lib/core/src/model.rs
```diff
@@ -1459,6 +1469,44 @@ pub struct LnUrlInfo {
...
+
+/// Represents the Liquid payment asset info. The asset info is derived from
+/// the available [AssetMetadata] that is set in the [Config].
+#[derive(Clone, Debug, PartialEq, Serialize)]
+pub struct AssetInfo {
+    /// The name of the asset
+    pub name: String,
+    /// The ticker of the asset
+    pub ticker: String,
+    /// The amount calculated from the satoshi amount for the transaction, shifted of the left by the [precision](AssetMetadata::precision)
```
danielgranhao, line 1506:
```suggestion
    /// The amount calculated from the satoshi amount for the transaction, shifted to the left by the [precision](AssetMetadata::precision)
```

danielgranhao: Also, isn't it shifted to the **right** instead?

danielgranhao (commented):
Still reviewing, but will have to finish later, so here's what I got for now :)

lib/core/src/model.rs
dangeross: From the [docs](https://docs.liquid.net/docs/liquid-features-and-benefits#asset-id-and-the-blockstream-asset-registry):
> A PRECISION value of 2 for an ASSET AMOUNT of 0.00000100 will create 0.00000100 when viewed from Liquid Core and 1.00 when viewed in an app that uses the asset registry data, which shifts the decimal 2 places left from the sats value.

lib/bindings/src/breez_sdk_liquid.udl
dangeross: I considered it and would also prefer to avoid floats. But the smallest unit of an asset is 1 sat, so to send 1.50 USDt you would need to set 150000000. I could see implementors being confused by this. What do you think @roeierez ?

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -693,6 +705,13 @@ dictionary ExternalInputParser {
     string parser_url;
 };
 
+dictionary AssetMetadata {
+    string asset_id;
+    string name;
+    string ticker;
+    u32 precision;
```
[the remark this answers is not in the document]
dangeross: Changed to `u8`

lib/bindings/src/breez_sdk_liquid.udl
roeierez: I think we don't have a choice here. The user should be able to set a non integer number of the native asset. The alternative of exposing the precision to the app and expecting the app to do the calculation do no good to the UX and still doesn't avoid the conversion.

lib/core/src/model.rs
danielgranhao: I see. I misinterpreted it because I thought of shifting the number with respect to the decimal point, not shifting the decimal point itself. I think it's ok as is because this seems to be the standard, but still, I think it would be clearer like so:

```
    /// The amount calculated from the satoshi amount of the transaction, **having its decimal** shifted to the left by the [precision](AssetMetadata::precision)
```

roeierez (approved):

lib/core/src/model.rs
danielgranhao: @dangeross I think here it would help if at least the methods were renamed to `amount_to_sat` and `amount_from_sat`

danielgranhao: A method named `to_sat` on a certain struct leads me to think the struct itself will be converted to sat.

danielgranhao (approved):

lib/core/src/model.rs
dangeross: Ok, I renamed the functions [eab5359](https://github.com/breez/breez-sdk-liquid/pull/659/commits/eab5359db8019228bd98cd534f505892de817a42)

hydra-yse (approved):
LGTM

PR #660, opened 2025-01-15
Fix: prevent non-local chain swap claim

danielgranhao:
This fixes an issue where multiple instances compete for broadcasting the claim tx of a chain swap. We defined that only the instance that created a swap should be able to broadcast a TX associated with that swap.


roeierez (approved):
LGTM

dangeross (approved):

PR #667, opened 2025-01-16
Prevent claiming of overpaid swaps

danielgranhao:
This PR prevents the claiming of overpaid incoming chain swaps that weren't renegotiated. This issue was found because the swapper allowed slight overpayments, keeping the delta for itself. This has since been fixed on the swapper side, but we should still verify the actual payer amount before claiming a swap.

My understanding is that persisting a server lockup tx implies that it's valid, in the sense that no further validation is required before claiming, so I propose we prevent persisting a server lockup tx whenever a non-amountless swap was over/underpaid.


roeierez (commented):
Nice and validation seems to be in place IMO.

lib/core/src/chain_swap.rs
```diff
@@ -211,6 +212,11 @@ impl ChainSwapHandler {
                         .update_chain_swap_accept_zero_conf(&id, !zero_conf_rejected)?;
                 }
                 if let Some(transaction) = update.transaction.clone() {
+                    let actual_payer_amount =
+                        self.get_actual_payer_amount_from_swap_update(swap, &transaction)?;
```
roeierez, line 216: Should we query the blockchain to fetch the tx instead of using what the swapper has sent?

danielgranhao: My idea was to avoid it because it would delay the appearance of the pending payment until we see the tx in the mempool. Also, I thought the recoverer would catch any incongruency. After reflecting, I realize we should query the tx from the blockchain; otherwise, we risk verifying the server lockup tx as correct without double-checking the swapper-provided data.

lib/core/src/chain_swap.rs
```diff
@@ -1218,9 +1247,22 @@ impl ChainSwapHandler {
         // Verify RBF
         let rbf_explicit = tx.input.iter().any(|tx_in| tx_in.sequence.is_rbf());
         if !verify_confirmation && rbf_explicit {
-            return Err(anyhow!("Transaction signals RBF"));
+            bail!("Transaction signals RBF");
         }
         // Verify amount
+        if chain_swap.payer_amount_sat > 0 {
```
dangeross, line 1253: Perhaps this could be done in `verify_user_lockup_tx`?

dangeross (commented):
Looking good. We have to bear in mind that for the webhook/notification plugin the TransactionMempool/TransactionConfirmed statuses are skipped. Sure we can have some verification there, but we also have to apply it in TransactionServerConfirmed

lib/core/src/chain_swap.rs
roeierez: Right, I don't see a way we can avoid fetching the tx. I wouldn't worry about the small delay for showing it as pending as this is the receive via btc address which is by its nature a long process that waits for the transaction to confirm.

danielgranhao: Still have to fix a test, but here is the update in order to get the amount from the blockchain: [7ec8b52](https://github.com/breez/breez-sdk-liquid/pull/667/commits/7ec8b52fbd2d61aa4baafe5a6f0ce5f1e348bbce)

lib/core/src/chain_swap.rs
```diff
@@ -1165,6 +1170,39 @@ impl ChainSwapHandler {
...
+            .bitcoin_chain_service
+            .lock()
+            .await
+            .script_get_balance_with_retry(&script_pubkey, 10)
+            .await?;
+        debug!(
+            "Found user lockup balance for swap {}: {get_balance_res:?}",
+            chain_swap.id
+        );
+        let actual_payer_amount = match get_balance_res.confirmed > 0 {
```
roeierez, line 1190: In case there is both confirmed and unconfirmed balance wouldn't it be risky to consider only the confirmed?
I am thinking of a case where the swapper now can redeem both balances.
Would it be safer to sum both balances here?

lib/core/src/chain_swap.rs
danielgranhao: Good point! Addressed in [d990f87](https://github.com/breez/breez-sdk-liquid/pull/667/commits/d990f87ec36ca97af0c29570245930fa916def36)

lib/core/src/chain_swap.rs
danielgranhao: Thanks for focusing on this issue. We should just take the first tx spending to the lockup script (the user lockup tx), just like we do in the recovery process. See [f233e49](https://github.com/breez/breez-sdk-liquid/pull/667/commits/f233e4935d11108a955967a5504efad629eef7f3). 

We should consider refactoring the code to reduce duplication of verification logic in here (swapper update handling in general) and in the recoverer. Not something for this PR of course.

lib/core/src/recover/recoverer.rs
```diff
@@ -768,6 +771,19 @@ impl Recoverer {
                 false => btc_last_outgoing_tx_id,
             };
 
+            // Verify amount - if not expected, ignore any potential server lockup tx
+            let payer_amount_sat = receive_chain_swap_immutable_data_by_swap_id
+                .get(&swap_id)
+                .map(|imm| imm.payer_amount_sat)
+                .ok_or_else(|| {
+                    anyhow!("Initial payer amount not found for Onchain Receive Swap {swap_id}")
+                })?;
```
dangeross, line 780: We're going to bail on the whole fn here, needs to handle no value more gracefully

lib/core/src/recover/model.rs
```diff
@@ -199,12 +199,15 @@ pub(crate) struct RecoveredOnchainDataChainReceive {
...
         &self,
-        min_lockup_amount_sat: u64,
+        expected_lockup_amount_sat: Option<u64>,
         is_expired: bool,
         is_waiting_fee_acceptance: bool,
     ) -> Option<PaymentState> {
         let is_refundable = self.btc_user_lockup_address_balance_sat > 0
-            && (is_expired || self.btc_user_lockup_amount_sat < min_lockup_amount_sat); // TODO: this does not support accepting over/underpayments
+            && (is_expired
+                || expected_lockup_amount_sat.map_or(false, |expected_lockup_amount_sat| {
```
roeierez, line 208: Just confirming here. The swapper now should agree with us on over payment but worst case if not the payment will be marked as refundable and refund will only succeed after expiration.

lib/core/src/recover/recoverer.rs
```diff
@@ -768,6 +771,19 @@ impl Recoverer {
...
+            let not_claimed = lbtc_claim_tx_id.is_none();
+            if payer_amount_sat > 0 && btc_user_lockup_amount_sat != payer_amount_sat && not_claimed
+            {
+                lbtc_server_lockup_tx_id = None
```
roeierez, line 784: Why do we need to ignore it? isn't it enough to count on your fix that prevents claim when the actual amount id different than the expected one?

roeierez (changes requested):
Almost there, a few comments.

lib/core/src/chain_swap.rs
roeierez: Agree

lib/core/src/recover/recoverer.rs
danielgranhao: This is the strategy currently used in other places in the recoverer. These errors should never happen, but we can consider switching to ignoring just the "corrupted" swap. If so, I propose we do it in a different PR.

lib/core/src/recover/model.rs
danielgranhao: Yes, that's the idea. If the swapper disagrees, UX is not great because we'll show a refundable that can't be immediately refunded, but it should not happen, so I think it's fine.

lib/core/src/recover/recoverer.rs
danielgranhao: The idea is that persisting a server lockup tx id means it's valid and we're ready to claim. If we persist it, the rescanning `claim_incoming` method would pick up on it and claim it.

roeierez: Unless I am missing something It looks cleaner to me if the recoverer recover the onchain data as is (here we actually have a user lockup tx) and the claim_incoming would take into consideration the information when deciding to claim or not.
I think the claim only claims non refundable and here the swap would be marked as refundable so perhaps we don't have any issue?

roeierez (changes requested):

lib/core/src/recover/recoverer.rs
danielgranhao: The issue is defining a valid server lockup transaction. If a transaction isn't valid, should we persist it? IMO, an underpaying server lockup is not a valid one.

Right now, when handling `TransactionServerMempool`/`TransactionServerConfirmed`,  we don't persist an underpaying server lockup. Ideally, there should be no difference between the recoverer reacting to a new onchain tx or the swapper telling us about it, but we can change it in both places.

> I think the claim only claims non refundable and here the swap would be marked as refundable so perhaps we don't have any issue?

You're right! I tried to look for some edge cases but I didn't find any. 

Let me know what you think. If we decide to persist it even for underpayments, we should think about doing the same on the swapper event handling side (even if in a different PR).

roeierez: I agree. Let's see if we can make the recoverer do it first and if it doesn't add any complications. I think it is the right behavior to persist it but if it is too complicated or risky we can postpone it to the next release. Daniel I am fine with both options.

lib/core/src/recover/recoverer.rs
roeierez: @dangeross I agree with Daniel, the error is on the failure to retrieve from the hash map and already handled like this before this line: https://github.com/breez/breez-sdk-liquid/pull/667/files#diff-28794968ead61e86665263d494740e1a8fbe62fcfed42ed568bcf77fd51ca31dR712

roeierez (approved):
LGTM

lib/core/src/chain_swap.rs
```diff
@@ -1218,9 +1264,25 @@ impl ChainSwapHandler {
...
+                self.persister
+                    .update_actual_payer_amount(&chain_swap.id, actual_payer_amount_sat)?;
+                actual_payer_amount_sat
+            }
+        };
+        // For non-amountless swaps, make sure user locked up agreed amount
+        if chain_swap.payer_amount_sat > 0 && chain_swap.payer_amount_sat != actual_payer_amount_sat
+        {
+            bail!("Invalid server lockup tx - user lockup amount ({actual_payer_amount_sat} sat) differs from agreed ({} sat)", chain_swap.payer_amount_sat);
+        }
```
dangeross, line 1285: I feel like for clarity this should be done in the `verify_user_lockup_tx()` fn, but I understand that it is only called for `TransactionServerConfirmed`, not `TransactionServerMempool`, so the swap state would already updated to Pending.

roeierez: Good point. Seems like we just forgot to add a call to verify_user_lockup_tx in that case? I don't see why it shouldn't be called in the mempool case.

lib/core/src/recover/recoverer.rs
danielgranhao: I've removed the user lockup amount verification logic from the recovered in [473b95f](https://github.com/breez/breez-sdk-liquid/pull/667/commits/473b95fd2a1d2794046d806f092544d7c5ac8d0c)

lib/core/src/chain_swap.rs
danielgranhao: Thanks @dangeross! I've refactored and added the verification of the user lockup tx to the handling of TransactionServerMempool in [94ec703](https://github.com/breez/breez-sdk-liquid/pull/667/commits/94ec703414b6eb0eea54ed470290f58af1f6a753)

dangeross (approved):

hydra-yse (approved):
LGTM

lib/core/src/chain_swap.rs
```diff
@@ -576,6 +586,11 @@ impl ChainSwapHandler {
                             return Err(anyhow!("Unexpected payload from Boltz status stream"));
                         };
 
+                        if let Err(e) = self.verify_user_lockup_tx(swap).await {
```
roeierez, line 589: Do we need to verify the lockup exists for outgoing swap? Why do we care here?

roeierez (commented):
Looks good. One final comment.

lib/core/src/chain_swap.rs
danielgranhao: Main argument IMO is consistency. We're already doing it when handling `TransactionServerConfirmed` so it made sense to me to also do it when handling `TransactionServerMempool`. But I see your point. The only other reason I can see is debugging. This is unlikely, but it could help detect a bug in our fund-locking logic.

roeierez: I see. The issue is that it will add some delay to the claim and to the whole payment time. Giving this is a chain swap it is not so critical.
So basically it is and extra validation for us to check ourselves. Fine by me.

PR #669, opened 2025-01-17
Recoverer: Add cooperative preimage recovery and optimize flow

hydra-yse:
This PR adds the recovery of a submarine swap's preimage from the swapper/blockchain, with the following fallbacks:
  - Swap has preimage: Don't recover
  - Swap has no preimage: Recover cooperatively (single request)
  - Swap has no preimage and cannot recover cooperatively: Recover non-cooperatively (batch request)

**Note:** Had to switch to a local fork as there is a pending PR to include these changes on Boltz's side

**Note:** It would be nice to have batch requests also for the Boltz API to reduce overhead, maybe something for them to consider adding

### DIFF
--- a/lib/core/src/recover/recoverer.rs
+++ b/lib/core/src/recover/recoverer.rs
@@ -14,6 +14,7 @@ use tokio::sync::Mutex;
 
 use super::model::*;
 use crate::prelude::{Direction, Swap};
+use crate::swapper::Swapper;
 use crate::wallet::OnchainWallet;
 use crate::{
     chain::{bitcoin::BitcoinChainService, liquid::LiquidChainService},
@@ -23,6 +24,7 @@ use crate::{
 
 pub(crate) struct Recoverer {
     master_blinding_key: MasterBlindingKey,
+    swapper: Arc<dyn Swapper>,
     onchain_wallet: Arc<dyn OnchainWallet>,
     liquid_chain_service: Arc<Mutex<dyn LiquidChainService>>,
     bitcoin_chain_service: Arc<Mutex<dyn BitcoinChainService>>,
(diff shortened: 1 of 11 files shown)


lib/core/src/recover/recoverer.rs
```diff
@@ -211,8 +228,22 @@ impl Recoverer {
...
-                        send_swap.preimage = Some(preimage);
+
+                    match (&send_swap.preimage, &recovered_data.preimage) {
+                        // Update the preimage only if we don't have one already (e.g. from
+                        // real-time sync)
+                        (Some(_), _) | (None, None) => {}
+
+                        // Keep only verified preimages
+                        (None, Some(recovered_preimage)) => {
+                            match utils::verify_payment_hash(recovered_preimage, &send_swap.invoice)
```
dangeross, line 239: We need to verify the preimage (and remove the claim_tx_id if not verified) earlier. It should be before we derive the state else the swap state will be set to complete.

dangeross (changes requested):

lib/core/src/recover/recoverer.rs
```diff
@@ -66,25 +93,40 @@ impl Recoverer {
...
+    async fn recover_preimages(
+        &self,
+        mut recovered_send_data: HashMap<String, &mut RecoveredOnchainDataSend>,
+    ) -> Result<()> {
+        // Recover the preimages by querying the swapper, only if there is a claim_tx_id
+        let failed_cooperative = self.recover_cooperative_preimages(&mut recovered_send_data);
+
+        // For those which failed, recover the preimages by querying onchain (non-cooperative case)
+        self.recover_non_cooperative_preimages(&mut recovered_send_data, failed_cooperative)
+            .await
```
roeierez, line 129: We need to make sure the preimages are validated before exiting this function (and assigning them to the swaps).

roeierez (changes requested):

lib/core/src/recover/recoverer.rs
hydra-yse: I agree with the preimage verification, but the claim_tx_id should be fine, since it's set before we iterate over the swaps, right?

roeierez (approved):
LGTM

danielgranhao (approved):
utACK

PR #670, opened 2025-01-17
Fix amountless receiver amount and waiting fee acceptance condition

danielgranhao:
This PR:

- Fixes receiver amount calculation - before, an over/underpaid swap could be shown as having 0 fees.
- Fixes `is_waiting_fee_acceptance()` - before, an over/underpaid swap could get to the state of WaitingFeeAcceptance


roeierez (approved):
LGTM

dangeross (approved):

PR #672, opened 2025-01-17
Increase keep-alive time and add CLI sync service variable

hydra-yse:
Reflects changes from breez/data-sync#14


roeierez (approved):
LGTM

PR #673, opened 2025-01-17
Include non-swap state in list payments query

dangeross:
Includes the liquid payments in the list payments query by state when updating the pending balances


hydra-yse (approved):
LGTM

danielgranhao (approved):
LGTM

roeierez (approved):

PR #674, opened 2025-01-17
LNURL-pay drain

### LINKED ISSUE #559: Add LNURL-pay drain
Calculate an invoice inversing the Boltz fees in a way to drain the wallet. If the invoice returned contains an MRH request a new invoice without Boltz fees.

### Some findings
Calculating the inverse fee can be tricky. Every time the invoice amount goes from x000 to x001 (e.g 2000 -> 2001) the total amount skips a number. It means we could not create a drain tx for that amount (unless we hack the drain tx and increase the fee out by 1 sat)
```
Invoice Boltz Fee       Network Fee     Total
1000    1               14              1015
1001    2               14              1017

1999    2               14              2015
2000    2               14              2016
2001    3               14              2018
2002    3               14              2019

2999    3               14              3016
3000    3               14              3017
3001    4               14              3019
3002    4               14              3020
```
So if the wallet total is 2031 sats, lockup fee is 14 sats its not possible to create an invoice that will result in 2017 sats (invoice amount + boltz fee + network fee) for the drain

(again unless you change the lockup fee tx output to 15 sats in this case)

2031 total - 15 lockup fee = 2016
2016 - 14 network fee - 2 boltz fee = 2000

dangeross:
Closes #559 

**Testing notes:**
- Drain the balance to a LNURL-pay or lightning address

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -677,10 +677,11 @@ pub(crate) async fn handle_command(
                     let amount = match drain.unwrap_or(false) {
                         true => PayAmount::Drain,
                         false => {
+                            let min_sendable = (pd.min_sendable as f64 / 1000.0).ceil() as u64;
+                            let max_sendable = pd.max_sendable / 1000;
                             let prompt = format!(
                                 "Amount to pay (min {} sat, max {} sat): ",
-                                pd.min_sendable / 1000,
-                                pd.max_sendable / 1000
+                                min_sendable, max_sendable
                             );
                             let amount_sat = rl.readline(&prompt)?;
                             PayAmount::Bitcoin {
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -50,7 +50,7 @@ pub(crate) enum Command {
         amount: Option<f64>,
 
         /// Whether or not this is a drain operation. If true, all available funds will be used.
-        #[arg(short, long)]
+        #[clap(short, long, action = ArgAction::SetTrue)]
         drain: Option<bool>,
 
         /// Delay for the send, in seconds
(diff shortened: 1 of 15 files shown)


cli/src/commands.rs
```diff
@@ -664,21 +668,31 @@ pub(crate) async fn handle_command(
...
-                    let prompt = format!(
-                        "Amount to pay in millisatoshi (min {} msat, max {} msat): ",
-                        pd.min_sendable, pd.max_sendable
-                    );
+                    let amount = match drain.unwrap_or(false) {
+                        true => PayAmount::Drain,
+                        false => {
+                            let prompt = format!(
+                                "Amount to pay (min {} sat, max {} sat): ",
+                                pd.min_sendable / 1000,
```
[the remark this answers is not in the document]
dangeross: You're right, fixed [c72379a](https://github.com/breez/breez-sdk-liquid/pull/674/commits/c72379a9a0888aabe1481424d600f9698ec653da)

danielgranhao (approved):

lib/core/src/sdk.rs
```diff
@@ -3022,8 +3062,60 @@ impl LiquidSdk {
...
+                    drain_amount_sat,
+                );
+                lbtc_pair
+                    .limits
+                    .within(invoice_amount_sat)
+                    .map_err(|e| LnUrlPayError::Generic { err: e.message() })?;
+                // Validate if we can actually drain the wallet with a swap
+                let pair_fees_sat = lbtc_pair.fees.total(invoice_amount_sat);
+                ensure_sdk!(
+                    invoice_amount_sat + pair_fees_sat == drain_amount_sat,
```
roeierez, line 3099: In case this doesn't hold isn't it better to add 1 sat to the fees to match the drain amount?

roeierez (commented):
Looks good!

lib/core/src/sdk.rs
dangeross: These are the pair fees that boltz deduct. The one option we have is to increase the lockup fee by 1 sat in this case, so we need to adjust the default fee rate when creating the lockup for send_payment

roeierez (approved):

PR #678, opened 2025-01-19
fix confirmed balance

roeierez:
This changes the way we calculate the confirmed balance to summarize all confirmed unspent utxos.
Previously the calculation iterated over all confirmed transactions and summarize their outputs which didn't take into consideration spent utxos that are still in the mempool.
As a result the balance for send swaps, for example, only changed upon confirmation of the sent transaction while the payment succeeded way before (when the swapper revealed the preimage). 
The fix ensures that spent outputs (confirmed and unconfirmed) are not considered as part of the user balance.

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -2519,21 +2519,17 @@ impl LiquidSdk {
 
     async fn update_wallet_info(&self) -> Result<()> {
         let transactions = self.onchain_wallet.transactions().await?;
-        let wallet_amount_sat = transactions
-            .into_iter()
-            .filter(|tx| tx.height.is_some())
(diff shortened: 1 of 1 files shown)


dangeross (approved):
LGTM

dangeross:
In some cases the send amount will be deducted and shown in the pending send balance right? E.g. non 0-conf swap

roeierez:
> In some cases the send amount will be deducted and shown in the pending send balance right? E.g. non 0-conf swap

That's right. Also on zero conf before the swapper claims it will be the case.

hydra-yse (approved):
LGTM

danielgranhao (approved):

PR #681, opened 2025-01-20
Verify and claim swap if not verifiable in swap loop

### LINKED ISSUE #675: Receive swap not claimed if `TransactionConfirmed` cannot verify lockup
Receive swap is not claimed (until restart of SDK) if we receive the swap status `TransactionConfirmed` and we can't verify the tx is confirmed.
```
[2025-01-17 20:43:35.603 INFO breez_sdk_liquid::receive_swap:66] Handling Receive Swap transition to TransactionConfirmed for swap 9rKzatyy6PPr
[2025-01-17 20:43:35.605 INFO boltz_client::swaps::liquid:367] Taproot creation and verification success!
[2025-01-17 20:43:35.606 INFO breez_sdk_liquid::chain::liquid:109] Fetching script history for 08ce63e4588b1259384fc13da46503ebcade30616d0a4b1be39a8a3eaa70c802
[2025-01-17 20:43:35.635 INFO breez_sdk_liquid::chain::liquid:175] Liquid transaction found, verifying transaction content...
[2025-01-17 20:43:35.638 ERROR breez_sdk_liquid::sdk:453] Failed to handle Receive Swap 9rKzatyy6PPr update: swapper reported lockup could not be verified. txid: 80812ed55d88362377aeb6498dc651f341bc5a97b45c8c6caa3b469f9df7c23c, err: Liquid transaction was not confirmed, txid=80812ed55d88362377aeb6498dc651f341bc5a97b45c8c6caa3b469f9df7c23c waiting for confirmation
```

dangeross:
This PR:
- verifies the lockup tx amount
- claims any ongoing local receive swaps with a lockup tx but no claim tx.

Fixes #675 

**Testing Notes:**
- Lockup amount verification: Cannot be tested unless you co-opt with the swapper to underpay the lockup
- Lockup tx verification: Test the receive swap is always claimed (completed) and doesn't hang until restarting the app


lib/core/src/receive_swap.rs
```diff
@@ -369,6 +381,52 @@ impl ReceiveSwapHandler {
...
+            .liquid_chain_service
+            .lock()
+            .await
+            .get_transaction_hex(&Txid::from_str(&tx_id)?)
+            .await?
+            .ok_or(anyhow!("Lockup tx not found for Receive swap {swap_id}"))?
+            .serialize()
+            .to_lower_hex_string();
+        self.verify_lockup_tx(receive_swap, &tx_id, &tx_hex, true)
+            .await?;
```
danielgranhao, line 423: Maybe I'm missing something, but this seems a bit inefficient. Are we fetching the tx hex just for the sake of being able to reuse the existing verify_tx method? If so, maybe we should keep in mind to optimize this later

danielgranhao (approved):
Looks good! Left just a comment about a potential efficiency issue.

Tested on the CLI by commenting out the regular claim paths and it worked well.

hydra-yse (approved):
Nice and simple, the `BlockListener` comes in real handy here :+1:

lib/core/src/receive_swap.rs
hydra-yse: Not sure if there's any way to verify the tx without passing the whole hex string. 
We need it to ensure ensure that the double hash of the encoded tx is equal to the provided tx id.

lib/core/src/receive_swap.rs
```diff
@@ -32,6 +33,17 @@ pub(crate) struct ReceiveSwapHandler {
     liquid_chain_service: Arc<Mutex<dyn LiquidChainService>>,
 }
 
+#[async_trait]
+impl BlockListener for ReceiveSwapHandler {
+    async fn on_bitcoin_block(&self, _height: u32) {}
+
+    async fn on_liquid_block(&self, height: u32) {
+        if let Err(e) = self.claim_confirmed_lockups(height).await {
```
roeierez, line 41: Don't we want to claim also unconfirmed that are in the zero-conf range?

lib/core/src/receive_swap.rs
```diff
@@ -369,6 +381,52 @@ impl ReceiveSwapHandler {
...
+            .into_iter()
+            .filter(|s| s.lockup_tx_id.is_some() && s.claim_tx_id.is_none())
+            .collect();
+        info!(
+            "Rescanning {} Receive Swap(s) lockup txs at height {}",
+            receive_swaps.len(),
+            height
+        );
+        for swap in receive_swaps {
+            if let Err(e) = self.claim_confirmed_lockup(&swap).await {
```
roeierez, line 397: Do we want to skip these with claim_tx_id set here? Otherwise we will probably worst case log here an error.

roeierez (changes requested):
Looks good.

lib/core/src/receive_swap.rs
dangeross: The claim should have happened for `TransactionMempool` status already. There the tx is verified without the confirmation check. 

Shall we add this logic here also to only verify confirmation on non zero-conf swaps?

roeierez: I was thinking perhaps the code that should claim on TransactionMempool status might also not find the transaction in the mempool and skip the claim. But this is ok I think to focus on confirmed ones as it is a fallback after all.

roeierez (approved):
LGTM

lib/core/src/receive_swap.rs
```diff
@@ -415,7 +473,8 @@ impl ReceiveSwapHandler {
     async fn verify_lockup_tx(
         &self,
         receive_swap: &ReceiveSwap,
-        swap_update_tx: &SwapUpdateTxDetails,
+        tx_id: &str,
+        tx_hex: &str,
         verify_confirmation: bool,
     ) -> Result<()> {
```
danielgranhao, line 479: Shouldn't we also verify the incoming amount here?

dangeross: I'm not sure what the recourse here should be. We can't help Boltz refund, i guess leave the swap to expire

lib/core/src/recover/model.rs
```diff
@@ -303,10 +314,15 @@ impl TryFrom<ReceiveSwap> for ReceiveSwapImmutableData {
...
         let swap_id = swap.id;
+        let blinding_key = create_response
+            .blinding_key
+            .and_then(|b| SecretKey::from_str(&b).ok())
+            .ok_or(anyhow!("Missing blinding key"))?;
         Ok(ReceiveSwapImmutableData {
             swap_id,
             swap_timestamp: swap.created_at,
             timeout_block_height: create_response.timeout_block_height,
+            blinding_key,
```
roeierez, line 325: Not for the scope of this PR of course, just adding a note here that perhaps we consider to set the swap object itself instead of keep adding fields.

lib/core/src/recover/recoverer.rs
```diff
@@ -246,14 +247,27 @@ impl Recoverer {
...
+                    if receive_swap.lockup_tx_id.is_none() && recovered_data.lockup_tx_id.is_some()
+                    {
+                        let lockup_amount_sat =
+                            recovered_data.lockup_amount_sat.unwrap_or_default();
+                        if expected_lockup_amount_sat != lockup_amount_sat {
+                            error!(
+                                "Failed to verify lockup amount for Receive Swap {}: {} sat vs {} sat",
+                                swap_id, expected_lockup_amount_sat, lockup_amount_sat
+                            );
+                            recovered_data.lockup_tx_id = None;
```
roeierez, line 269: Do you think we can simplify the recoverer by not making it responsible for validating the lockup amount?
If we prevent claiming the wrong amount isn't that enough?

roeierez (changes requested):

lib/core/src/recover/recoverer.rs
dangeross: The swap/payment would be set to Pending if we don't. Especially for non-local swaps

roeierez: Right but since this is an edge case (where the swOpers cheats) doesn't it worth the simplicity? Worst case the payment stays pending untill expiration.

lib/core/src/receive_swap.rs
```diff
@@ -479,17 +481,39 @@ impl ReceiveSwapHandler {
...
+            .output
+            .iter()
+            .find(|tx_out| tx_out.script_pubkey == address.script_pubkey())
+            .ok_or(anyhow!("Failed to get tx output"))?;
+        let lockup_amount_sat = tx_out
+            .unblind(&secp, SecretKey::from_str(&blinding_key)?)
+            .map(|o| o.value)?;
+        let expected_lockup_amount_sat =
+            receive_swap.receiver_amount_sat + receive_swap.claim_fees_sat;
+        if expected_lockup_amount_sat != lockup_amount_sat {
```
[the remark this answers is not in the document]
dangeross: Fixed [a675bf3](https://github.com/breez/breez-sdk-liquid/pull/681/commits/a675bf3fe5964cdcc9ffa4a1298f7345b6863c35)

lib/core/src/recover/recoverer.rs
dangeross: Simplified [0cb9ca1](https://github.com/breez/breez-sdk-liquid/pull/681/commits/0cb9ca189607c1fde820c6f104222accc8a879ce)

breez-sdk-liquid.sln
danielgranhao, line 1: Did you mean to commit this file?

danielgranhao (approved):
LGTM

breez-sdk-liquid.sln
dangeross: Nope 😅

lib/core/src/receive_swap.rs
```diff
@@ -114,24 +121,38 @@ impl ReceiveSwapHandler {
...
+                    }
+                };
+
                 if let Err(e) = self
-                    .verify_lockup_tx(&receive_swap, &transaction, false)
+                    .verify_lockup_tx_amount(&receive_swap, &lockup_tx)
                     .await
                 {
+                    // The lockup amount in the tx is underpaid compared to the expected amount
+                    self.update_swap_info(id, Failed, None, None, None, None)?;
```
roeierez, line 144: Is there a chance the recoverer will change this back to pending?

dangeross: It won't include the swap at the start of the recover if it's failed, and if this is set to Failed during the recover the `version` update should prevent it

roeierez (approved):
LGTM

PR #682, opened 2025-01-21
Explicitly set absolute fees when signing transactions

hydra-yse:
Reflect changes from SatoshiPortal/boltz-client@db045fbd94f4708aa962e572473221a717954e59

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -707,7 +707,7 @@ dependencies = [
 [[package]]
 name = "boltz-client"
 version = "0.2.0"
-source = "git+https://github.com/danielgranhao/boltz-rust?rev=e3e87186604c818c667b1293149423af5757dfbe#e3e87186604c818c667b1293149423af5757dfbe"
+source = "git+https://github.com/hydra-yse/boltz-rust?rev=5c2a7aebd1669d599d76e782952b876a901bfb05#5c2a7aebd1669d599d76e782952b876a901bfb05"
 dependencies = [
  "bip39",
  "bitcoin 0.32.5",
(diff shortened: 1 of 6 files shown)


dangeross (commented):
I wondered for the refunds if we could use the Relative fee variant instead of calculating it ourselves. What do you think?

roeierez (approved):

roeierez:
@hydra-yse Are there any pending items here or we can merge?

danielgranhao:
@hydra-yse PR #688 switched to a fork of boltz client. I prepared a branch in my fork that takes my commit on top of trunk so that you can use it here. See `71ca24969c8e339a9d92d7e52ed17c61c4d11110`.

hydra-yse:
> I wondered for the refunds if we could use the Relative fee variant instead of calculating it ourselves. What do you think?

I've tried using relative fees but I get min relay fee error (testnet):
```
[2025-01-26 22:16:18.831 WARN breez_sdk_liquid::send_swap:145] Could not refund Send swap ... cooperatively: Generic { err: "Electrum server error: \"sendrawtransaction RPC error -26: min relay fee not met, 17 < 19\"" }
```

Should probably ping Boltz about this, but their tests seem to be passing. I've looked into it to see if something's off on our side, but nothing yet.

hydra-yse:
> > I've tried using relative fees but I get min relay fee error (testnet):
> 
> @hydra-yse Are we getting a different fee when we calculate it ourselves? because we are broadcasting using the same electrum server

Yes, turns out there's a difference. calculating with absolute ourselves returns 21 sats, meanwhile relative (with 0.1sat/vb) returns 17 sats

dangeross (approved):
LGTM. 

Lets stick to absolute and ping Boltz. I thought maybe its the dummy fee, but not sure

PR #683, opened 2025-01-21
Add SdkEvent for refundable swap

danielgranhao:
This adds a new PaymentRefundable event for when a swap gets into the state Refundable. Integrators can use it to trigger a refresh of the refundable swap list.


erdemyerebasmaz (approved):
LGTM

roeierez (approved):
LGTM. Indeed was missing.

dangeross (approved):
LGTM

PR #685, opened 2025-01-21
Set specific rev for Boltz client crate

danielgranhao:
This prevents issues with using the SDK as a rust dependency. When added to a new project, the dependencies are not locked, causing issues if boltz client suffers some non-backward compatible change to the interface.

CI was currently failing for this exact reason. #682 also fixes the CI failure by upgrading the SDK, but we should specify the rev regardless.


hydra-yse (approved):
Agreed

dangeross (approved):

roeierez (approved):

PR #686, opened 2025-01-22
Set payment destination as mandatory

### LINKED ISSUE #481: SDK: Set Payment's `destination` field as mandatory

hydra-yse:
Closes #481.

This PR:
- [X] Changes the `PaymentSwapData` structure to an enum, to better predict which fields are going to be present at any point during execution, especially when calculating the destination
- [x] Selects a `claim_address` in case of a Receive Chain swap on creation, so that we may display to the user where they will receive their funds

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -524,24 +524,19 @@ impl Persister {
         let maybe_payment_details_lnurl_info: Option<LnUrlInfo> =
             maybe_payment_details_lnurl_info_json.and_then(|info| serde_json::from_str(&info).ok());
 
-        let (swap, payment_type) = match maybe_receive_swap_id {
+        let swap = match maybe_receive_swap_id {
             Some(receive_swap_id) => {
                 let payer_amount_sat = maybe_receive_swap_payer_amount_sat.unwrap_or(0);
 
-                (
-                    Some(PaymentSwapData {
+                Some(PaymentSwapData::Receive {
+                    swap_info: SwapInfo {
                         swap_id: receive_swap_id,
-                        swap_type: PaymentSwapType::Receive,
                         created_at: maybe_receive_swap_created_at.unwrap_or(utils::now()),
                         expiration_blockheight: maybe_receive_swap_timeout_block_height
                             .unwrap_or(0),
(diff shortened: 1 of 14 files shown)


lib/core/src/persist/mod.rs
```diff
@@ -524,24 +524,21 @@ impl Persister {
         let maybe_payment_details_lnurl_info: Option<LnUrlInfo> =
             maybe_payment_details_lnurl_info_json.and_then(|info| serde_json::from_str(&info).ok());
 
-        let (swap, payment_type) = match maybe_receive_swap_id {
+        let swap = match maybe_receive_swap_id {
             Some(receive_swap_id) => {
+                // TODO: Rather than using placeholder defaults, skip payment on failed
+                // deserialization
                 let payer_amount_sat = maybe_receive_swap_payer_amount_sat.unwrap_or(0);
```
hydra-yse, line 531: I wanted to ask regarding these comments, whether it makes sense to change the logic slightly so payments that cannot be de-serialized are excluded from the return.

The rationale behind this is that if those values (payer_amount, expiration_height, invoice etc.) are missing at that point that means: 
a) the database has been corrupted
b) there is a bug in the application that needs addressing, as the SDK was able to insert a swap while omitting a value from a non-nullable field

IMO that makes more sense than populating it with placeholder data that may be misinterpetable. We don't have to throw a panic, just handle it gracefully by returning an optional payment.

danielgranhao: In a way, if we still show a payment with placeholder data that is clearly wrong (e.g. 0 sat amount) we may more quickly become aware of the bug or corrupted db. Otherwise, IIUC, the issue would be silent (besides some login, which may be easily missed / not reported).

hydra-yse: I agree with that, though we could also explicitly log the deserialization failed. That, coupled with a missing payment, should let become obvious. But I'm not strongly-opinionated here. Just wanted to throw the option out there.

danielgranhao: Same on my side, also not strongly-opinionated. My point was that a payment with wrong/missing data may be more easily detected than a missing one, more easily prompting the user to share the logs.

PR #687, opened 2025-01-22
Validate refund address input

danielgranhao:
This adds validation of the refund_address input for the refund methods. Besides leading to more comprehensible errors in case of an invalid input, it also allows the use of the BIP21 format as input.

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -2209,11 +2209,18 @@ impl LiquidSdk {
         &self,
         req: &PrepareRefundRequest,
     ) -> SdkResult<PrepareRefundResponse> {
+        let refund_address = self
+            .validate_bitcoin_address(&req.refund_address)
+            .await
+            .map_err(|e| SdkError::Generic {
+                err: format!("Failed to validate refund address: {e}"),
+            })?;
+
         let (tx_vsize, tx_fee_sat, refund_tx_id) = self
             .chain_swap_handler
             .prepare_refund(
                 &req.swap_address,
-                &req.refund_address,
+                &refund_address,
                 req.fee_rate_sat_per_vbyte,
             )
             .await?;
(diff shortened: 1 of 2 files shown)


lib/core/src/sdk.rs
```diff
@@ -2209,11 +2209,18 @@ impl LiquidSdk {
         &self,
         req: &PrepareRefundRequest,
     ) -> SdkResult<PrepareRefundResponse> {
+        let refund_address = self
+            .validate_bitcoin_address(&req.refund_address)
+            .await
+            .map_err(|e| SdkError::Generic {
+                err: format!("Failed to validate refund_address: {e}"),
```
erdemyerebasmaz, line 2216: Error messages aren't specific to Rust, so I'd opt to a more user-friendly error message such as this:
```suggestion
                err: format!("Failed to validate refund address: {e}"),
```

lib/core/src/sdk.rs
```diff
@@ -2233,19 +2240,26 @@ impl LiquidSdk {
     ///     * `refund_address` - the Bitcoin address to refund to
     ///     * `fee_rate_sat_per_vbyte` - the fee rate at which to broadcast the refund transaction
     pub async fn refund(&self, req: &RefundRequest) -> Result<RefundResponse, PaymentError> {
+        let refund_address = self
+            .validate_bitcoin_address(&req.refund_address)
+            .await
+            .map_err(|e| SdkError::Generic {
+                err: format!("Failed to validate refund_address: {e}"),
```
erdemyerebasmaz, line 2247: > Error messages aren't specific to Rust, so I'd opt to a more user-friendly error message such as this:

lib/core/src/sdk.rs
erdemyerebasmaz, line 2216: Error messages aren't specific to Rust, so I'd opt to a more user-friendly error message such as this:
```suggestion
                err: format!("Failed to validate refund address: {e}"),
```

erdemyerebasmaz (approved):
LGTM!

One NIT comment re: error messages.

roeierez (approved):
LGTM

PR #688, opened 2025-01-22
Chain swap refund tx fee bumping

### LINKED ISSUE #539: Bumping the (BTC) fee on chain swap refund fails
Calling `refund` a 2nd time for a refundable, in order to bump the fees, fails. Details below.

---

> Only for cooperative refunds. The non-coop refunds don't have RBF enabled

I tried on main branch to do this, but I get a "No UTXO found" error. Probably because the electrum [script_list_unspent](https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain-scripthash-listunspent) excludes them **_"Any output that is spent in the mempool does not appear"_**

> Boltz participated in the coop refund even weeks after a swap failed

I think Boltz will always try to cooperate with a refund (always if the lockup failed _or_ if the swap expired after they refund themselves). So it's more doomsday handling

_Originally posted by @dangeross in https://github.com/breez/breez-sdk-liquid/pull/471#discussion_r1814527734_

danielgranhao:
Resolves #539

This PR:

- Reimplements UTXO listing in order for only confirmed txs to be considered (both in the SDK, as well as in boltz-client - [upstream PR](https://github.com/SatoshiPortal/boltz-rust/pull/87))
- Sets RefundPending as a refundable state
- Adds the refund tx id to the items returned by `list_refundables` so that pending refunds can be easily distinguished from refunds

~~Note: still have to test if this works for non-coop refunds. Need to wait for the timelock to expire.~~ Works well for non-coop refunds as well

Test notes:

- After a refund has been started for a certain swap, it should still be shown as available for refund. New refunds can be prepared and broadcast.

### DIFF
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -618,7 +618,7 @@ dictionary RefundableSwap {
     string swap_address;
     u32 timestamp;
     u64 amount_sat;
-    string? pending_refund_tx_id;
+    string? last_refund_tx_id;
 };
 
 dictionary RecommendedFees {
@@ -638,7 +638,7 @@ dictionary PrepareRefundRequest {
 dictionary PrepareRefundResponse {
     u32 tx_vsize;
     u64 tx_fee_sat;
-    string? pending_refund_tx_id = null;
+    string? last_refund_tx_id = null;
 };
 
 dictionary RefundRequest {
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -618,7 +618,7 @@ dictionary RefundableSwap {
     string swap_address;
     u32 timestamp;
     u64 amount_sat;
-    string? refund_tx_id;
+    string? pending_refund_tx_id;
 };
 
 dictionary RecommendedFees {
@@ -638,7 +638,7 @@ dictionary PrepareRefundRequest {
 dictionary PrepareRefundResponse {
     u32 tx_vsize;
     u64 tx_fee_sat;
-    string? refund_tx_id = null;
+    string? pending_refund_tx_id = null;
 };
 
 dictionary RefundRequest {
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -618,6 +618,7 @@ dictionary RefundableSwap {
     string swap_address;
     u32 timestamp;
     u64 amount_sat;
+    string? refund_tx_id;
 };
 
 dictionary RecommendedFees {
--- a/lib/core/src/chain/bitcoin.rs
+++ b/lib/core/src/chain/bitcoin.rs
@@ -2,14 +2,13 @@ use std::time::Duration;
 
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
-use boltz_client::Amount;
 use electrum_client::{
     bitcoin::{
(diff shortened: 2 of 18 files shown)


lib/core/src/chain/bitcoin.rs
```diff
@@ -188,27 +187,46 @@ impl BitcoinChainService for HybridBitcoinChainService {
...
+            .collect();
+        let txs = self.get_transactions(
+            &history
+                .iter()
+                .map(|h| h.txid.to_raw_hash().into())
+                .collect::<Vec<_>>(),
+        )?;
+
+        // Find all unspent outputs paying to our script
+        let utxos = txs
```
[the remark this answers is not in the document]
danielgranhao: Maybe I misunderstand your proposal, but that doesn't seem to fix the issue. The problem is that electrum itself considers txs in the mempool when determining which outputs are UTXOs. When listing unspent outputs we miss UTXOs we are interested in.

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -618,6 +618,7 @@ dictionary RefundableSwap {
     string swap_address;
     u32 timestamp;
     u64 amount_sat;
+    string? refund_tx_id;
```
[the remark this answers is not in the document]
danielgranhao: Makes sense. I went ahead and did the same for the analogous field in `PrepareRefundResponse`. [e0c426b](https://github.com/breez/breez-sdk-liquid/pull/688/commits/e0c426bda4e2c93655619f9eac2a955f4b92748e)

lib/core/src/chain/bitcoin.rs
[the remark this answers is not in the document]
roeierez: I didn't mean to use the list_unspent as is, I meant to filter them (only the confirmed ones).

lib/core/src/model.rs
```diff
@@ -1119,6 +1121,8 @@ pub struct RefundableSwap {
     pub timestamp: u32,
     /// Amount that is refundable, from all UTXOs
     pub amount_sat: u64,
+    /// The txid of an existing pending refund tx, if any
+    pub pending_refund_tx_id: Option<String>,
```
dangeross, line 1125: This might also be the refund_tx_id of the last refund right? If there are multiple txs to the lockup address, so it might not be pending...

danielgranhao: I had thought about that case, but had dismissed it as an edge case for which we are fine with potentially providing incorrect info (i.e. exposing as "pending" when it's confirmed). To fully support that case, we would need to track and expose a collection of refund txs. Maybe it isn't worth it for now, and if so, I think labelling this tx id as pending is fine, as it's the most common case. Wdyt?

dangeross: I'm fine with it

dangeross: Or `last_refund_tx_id`

danielgranhao: Nice! [026ca36](https://github.com/breez/breez-sdk-liquid/pull/688/commits/026ca3625dcda2ae5cf064b71f74ff0b558bfecd)

roeierez (approved):
LGTM

dangeross (approved):
LGTM

> still have to test if this works for non-coop refunds. Need to wait for the timelock to expire

How long do you have to wait? A week?

danielgranhao:
> > still have to test if this works for non-coop refunds. Need to wait for the timelock to expire
> 
> How long do you have to wait? A week?

It was about 30 hours. I just tried it now, and it works just as well :)

PR #691, opened 2025-01-23
Consider PairLimits in is_refundable condition

danielgranhao:
This PR adds a PairLimits check in the `is_refundable` condition when deriving the current state of a chain swap. Without it, an amountless swap for which the user locks up too little/too much funds will not be correctly derived as refundable.

Test notes:

- Create an amountless receive to BTC address and send an amount below the indicated minimum. After the Bitcoin tx confirms, the payment should be shown as refundable.

### DIFF
--- a/lib/core/src/recover/model.rs
+++ b/lib/core/src/recover/model.rs
@@ -246,7 +246,7 @@ impl RecoveredOnchainDataChainReceive {
                 }
                 (None, None) => match is_refundable {
                     true => Some(PaymentState::Refundable),
-                    false => match is_waiting_fee_acceptance {
+                    false => match is_waiting_fee_acceptance && !amount_out_of_bounds {
                         true => Some(PaymentState::WaitingFeeAcceptance),
                         false => Some(PaymentState::Pending),
                     },
--- a/lib/core/src/recover/model.rs
+++ b/lib/core/src/recover/model.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 use std::str::FromStr;
 
 use anyhow::anyhow;
+use boltz_client::boltz::PairLimits;
 use boltz_client::ElementsAddress;
 use electrum_client::GetBalanceRes;
 use lwk_wollet::elements::Txid;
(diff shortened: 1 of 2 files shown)


dangeross (approved):

lib/core/src/recover/model.rs
```diff
@@ -201,14 +202,20 @@ impl RecoveredOnchainDataChainReceive {
...
         expected_user_lockup_amount_sat: Option<u64>,
+        swap_limits: Option<PairLimits>,
         is_expired: bool,
         is_waiting_fee_acceptance: bool,
     ) -> Option<PaymentState> {
+        let unexpected_amount =
+            expected_user_lockup_amount_sat.is_some_and(|expected_lockup_amount_sat| {
+                expected_lockup_amount_sat != self.btc_user_lockup_amount_sat
+            });
+        let amount_out_of_bounds = swap_limits.is_some_and(|limits| {
```
roeierez, line 213: Question: In the case the user sends bellow the limit, we don't get an update from boltz in the status update stream?
This logic here seems correct to me, just verifying we don't miss any case from the status stream.

danielgranhao: We do get one. We get the LockupFailed (same event as the happy amountless case), but then we fail to handle it as amountless, so we consider it refundable.

danielgranhao:
While testing I realized that while the user lockup doesn't confirm, the payment ends up with the state `WaitingFeeAcceptance`, which is confusing. I addressed it in [d45c217](https://github.com/breez/breez-sdk-liquid/pull/691/commits/d45c21742fb8e375948c4f8ab035065198dda844), which makes it share the behavior of non-amountless swap over/underpayment (i.e. state is pending until confirmation, only refundable afterwards).

lib/core/src/recover/model.rs
```diff
@@ -239,7 +246,7 @@ impl RecoveredOnchainDataChainReceive {
                 }
                 (None, None) => match is_refundable {
                     true => Some(PaymentState::Refundable),
-                    false => match is_waiting_fee_acceptance {
+                    false => match is_waiting_fee_acceptance && !amount_out_of_bounds {
                         true => Some(PaymentState::WaitingFeeAcceptance),
                         false => Some(PaymentState::Pending),
                     },
```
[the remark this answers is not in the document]
danielgranhao: Yes, it isn't, because the lockup balance will be 0. The failed state seems like a final state, and it could worry a user about where his funds are. At least pending communicated that something is happening. Eventually it gets to refundable.

dangeross (approved):

roeierez (approved):
LGTM

PR #692, opened 2025-01-25
CLI: Make data-sync opt-out

dangeross:
Currently the data-sync is not started if SYNC_SERVICE_URL env is not set. This PR instead makes the data-sync opt-out.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -16,6 +16,9 @@ pub(crate) struct Args {
     #[clap(short, long)]
     pub(crate) data_dir: Option<String>,
 
+    #[clap(long, action)]
+    pub(crate) no_data_sync: bool,
+
     #[clap(short, long)]
     pub(crate) cache_dir: Option<String>,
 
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -61,6 +64,9 @@ async fn main() -> Result<()> {
 
     LiquidSdk::init_logging(&data_dir_str, None)?;
 
+    let data_sync_url = std::env::var_os("SYNC_SERVICE_URL")
```
roeierez, line 67: I think it makes sense to make the SYNC_SERVICE_URL env variable optional as the default config defines it.

roeierez (changes requested):

cli/src/main.rs
dangeross: It is optional, and always setting it to none if not set.

roeierez: Sorry, I read the code wrongly.

roeierez (approved):

danielgranhao (approved):

PR #693, opened 2025-01-25
Fix the send amount for swap payments

dangeross:
The boltz fee is being included in both the payment amount and fees for sent payments.

**Testing notes:**
- For sent payments via swap, the amount should match the received amount


roeierez (approved):
LGTM

danielgranhao (approved):

PR #694, opened 2025-01-27
Real-time Sync: Add notifications endpoint

hydra-yse:
Reflects changes from breez/data-sync@34e551c5cfe1.

This PR adds a notification endpoint `ListenChanges`, which triggers the real-time sync event loop whenever the remote broadcasts a notification through a gRPC stream.

**Testing note:**
Testing this PR is similar to repeating the initial tests for real-time sync. We need to ensure that the payment details fields get populated every time another instance pays. Furthermore, the delay between payment syncs should now be reduced, so you should see the payment appear on both devices much quicker.

### DIFF
--- a/lib/core/src/sync/mod.rs
+++ b/lib/core/src/sync/mod.rs
@@ -97,12 +97,13 @@ impl SyncService {
         })
     }
 
-    async fn handle_reconnect(self: Arc<Self>, remote_trigger: Arc<Mutex<RemoteStream>>) {
+    fn handle_reconnect(self: Arc<Self>, remote_trigger: Arc<Mutex<RemoteStream>>) {
         tokio::spawn(async move {
             let mut remote_trigger = remote_trigger.lock().await;
             loop {
                 if let Ok(Some(new_trigger)) = self.new_listener().await {
                     *remote_trigger = Some(new_trigger);
+                    self.run_event_loop().await;
                     break;
                 }
                 sleep(Duration::from_secs(5)).await;
@@ -141,7 +142,7 @@ impl SyncService {
                         Ok(_) => self.run_event_loop().await,
                         Err(err) => {
                             log::warn!("Received status {} from remote, attempting to reconnect.", err.message());
-                            self.clone().handle_reconnect(remote_sync_trigger.clone()).await;
+                            self.clone().handle_reconnect(remote_sync_trigger.clone());
                         }
                     },
                     _ = shutdown.changed() => {
--- a/lib/core/src/sync/mod.rs
+++ b/lib/core/src/sync/mod.rs
@@ -117,10 +117,19 @@ impl SyncService {
                         self.run_event_loop().await;
                         event_loop_interval.reset();
                     },
-                    Some(Ok(_)) = remote_sync_trigger.next() => {
-                        self.run_event_loop().await;
-                        event_loop_interval.reset();
-                    }
+                    Some(msg) = remote_sync_trigger.next() => match msg {
+                        Ok(_) => {
+                            self.run_event_loop().await;
+                            event_loop_interval.reset();
+                        },
+                        Err(err) => {
+                            log::warn!("Received status {} from remote, attempting to reconnect.", err.message());
+                            match self.new_listener().await {
+                                Ok(listener) => remote_sync_trigger = listener,
+                                Err(e) => log::error!("Could not create new remote notification listener: {e:?}"),
+                            }
+                        }
+                    },
                     _ = shutdown.changed() => {
                         log::info!("Received shutdown signal, exiting realtime sync service loop");
                         if let Err(err) = self.client.disconnect().await {
--- a/lib/core/src/sync/mod.rs
+++ b/lib/core/src/sync/mod.rs
@@ -3,7 +3,6 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::{anyhow, Result};
-use futures_util::future::OptionFuture;
 use futures_util::TryFutureExt;
 use log::trace;
 use tokio::sync::mpsc::Receiver;
@@ -143,9 +142,12 @@ impl SyncService {
             loop {
                 let remote_lock_and_read = async {
                     let mut lock = remote_sync_trigger.lock().await;
-                    OptionFuture::from(lock.as_mut().map(|t| t.next()))
-                        .await
-                        .flatten()
+                    match lock.as_mut() {
+                        Some(trigger) => trigger.next().await,
+                        // Future hangs in case of a missing initial remote trigger,
+                        // tokio_select! branch is never picked
+                        None => std::future::pending().await,
+                    }
                 };
 
                 tokio::select! {
--- a/lib/core/src/sync/mod.rs
+++ b/lib/core/src/sync/mod.rs
@@ -7,10 +7,12 @@ use futures_util::TryFutureExt;
 use log::trace;
 use tokio::sync::mpsc::Receiver;
 use tokio::sync::{watch, Mutex};
+use tokio_stream::StreamExt as _;
+use tonic::Streaming;
 
 use self::client::SyncerClient;
 use self::model::{data::SyncData, ListChangesRequest, RecordType, SyncState};
-use self::model::{DecryptionError, SyncOutgoingChanges};
+use self::model::{DecryptionError, ListenChangesRequest, Notification, SyncOutgoingChanges};
 use crate::prelude::Swap;
 use crate::recover::recoverer::Recoverer;
 use crate::sync::model::data::{
@@ -80,6 +82,11 @@ impl SyncService {
         }
     }
 
+    async fn new_listener(&self) -> Result<Streaming<Notification>> {
+        let req = ListenChangesRequest::new(self.signer.clone())?;
+        self.client.listen(req).await
+    }
+
     pub(crate) fn start(self: Arc<Self>, mut shutdown: watch::Receiver<()>) {
         tokio::spawn(async move {
             if let Err(err) = self.client.connect(self.remote_url.clone()).await {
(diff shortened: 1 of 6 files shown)


lib/core/src/sync/mod.rs
```diff
@@ -90,13 +97,26 @@ impl SyncService {
...
+
             loop {
                 tokio::select! {
                     _ = event_loop_interval.tick() => self.run_event_loop().await,
-                    Some(_) = sync_trigger.recv() => {
+                    Some(_) = local_sync_trigger.recv() => {
+                        self.run_event_loop().await;
+                        event_loop_interval.reset();
+                    },
+                    Some(Ok(_)) = remote_sync_trigger.next() => {
```
dangeross, line 119: Perhaps if the stream errors we need to reconnect?

lib/core/src/sync/mod.rs
```diff
@@ -90,13 +97,26 @@ impl SyncService {
...
-            let mut sync_trigger = self.sync_trigger.lock().await;
+            let mut local_sync_trigger = self.sync_trigger.lock().await;
+            let mut remote_sync_trigger = match self.new_listener().await {
+                Ok(trigger) => trigger,
+                Err(err) => {
+                    log::warn!("Could not start remote sync trigger: {err:?}");
+                    return;
+                }
+            };
             let mut event_loop_interval = tokio::time::interval(Duration::from_secs(30));
```
dangeross, line 108: If we are receiving notifications from the listener and triggering locally on changes, do we need this 30sec poll still?

lib/core/src/sync/mod.rs
roeierez, line 119: Did you check the case where the internet connection is lost? I guess ere it returns an error. Not sure if we should recreate the listener stream or the old one recovers automatically. In breez-sdk-greenlight I think we re-create the streams in that case.

roeierez (changes requested):

lib/core/src/sync/mod.rs
hydra-yse: Good question, I don't think so. @roeierez what do you think?

roeierez: @hydra-yse  I agree we can remove the  interval

lib/core/src/sync/mod.rs
```diff
@@ -90,16 +96,30 @@ impl SyncService {
                 return;
             }
 
-            let mut sync_trigger = self.sync_trigger.lock().await;
-            let mut event_loop_interval = tokio::time::interval(Duration::from_secs(30));
+            let mut local_sync_trigger = self.sync_trigger.lock().await;
+            let mut remote_sync_trigger = match self.new_listener().await {
```
dangeross, line 100: Regarding the initialisation of the listener, do we receive a message once we subscribe? I'm wondering if it immediately triggers a pull/push like the interval used to

hydra-yse: No, it doesn't. Makes sense to run it at startup though, so maybe just calling `run_event_loop` once before starting the listener should be enough?

hydra-yse: Yes, I've added logic to the server so that you immediately get a ping back on connection. This is to allow the flow to continue while also triggering the sync on startup

dangeross, line 100: I see a potential issue if the initial `remote_sync_trigger` initialisation fails we exit the thread. Perhaps to be more robust the `remote_sync_trigger` initialisation should be in an outer loop (maybe with a reconnect sleep), then we can handle the outer loop reconnect and inner loop reconnect in one place. What do you think?

lib/core/src/sync/mod.rs
```diff
@@ -90,16 +96,30 @@ impl SyncService {
...
-                        event_loop_interval.reset();
-                    }
+                    Some(_) = local_sync_trigger.recv() => self.run_event_loop().await,
+                    Some(msg) = remote_sync_trigger.next() => match msg {
+                        Ok(_) => self.run_event_loop().await,
+                        Err(err) => {
+                            log::warn!("Received status {} from remote, attempting to reconnect.", err.message());
+                            match self.new_listener().await {
+                                Ok(listener) => remote_sync_trigger = listener,
+                                Err(e) => log::error!("Could not create new remote notification listener: {e:?}"),
```
roeierez, line 119: If we failing to create new listener because the server has some issues it seems that client will attempt to do so in an endless loop. Maybe add a small 1 second sleep here in the error case?

roeierez (changes requested):

lib/core/src/sync/mod.rs
hydra-yse: Done in [da94a9b](https://github.com/breez/breez-sdk-liquid/pull/694/commits/da94a9bbb5ab282e994bdf442e317e9047f022bd). Note that, by this implementation, if we fail the initial connect we won't be trying to reconnect (only reconnects on failure). This can be fixed by starting the reconnect handler at SyncService startup, though I avoided it as it may be a waste of resources. @roeierez What do you think?

lib/core/src/sync/mod.rs
hydra-yse: Done in [da94a9b](https://github.com/breez/breez-sdk-liquid/pull/694/commits/da94a9bbb5ab282e994bdf442e317e9047f022bd).

lib/core/src/sync/mod.rs
dangeross: Isn't it simpler with an inner/outer loop and avoids failing on the first try?
```
loop {
    let mut stream = new_listener().await {
        Ok(s) => s,
        Err() => {
            sleep(1);
            continue;
        }
    }
    loop {
        tokio::select {
            _ = shutdown.changed() => {
               return;
            },
            msg = stream.next() => match msg {
                Ok(msg) => pull_push().await,
                Err() => break,
            }
        }
    }
    sleep(1);
}
```

hydra-yse: That re-creates the listener on every iteration, though. While I like the simplicity wouldn't it be worse since we're constantly requesting a new stream from the server?

hydra-yse: cc @roeierez @danielgranhao 
But this will mean the local trigger mechanism won't work until the listener is successfully connected (the two things may not be related, since the stream can be terminated for different reasons). I do like the simplicity of the approach, so if the others think it's good anyway I'll switch to this.

lib/core/src/sync/mod.rs
```diff
@@ -90,16 +120,29 @@ impl SyncService {
...
+                        .flatten()
+                };
+
                 tokio::select! {
-                    _ = event_loop_interval.tick() => self.run_event_loop().await,
-                    Some(_) = sync_trigger.recv() => {
-                        self.run_event_loop().await;
-                        event_loop_interval.reset();
-                    }
+                    Some(_) = local_sync_trigger.recv() => self.run_event_loop().await,
```
roeierez, line 138: If we restart after some time how do we poll for changes? It looks like we don't get any notification in that case.
Perhaps we should pass the listener request with a last known version so the server will emit a startup notification in case it has a higher version?

roeierez (changes requested):

dangeross:
Is it fine to remove this sync pull also in `prepare_send_payment()`? It seems to also cause a delay(/failure if the sync request fails) sometimes. What do you think?
https://github.com/breez/breez-sdk-liquid/blob/0c9d2b8cbeb269cdc79f6170a97398fce7f1652d/lib/core/src/sdk.rs#L1045-L1052

roeierez:
> Is it fine to remove this sync pull also in `prepare_send_payment()`? It seems to also cause a delay(/failure if the sync request fails) sometimes. What do you think?
> 
> https://github.com/breez/breez-sdk-liquid/blob/0c9d2b8cbeb269cdc79f6170a97398fce7f1652d/lib/core/src/sdk.rs#L1045-L1052

I agree it can be removed.

lib/core/src/sync/mod.rs
danielgranhao: Have you considered sending the listener events through a channel, thus isolating the listener/stream reconnection logic from the handling logic?

lib/core/src/sync/mod.rs
roeierez: Perhaps we don't need to change the notification interface. I think we can solve both issues (this and the comment above) by the following:
1. Adding sync right on start (on each restart)
2. Adding sync on each reconnect

hydra-yse: > If we restart after some time how do we poll for changes? It looks like we don't get any notification in that case.

As soon as we reconnect to the remote listener we get a ping back (Notification) which triggers an event loop. I've added this also to ensure the initial stream connection proceeds as gRPC has the whole "waiting for first message" hang when the listener is first instantiated.

roeierez (approved):
LGTM. Please add testing instructions.

lib/core/src/sync/mod.rs
```diff
@@ -80,6 +86,31 @@ impl SyncService {
...
+        Ok(match self.client.listen(req).await {
+            Ok(trigger) => Some(trigger),
+            Err(err) => {
+                log::warn!("Could not create new remote sync trigger: {err:?}");
+                None
+            }
+        })
+    }
+
+    fn handle_reconnect(self: Arc<Self>, remote_trigger: Arc<Mutex<RemoteStream>>) {
```
dangeross, line 100: Should we add a shutdown Receiver here?

hydra-yse: Yep, sounds good :+1:

dangeross (approved):
LGTM

roeierez:
@hydra-yse can we merge this?

PR #695, opened 2025-01-27
Allow incoming chain swap refund tx id overwrite

danielgranhao:
This PR changes the chain swap update query to allow overwrites of the refund tx id, which can now change due to potential fee bumps using RBF. 

Additionally, it simplifies the code using coalesce in the same query and the analogous ones from the other swap types.


dangeross (approved):
LGTM

hydra-yse (approved):

PR #696, opened 2025-01-27
Use cached electrum client connection for full_scan

roeierez:
This PR does 2 things:
1. Use a single point to scan the lwk_wallet (from the background task)
2. Use a cached electrum client (and connection) to scan the wallet. It saves the connection creation on each scan.

The ElectrumClient wraps a Client (from electrum-client) crate which holds the tcp connection.
I think we should aim for holding one client with reasonable timeout and on top of it we can instantiate several ElectrumClient when we need. The goal of the high level ElectrumClient is only to implement the API in a more convenient way (just a wrapper) by implementing BlockchainBackend for lwk_wallet. Currently the API doesn't allow it and I will work on that in a following PR.

Testing Notes: Test sync with two devices for send/receive

### DIFF
--- a/lib/core/src/wallet.rs
+++ b/lib/core/src/wallet.rs
@@ -2,12 +2,13 @@ use std::collections::HashMap;
 use std::fs::{self, create_dir_all};
 use std::io::Write;
 use std::path::PathBuf;
+use std::time::Instant;
 use std::{path::Path, str::FromStr, sync::Arc};
 
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use boltz_client::ElementsAddress;
-use log::{debug, warn};
+use log::{debug, info, warn};
 use lwk_common::Signer as LwkSigner;
 use lwk_common::{singlesig_desc, Singlesig};
 use lwk_wollet::elements::{AssetId, Txid};
@@ -103,6 +104,7 @@ pub(crate) struct LiquidOnchainWallet {
     config: Config,
     persister: Arc<Persister>,
     wallet: Arc<Mutex<Wollet>>,
+    electrum_client: Mutex<Option<ElectrumClient>>,
     working_dir: String,
     pub(crate) signer: SdkLwkSigner,
 }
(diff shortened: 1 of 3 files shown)


lib/core/src/wallet.rs
```diff
@@ -354,35 +357,46 @@ impl OnchainWallet for LiquidOnchainWallet {
...
-        let mut electrum_client = ElectrumClient::new(&electrum_url)?;
+        let full_scan_started = Instant::now();
+
+        // create electrum client if doesn't already exist
+        let mut electrum_client = self.electrum_client.lock().await;
+        if electrum_client.is_none() {
+            let electrum_url = ElectrumUrl::new(&self.config.liquid_electrum_url, true, true)?;
+            *electrum_client = Some(ElectrumClient::new(&electrum_url)?);
+        }
+        let client = electrum_client.as_mut().unwrap();
```
dangeross, line 368: Maybe avoid the unwrap?
```suggestion
        let client = electrum_client.as_mut()?;
```

dangeross (commented):
The electrum client gracefully handles reconnection right? I'm just wondering how it fairs with hibernation

danielgranhao (commented):
Assuming there are no issues with reconnection, as brought up by @dangeross, LGTM.

It would also be nice to know how much time is saved by always using the same client, even if only to track optimization progress.

roeierez:
> Assuming there are no issues with reconnection, as brought up by @dangeross, LGTM.
> 
> It would also be nice to know how much time is saved by always using the same client, even if only to track optimization progress.

for 30 transactions wallet the full_scan took ~3 seconds before the change and now it takes ~2.5 seconds.
It takes an additional 300ms-500ms to connect.



> The electrum client gracefully handles reconnection right? I'm just wondering how it fairs with hibernation

Yes, from my tests it reconnects.

danielgranhao:
> for 30 transactions wallet the full_scan took ~3 seconds before the change and now it takes ~2.5 seconds. It takes an additional 300ms-500ms to connect.

Ouch!

danielgranhao (approved):

lib/core/src/wallet.rs
roeierez: Fixed

dangeross (approved):

PR #697, opened 2025-01-28
Prevent missing notifications due to concurrent SDK instances

### LINKED ISSUE #636: Notification might be skipped if app keeps working in the background
There is no evidence this has happened in practice, but as @roeierez theorized [here](https://github.com/breez/breez-sdk-liquid/pull/634#discussion_r1899543562), it might be possible for the notification plugins to skip notifications if the app runs simultaneously in the background. 

The notification plugins wait for some events issued by the SDK. If the app is running in the background, 2 SDK instances will run in parallel, and it's possible only a single one will emit the event. Let's investigate if this is indeed a potential issue and if so, decide how to fix it.

danielgranhao:
Resolves #636 

This PR prevents a potential issue where notifications might be missed due to the SDK instance of a backgrounded app still being able to issue an event. Changes:

- Adds the swap id hash as a possible identifier for the `get_payment` method
- Exposes the claim tx id in payments to allow determining if the claim has been broadcasted
- Notification plugins, in addition to listening to events, now also poll the payment state

#### Test notes

- With the app in the background, payment notifications should work reliably without missed or duplicate notifications
    - Lightning payments
    - Onchain payments

### DIFF
--- a/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
+++ b/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
@@ -2,8 +2,10 @@ package breez_sdk_liquid_notification.job
 
 import android.content.Context
 import breez_sdk_liquid.BindingLiquidSdk
+import breez_sdk_liquid.GetPaymentRequest
 import breez_sdk_liquid.Payment
 import breez_sdk_liquid.PaymentDetails
+import breez_sdk_liquid.PaymentState
 import breez_sdk_liquid.PaymentType
 import breez_sdk_liquid.SdkEvent
 import breez_sdk_liquid_notification.Constants.DEFAULT_PAYMENT_RECEIVED_NOTIFICATION_TEXT
@@ -27,6 +29,11 @@ import breez_sdk_liquid_notification.NotificationHelper.Companion.notifyChannel
 import breez_sdk_liquid_notification.ResourceHelper.Companion.getString
 import breez_sdk_liquid_notification.SdkForegroundService
 import breez_sdk_liquid_notification.ServiceLogger
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.Json
 import java.security.MessageDigest
@@ -42,9 +49,12 @@ class SwapUpdatedJob(
     private val fgService: SdkForegroundService,
     private val payload: String,
     private val logger: ServiceLogger,
+    private val scope: CoroutineScope = CoroutineScope(Dispatchers.Default)
 ) : Job {
     private var swapIdHash: String? = null
     private var notified: Boolean = false
+    private var pollingJob: kotlinx.coroutines.Job? = null
+    private val pollingInterval: Long = 5000
 
     companion object {
         private const val TAG = "SwapUpdatedJob"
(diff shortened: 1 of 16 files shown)


lib/core/src/model.rs
```diff
@@ -724,6 +724,8 @@ pub enum ListPaymentDetails {
 pub enum GetPaymentRequest {
     /// The Lightning payment hash of the payment
     Lightning { payment_hash: String },
+    /// The SHA256-hash of the swap id in hex
+    SwapIdHash { hash: String },
```
dangeross, line 728: I think with the enum variants we wanted to keep it defined by payment methods. Maybe in this case, where the hash applies to both Lightning and Bitcoin methods, we flatten the enum into a struct and have an optional payment_hash and optional swap_id. I would keep it as swap_id and take into account in the query they could be using the id or id hash. What do you think?

hydra-yse (approved):
utACK

lib/core/src/model.rs
danielgranhao: As discussed out-of-band, we can even stop requiring the distinction between identifier types. [3f5100d](https://github.com/breez/breez-sdk-liquid/pull/697/commits/3f5100dc6c796ec57353ff63d26f5c52d622f784)

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -567,9 +567,8 @@ interface ListPaymentDetails {
     Bitcoin(string? address);
 };
 
-[Enum]
-interface GetPaymentRequest {
-    Lightning(string payment_hash);
+dictionary GetPaymentRequest {
+    string id;
```
roeierez, line 571: It is a bit strange to have a request by id and not having the id field on payment.  What do you think?

roeierez (changes requested):

lib/bindings/src/breez_sdk_liquid.udl
danielgranhao: Yeah, that's a valid concern. Maybe it would help to call it `identifier`, or maybe `reference`?

roeierez: What do  you think about a variant per id  type?

danielgranhao: I think that's also a valid option. The trade-off is between "type-safety" (integrators can very clearly tell which types are accepted) and a potentially easier-to-use interface. Without the enum, integrators can provide any id type without knowing its kind, which might also be helpful. I'm not strongly opinionated here.

roeierez: I don't have a strong opinion here either, @dangeross what do you think?

dangeross: I think in hindsight, although id is more extensible, the enum approach is clearer to the developer
```
interface GetPaymentRequest {
    PaymentHash(string payment_hash);
    SwapId(string swap_id);
}
```
I also wouldn't mind if TxId is included also

hydra-yse: IMO the enum would be clearer as well

danielgranhao: Let's go with enum them. I applied your suggestion @dangeross: [83e26ce](https://github.com/breez/breez-sdk-liquid/pull/697/commits/83e26ced2e8c428025108f775910aabca13e4a7d).

Decided to keep the txid out for now because we need to think about which kind of txid would be accepted and how to make that clear to integrators. We can add it in a later PR.

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
```diff
@@ -54,11 +64,62 @@ class SwapUpdatedJob(
...
+                    delay(pollingInterval)
+                } catch (e: Exception) {
+                    stopPolling(e)
+                    return@launch
+                }
+            }
+        }
+    }
+
+    private fun stopPolling(error: Exception? = null) {
```
dangeross, line 113: Do we also need to stop polling if onShutdown is called?

lib/bindings/langs/swift/Sources/BreezSDKLiquid/Task/SwapUpdated.swift
```diff
@@ -31,6 +34,52 @@ class SwapUpdatedTask : TaskProtocol {
...
+                }
+            } catch {
+                self.stopPolling(withError: error)
+            }
+        }
+
+        pollingTimer?.fire()
+    }
+    
+    private func stopPolling(withError error: Error? = nil) {
```
dangeross, line 75: Do we also need to stop polling if onShutdown is called?

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/SwapUpdated.kt
danielgranhao: I don't think so. That only happens when the OS wants to clean up the notification service so it would automatically stop polling.

lib/bindings/langs/swift/Sources/BreezSDKLiquid/Task/SwapUpdated.swift
danielgranhao: Same as above

dangeross (approved):
LGTM

roeierez (approved):

PR #698, opened 2025-01-28
Lock free chain service

roeierez:
Previously we had a lock wrapping the liquid chain service. This caused every endpoint that was called to block other endpoints. For example when the recoverer from the realtime sync fetching script history (which may take several seconds) was blocking a manual user Sync calls that use the same chain service and both were blocking other endpoint such as rescan_chain_swaps.
This PR removes that lock and only apply a lock internally on the tip endpoint since it is the only one that is mutable.
A dedicated tip_client is used for that as the alternative of creating electrum connection for each tip call is more expensive:
* cached electrum tip: ~70 ms
* new electrum tip: ~600 ms

Edit:  Applied the same approach also for bitcoin chain service.
Another significant bottleneck was get_script_history_with_retry which was internally retrying after 5 seconds. This has locked every concurrent usage of the bitcoin chain service during this waiting time.

### DIFF
--- a/lib/core/src/chain/bitcoin.rs
+++ b/lib/core/src/chain/bitcoin.rs
@@ -107,24 +107,27 @@ impl BitcoinChainService for HybridBitcoinChainService {
             maybe_popped_header = Some(header)
         }
 
-        match maybe_popped_header {
-            Some(popped_header) => {
-                let tip: HeaderNotification = popped_header.try_into()?;
-                *self.tip.lock().unwrap() = tip;
-            }
+        let new_tip = match maybe_popped_header {
+            Some(popped_header) => Some(popped_header.try_into()?),
             None => {
                 // https://github.com/bitcoindevkit/rust-electrum-client/issues/124
                 // It might be that the client has reconnected and subscriptions don't persist
                 // across connections. Calling `client.ping()` won't help here because the
                 // successful retry will prevent us knowing about the reconnect.
                 if let Ok(header) = self.client.block_headers_subscribe_raw() {
-                    let tip: HeaderNotification = header.try_into()?;
-                    *self.tip.lock().unwrap() = tip;
+                    Some(header.try_into()?)
+                } else {
+                    None
                 }
             }
+        };
+
+        let mut tip = self.tip.lock().unwrap();
+        if let Some(new_tip) = new_tip {
+            *tip = new_tip;
         }
 
-        Ok(self.tip.lock().unwrap().clone())
+        Ok(tip.clone())
     }
 
     fn broadcast(&self, tx: &Transaction) -> Result<Txid> {
--- a/lib/core/src/chain/bitcoin.rs
+++ b/lib/core/src/chain/bitcoin.rs
@@ -1,4 +1,4 @@
-use std::time::Duration;
+use std::{sync::Mutex, time::Duration};
 
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
@@ -24,7 +24,7 @@ use crate::{
 #[async_trait]
 pub trait BitcoinChainService: Send + Sync {
     /// Get the blockchain latest block
-    fn tip(&mut self) -> Result<HeaderNotification>;
+    fn tip(&self) -> Result<HeaderNotification>;
 
     /// Broadcast a transaction
     fn broadcast(&self, tx: &Transaction) -> Result<Txid>;
@@ -76,7 +76,7 @@ pub trait BitcoinChainService: Send + Sync {
 
 pub(crate) struct HybridBitcoinChainService {
     client: Client,
-    tip: HeaderNotification,
+    tip: Mutex<HeaderNotification>,
     config: Config,
 }
 impl HybridBitcoinChainService {
@@ -93,7 +93,7 @@ impl HybridBitcoinChainService {
 
         Ok(Self {
             client,
-            tip,
+            tip: Mutex::new(tip),
             config,
         })
     }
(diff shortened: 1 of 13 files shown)


lib/core/src/chain/bitcoin.rs
```diff
@@ -110,7 +110,7 @@ impl BitcoinChainService for HybridBitcoinChainService {
         match maybe_popped_header {
             Some(popped_header) => {
                 let tip: HeaderNotification = popped_header.try_into()?;
-                self.tip = tip;
+                *self.tip.lock().unwrap() = tip;
```
dangeross, line 113: Can we more safely unwrap the lock here and the lines below?

lib/core/src/chain/liquid.rs
```diff
@@ -58,20 +59,25 @@ pub trait LiquidChainService: Send + Sync {
...
+        })
     }
 }
 
 #[async_trait]
 impl LiquidChainService for HybridLiquidChainService {
-    async fn tip(&mut self) -> Result<u32> {
-        Ok(self.electrum_client.tip()?.height)
+    async fn tip(&self) -> Result<u32> {
+        Ok(self.tip_client.lock().unwrap().tip()?.height)
```
dangeross, line 80: Safe unwrap?
```suggestion
        Ok(self.tip_client.lock()?.tip()?.height)
```

lib/core/src/recover/recoverer.rs
```diff
@@ -166,6 +163,20 @@ impl Recoverer {
...
+                    println!("send_swap: {:?}", send_swap.id);
+                }
+                Swap::Receive(receive_swap) => {
+                    println!("receive_swap: {:?}", receive_swap.id);
+                }
+                Swap::Chain(chain_swap) => {
+                    println!("chain_swap: {:?}", chain_swap.id);
+                }
+            }
+        }
```
dangeross, line 179: Do you want to add these println's to the logs instead?

lib/core/src/sdk.rs
```diff
@@ -353,8 +351,13 @@ impl LiquidSdk {
             loop {
                 tokio::select! {
                     _ = interval.tick() => {
+                      info!("Track blocks loop ticked");
```
dangeross, line 354:
```suggestion
                        info!("Track blocks loop ticked");
```

dangeross (commented):
Looking a lot cleaner

lib/core/src/chain/bitcoin.rs
roeierez: I actually think we want to panic here. The only reason the lock() returns error is if one of the holders thread panics. I think we want to propagate that instead of hiding it and keep maintaining an unusable poisoned lock.

roeierez: The only reason for this to panic here is if one of the holder threads panics. I think we want to propagate that rather then maintaining an unusable poisoned lock.

lib/core/src/chain/liquid.rs
roeierez: Same here

lib/core/src/recover/recoverer.rs
roeierez: Sorry went in by mistake. Deleted the whole new block.

lib/core/src/sdk.rs
roeierez: fixed

lib/core/src/chain/bitcoin.rs
```diff
@@ -119,12 +119,12 @@ impl BitcoinChainService for HybridBitcoinChainService {
...
                 if let Ok(header) = self.client.block_headers_subscribe_raw() {
                     let tip: HeaderNotification = header.try_into()?;
-                    self.tip = tip;
+                    *self.tip.lock().unwrap() = tip;
                 }
             }
         }
 
-        Ok(self.tip.clone())
+        Ok(self.tip.lock().unwrap().clone())
```
danielgranhao, line 127: It seems like we could avoid the double locking in this method.

danielgranhao (commented):
Nice! Left just a minor improvement suggestion

lib/core/src/chain/bitcoin.rs
roeierez: I didn't find a way without holding the lock while inside the network request here: https://github.com/breez/breez-sdk-liquid/pull/698/files/fb37e7bcccad3df2513869a3fbb89c3132cc1a1f#diff-3271fa8e986665ac7110bad9866758686a5b41b5f60536fd27e48291b615e8ecR120

lib/core/src/chain/bitcoin.rs
dangeross: Just to understand, this means our `track_new_blocks` thread will panic in most cases and any call from the main thread if `next_unused_address()` is used?

lib/core/src/chain/bitcoin.rs
danielgranhao: Makes sense to avoid holding the lock for long yes. But the match statement could return a new tip, if any, and locking the current tip would only be done afterwards.

Example:

```rust
fn tip(&self) -> Result<HeaderNotification> {
    let mut maybe_popped_header = None;
    while let Some(header) = self.client.block_headers_pop_raw()? {
        maybe_popped_header = Some(header)
    }

    let new_tip: Option<HeaderNotification> = match maybe_popped_header {
        Some(popped_header) => Some(popped_header.try_into()?),
        None => {
            // https://github.com/bitcoindevkit/rust-electrum-client/issues/124
            // It might be that the client has reconnected and subscriptions don't persist
            // across connections. Calling `client.ping()` won't help here because the
            // successful retry will prevent us knowing about the reconnect.
            if let Ok(header) = self.client.block_headers_subscribe_raw() {
                Some(header.try_into()?)
            } else {
                None
            }
        }
    };

    let mut tip = self.tip.lock().unwrap();
    if let Some(new_tip) = new_tip {
        *tip = new_tip;
    }

    Ok(tip.clone())
}
```

roeierez: fixed

danielgranhao (approved):

dangeross (approved):
LGTM

hydra-yse (approved):
Looks good!

PR #702, opened 2025-01-29
Increase onchain fetching retries

danielgranhao:
This PR increases the number of retries on all onchain fetching methods that use retries and were retrying less than 10 times. 

I noticed in the wild that while processing a chain swap LockupFailed event, we were setting the status to `Failed` instead of `Refundable` because we would fail to find the user lockup in the 5 retries that were attempted. This should hopefully reduce the likelihood of this kind of issues.


dangeross (approved):
LGTM. I guess ok if it's in combination with the non-locking chain service

hydra-yse (approved):

roeierez (approved):

PR #703, opened 2025-01-29
Distinguish between failed fetch and tx not found

danielgranhao:
This fixes an issue where a `Refundable` chain swap would be set to `Failed` if the onchain script history fetching failed.

Now, we can distinguish between a failed fetch and a successful one where the the user lockup tx was not found.


dangeross (approved):
LGTM. So if it is a real error fetching script history, the recoverer will pick it up and resolve the swap state

roeierez (approved):

PR #705, opened 2025-01-29
Add support for multiple blockchain backends

hydra-yse:
This PR adds support for multiple blockchain backends to the SDK (Liquid and Bitcoin), including Esplora and its waterfalls variant

### DIFF
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -326,10 +326,15 @@ enum PaymentError {
     "SignerError",
 };
 
+[Enum]
+interface BlockchainExplorer {
+    Electrum(string url);
+    Esplora(string url, boolean use_waterfalls);
+};
+
 dictionary Config {
-    string liquid_electrum_url;
-    string bitcoin_electrum_url;
-    string mempoolspace_url;
+    sequence<BlockchainExplorer> liquid_explorers;
+    sequence<BlockchainExplorer> bitcoin_explorers;
     string working_dir;
     LiquidNetwork network;
     u64 payment_timeout_sec;
--- a/lib/core/src/chain/bitcoin.rs
+++ b/lib/core/src/chain/bitcoin.rs
@@ -1,10 +1,10 @@
 use std::{
     collections::HashMap,
-    sync::{Mutex, OnceLock},
+    sync::{Mutex, RwLock, RwLockReadGuard},
     time::Duration,
 };
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Result};
 use async_trait::async_trait;
 use electrum_client::{
     bitcoin::{
@@ -81,41 +81,67 @@ pub trait BitcoinChainService: Send + Sync {
     async fn recommended_fees(&self) -> Result<RecommendedFees>;
 }
 
+macro_rules! get_client {
+    ($chain_service:ident,$client:ident) => {
+        let lock = $chain_service.get_client()?;
+        let Some($client) = lock.as_ref() else {
+            bail!("Could not read Bitcoin electrum client");
+        };
+    };
+}
+
 pub(crate) struct HybridBitcoinChainService {
-    client: OnceLock<Client>,
     config: Config,
+    client: RwLock<Option<Client>>,
     last_known_tip: Mutex<Option<HeaderNotification>>,
 }
 impl HybridBitcoinChainService {
     pub fn new(config: Config) -> Result<Self, Error> {
         Ok(Self {
             config,
-            client: OnceLock::new(),
+            client: RwLock::new(None),
             last_known_tip: Mutex::new(None),
         })
     }
 
-    fn get_client(&self) -> Result<&Client> {
-        if let Some(c) = self.client.get() {
-            return Ok(c);
-        }
+    fn build_client(&self, url: &str, options: &ElectrumOptions) -> Result<Client> {
         let (tls, validate_domain) = match self.config.network {
             LiquidNetwork::Mainnet | LiquidNetwork::Testnet => (true, true),
             LiquidNetwork::Regtest => (false, false),
         };
-        let electrum_url =
-            ElectrumUrl::new(&self.config.bitcoin_electrum_url, tls, validate_domain)?;
-        let client = electrum_url.build_client(&ElectrumOptions { timeout: Some(3) })?;
+        let electrum_url = ElectrumUrl::new(url, tls, validate_domain)?;
+        Ok(electrum_url.build_client(options)?)
+    }
 
-        let client = self.client.get_or_init(|| client);
-        Ok(client)
+    fn get_client(&self) -> Result<RwLockReadGuard<Option<Client>>> {
+        let lock = self.client.read().unwrap();
+        if let Some(client) = lock.as_ref() {
+            if client.ping().is_ok() {
+                return Ok(lock);
+            }
+        }
+
+        let mut lock = self.client.write().unwrap();
+        for electrum_url in &self.config.bitcoin_electrum_explorers() {
+            if let Ok(electrum_client) =
+                self.build_client(electrum_url, &ElectrumOptions { timeout: Some(3) })
+            {
+                if electrum_client.ping().is_ok() {
+                    *lock = Some(electrum_client);
+                    drop(lock);
+                    return Ok(self.client.read().unwrap());
+                }
+            }
+        }
+
+        bail!("Could not create Bitcoin electrum client: no url specified in the configuration");
     }
 }
 
 #[async_trait]
 impl BitcoinChainService for HybridBitcoinChainService {
     fn tip(&self) -> Result<HeaderNotification> {
-        let client = self.get_client()?;
+        get_client!(self, client);
         let mut maybe_popped_header = None;
         while let Some(header) = client.block_headers_pop_raw()? {
             maybe_popped_header = Some(header)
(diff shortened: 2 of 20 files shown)


lib/core/src/chain/bitcoin.rs
```diff
@@ -80,13 +80,18 @@ pub(crate) struct HybridBitcoinChainService {
...
 
     /// Creates an Electrum client specifying non default options like timeout
-    pub fn with_options(config: Config, options: ElectrumOptions) -> Result<Self, Error> {
-        let electrum_url = ElectrumUrl::new(&config.bitcoin_electrum_url, true, true)?;
+    pub fn with_options(config: Config, options: ElectrumOptions) -> Result<Self> {
+        let Some(electrum_url) = config.bitcoin_electrum_explorers().first().cloned() else {
+            bail!(
+                "Could not create Bitcoin electrum client: no url specified in the configuration"
+            );
+        };
```
dangeross, line 93: Would it make sense for each chain service to also cycle though the electrum explorers list and pick/validate (using `client.ping()`) to select a client?

lib/core/src/wallet.rs
```diff
@@ -356,51 +416,52 @@ impl OnchainWallet for LiquidOnchainWallet {
...
-                        client,
-                    )?;
-                    *wallet = new_wallet;
-                    Ok(())
+        let mut clients = self.clients.lock().await;
+
+        let mut failed = vec![];
+        for explorer in &self.config.liquid_explorers {
+            let explorer_url = explorer.url();
+            match clients.get(explorer_url) {
```
dangeross, line 433: Seems like if we have a client initialised we should try to use it first, else if our first explorer in the list is unavailable we'll always be trying it and failing

hydra-yse: I see, so rather than iterating through the explorers we can have something like a priority queue. I'll try to implement it

lib/core/src/chain/bitcoin.rs
hydra-yse: So selecting the one with the lowest latency? It makes sense, but may add to the initialization overhead. @roeierez What do you think?

dangeross: Or just get the first valid/working one

danielgranhao: IMO getting the first working one makes sense as a first step. Selecting the one with lowest latency, besides the init overhead issue, may not yield the most performant one (server could be "close" but be slow to process actual requests)

lib/core/src/chain/bitcoin.rs
```diff
@@ -314,11 +319,24 @@ impl BitcoinChainService for HybridBitcoinChainService {
...
-        )
-        .await
-        .map_err(Into::into)
+        if self.config.bitcoin_esplora_explorers().is_empty() {
+            bail!("Cannot fetch recommended fees without specifying a Bitcoin Esplora backend.");
+        }
+
+        for esplora_url in self.config.bitcoin_esplora_explorers() {
+            let res =
+                get_parse_and_log_response(&format!("{esplora_url}/v1/fees/recommended"), true)
```
danielgranhao, line 328: Does esplora api expose this endpoint? I thought this was specific to mempool.space

hydra-yse: You're right, just checked. Mb for thinking mempool.space was using the same API. I'll add a new explorer type for that.

roeierez (changes requested):

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -326,10 +326,15 @@ enum PaymentError {
...
+    Electrum(string url);
+    Esplora(string url, boolean use_waterfalls);
+};
+
 dictionary Config {
-    string liquid_electrum_url;
-    string bitcoin_electrum_url;
-    string mempoolspace_url;
+    sequence<BlockchainExplorer> liquid_explorers;
+    sequence<BlockchainExplorer> bitcoin_explorers;
```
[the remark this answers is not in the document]
roeierez: I agree with the advantages but currently we are creating a regression since we are using mempool space internally and not exposing the mempool space url as before.
If we are only using the mempool space to fetch recommended fees perhaps we can implement that using electrum instead and abandon mempool space all together?

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -326,10 +326,15 @@ enum PaymentError {
     "SignerError",
 };
 
+[Enum]
+interface BlockchainExplorer {
+    Electrum(string url);
+    Esplora(string url, boolean use_waterfalls);
```
roeierez, line 332: Esplora won't work well on the sdk since it lacks the batching option. IMO we can go with Waterfall variant instead.

lib/core/src/model.rs
```diff
@@ -32,13 +32,26 @@ pub const LIQUID_FEE_RATE_SAT_PER_VBYTE: f64 = 0.1;
...
+pub enum BlockchainExplorer {
+    Electrum {
+        url: String,
+    },
+    Esplora {
+        url: String,
+        /// Whether or not to use the "waterfalls" extension
+        use_waterfalls: bool,
+    },
+    MempoolSpace {
```
roeierez, line 45: The MempoolSpace is only used to fetch recommended fees and not really used as fully blockchain explorer. I think we could treat it like before by having the mempoolspace_url in the config.
I think it will simplify a lot (and also be more clear) if we have only two variants here:
Electrum
Waterfall

roeierez (changes requested):

lib/core/src/model.rs
danielgranhao: I think we should have another look at dropping the mempool.space dependency. Esplora directly provides sat/vb fee rate estimates ([docs](https://github.com/blockstream/esplora/blob/master/API.md#get-fee-estimates)). Electrum provides fees, not fee rates, but we could figure out which tx size they are using for the estimations and convert them to a fee rate ([docs](https://electrum.readthedocs.io/en/latest/protocol.html#blockchain-estimatefee)).

lib/core/src/chain/bitcoin.rs
```diff
@@ -81,41 +81,67 @@ pub trait BitcoinChainService: Send + Sync {
...
-            ElectrumUrl::new(&self.config.bitcoin_electrum_url, tls, validate_domain)?;
-        let client = electrum_url.build_client(&ElectrumOptions { timeout: Some(3) })?;
+        let electrum_url = ElectrumUrl::new(url, tls, validate_domain)?;
+        Ok(electrum_url.build_client(options)?)
+    }
+
+    fn get_client(&self) -> Result<RwLockReadGuard<Option<Client>>> {
+        let lock = self.client.read().unwrap();
+        if let Some(client) = lock.as_ref() {
+            if client.ping().is_ok() {
```
danielgranhao, line 119: Isn't this adding an unnecessary round trip to every request?

lib/core/src/chain/bitcoin.rs
```diff
@@ -81,41 +81,67 @@ pub trait BitcoinChainService: Send + Sync {
...
+                    *lock = Some(electrum_client);
+                    drop(lock);
+                    return Ok(self.client.read().unwrap());
+                }
+            }
+        }
 
-        let client = self.client.get_or_init(|| client);
-        Ok(client)
+        bail!("Could not create Bitcoin electrum client: no url specified in the configuration");
```
danielgranhao, line 137:
```suggestion
        bail!("Could not create Bitcoin electrum client: no working instance is specified in the configuration");
```

lib/core/src/chain/bitcoin.rs
```diff
@@ -364,11 +392,26 @@ impl BitcoinChainService for HybridBitcoinChainService {
...
     async fn recommended_fees(&self) -> Result<RecommendedFees> {
-        get_parse_and_log_response(
-            &format!("{}/v1/fees/recommended", self.config.mempoolspace_url),
-            true,
-        )
-        .await
-        .map_err(Into::into)
+        if self.config.bitcoin_esplora_explorers().is_empty() {
+            bail!("Cannot fetch recommended fees without specifying a Bitcoin Esplora backend.");
+        }
```
danielgranhao, line 397: Should this check be on `bitcoin_mempool_space_explorers`?

danielgranhao (changes requested):

lib/core/src/wallet.rs
```diff
@@ -98,19 +100,101 @@ pub trait OnchainWallet: Send + Sync {
...
+            index,
+        ))
+    }
+
+    fn get_next_available_client(
+        config: &Config,
+        start_index: usize,
+    ) -> Result<(BlockchainClient, usize)> {
+        let total_explorers = config.liquid_explorers.len();
+        for index in start_index..start_index + total_explorers {
```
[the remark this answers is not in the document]
hydra-yse: It's iterating over each every time, so from current to last, independently from the position.

**Note:** Just noticed your comment below

> Would it make sense to allow cycling back to the beginning?

Yes, it's doing cyclical iteration here, wrapping by using the modulus.

lib/core/src/chain/bitcoin.rs
hydra-yse: Yes, but since it's a ping maybe it's worth doing for reliability? Not sure, would like everyone's opinion on this

hydra-yse:
Will reference all issues in #761

Issue #635, opened 2024-12-31
Regtest end-to-end test suite 


danielgranhao:
We should explore the possibility of running end-to-end tests on regtest. These would likely be time-consuming to run, but that's fine if we choose wisely when to run them. 

It should be possible to set up all of the SDK's runtime dependencies (Bitcoin, Liquid, Lightning, Swapper) locally, so I don't see any reason not to have this.

Progress: 

- [x] Support regtest in SDK: #707 
- [x] Set up a regtest environment: #707 
- [x] Create guide for how to use regtest setup (e.g. how to send and receive the different payment types): #811
- [x] Develop tests: #816 
- [x] Decide when to run them and configure CI accordingly: #816
- [ ] Add support for Wasm and run end-to-end tests in browser #846

PR #707, opened 2025-01-31
Support regtest

danielgranhao:
This PR is a first step towards #635. It doesn't provide any tests, but it already provides value as it allows using the CLI to manually test most (if not all - haven't fully tested) types of payments supported by the SDK. This is especially useful for chain swaps, as up until now we have been testing on mainnet, given the high fees on testnet3.

#### Changes:

- Adjusts the SDK to support regtest
    - Relies on changes to [boltz client](https://github.com/SatoshiPortal/boltz-rust/pull/88) and [sdk-common](https://github.com/breez/breez-sdk-greenlight/pull/1162)
    - Disables tls for electrum clients when on regtest
    - Adds variants to network enums
    - Sets default BTC asset id for regtest (existing migration is adjusted directly as it is assumed that no node will switch networks)
- Regtest setup is added. I took the [setup from boltz ](https://github.com/BoltzExchange/regtest)and adjusted it to our needs.
    - The main changes were the addition of mempool (given we rely on it for fee estimations), removal of Otterscan explorer, and setting the platform to allow the images to run emulated on arm Macs. (edit) Also, it now adds a local rtsync instance (thanks, @hydra-yse 🙂 )
    - It can be further optimized (e.g. it still runs anvil, ~~startup is a bit finicky and sometimes fails~~)

See the `regtest/README.md` for how to start and use the test setup.

I've tested basic cases (Sending and receiving to/from BTC address and sending and receiving using bolt11) and these have worked fine.


lib/core/src/sdk.rs
```diff
@@ -2105,6 +2107,7 @@ impl LiquidSdk {
                 };
 
                 let address = self.onchain_wallet.next_unused_address().await?.to_string();
+                println!("address: {address}");
```
dangeross, line 2110: Can remove this I guess

dangeross (commented):
Awesome! I'll give it a test. It looks like the UDL is missing LiquidNetwork::Regtest

lib/core/src/model.rs
```diff
@@ -120,6 +121,26 @@ impl Config {
...
+        Config {
+            liquid_electrum_url: "localhost:19002".to_string(),
+            bitcoin_electrum_url: "localhost:19001".to_string(),
+            mempoolspace_url: "http://localhost/api".to_string(),
+            working_dir: ".".to_string(),
+            cache_dir: None,
+            network: LiquidNetwork::Regtest,
+            payment_timeout_sec: 15,
+            zero_conf_min_fee_rate_msat: DEFAULT_ZERO_CONF_MIN_FEE_RATE,
+            sync_service_url: Some(BREEZ_SYNC_SERVICE_URL.to_string()),
```
dangeross, line 134: Does it make sense to disable the sync service or run a local service also?

regtest/docker-compose.yml
```diff
@@ -0,0 +1,653 @@
...
+      - electrs:19001
+      - --http-addr
+      - electrs:3002
+      - --daemon-rpc-addr
+      - bitcoind:18443
+      - --network
+      - regtest
+      - --jsonrpc-import
+    healthcheck:
+      test: ["CMD-SHELL", "curl -s $(hostname):3002/blocks/tip/height"]
```
dangeross, line 425: I got this error: 
```
invalid interpolation format for services.electrs-liquid.healthcheck.test.[]: "curl -s $(hostname):3003/blocks/tip/height". You may need to escape any $ with another $
```
```suggestion
      test: ["CMD-SHELL", "curl -s $$(hostname):3002/blocks/tip/height"]
```

regtest/docker-compose.yml
```diff
@@ -0,0 +1,653 @@
...
+      - /root/.elements
+      - --cookie
+      - regtest:regtest
+      - --network
+      - liquidregtest
+      - --parent-network
+      - regtest
+      - --jsonrpc-import
+    healthcheck:
+      test: ["CMD-SHELL", "curl -s $(hostname):3003/blocks/tip/height"]
```
dangeross, line 464: I got this error: 
```
invalid interpolation format for services.electrs-liquid.healthcheck.test.[]: "curl -s $(hostname):3003/blocks/tip/height". You may need to escape any $ with another $
```
```suggestion
      test: ["CMD-SHELL", "curl -s $$(hostname):3003/blocks/tip/height"]
```

regtest/docker-compose.yml
```diff
@@ -0,0 +1,653 @@
...
+      test: ["CMD", "lightning-cli", "--network=regtest", "getinfo"]
+      timeout: 1s
+      retries: 20
+      interval: 1s
+      start_period: 0s
+
+services:
+  cleanup:
+    <<: *base-scripts
+    restart: no
```
dangeross, line 92: I got this error:
```
services.cleanup.restart must be a string
```
All `restart:` values needed to be wrapped in double quotes (`restart: "no"`)

dangeross (commented):
These are the errors I got running locally. After fixing the ones below I eventually got:
```
Error response from daemon: failed to mount local volume: mount ./data/mempool-db:/var/lib/docker/volumes/mempool-db-data/_data, flags: 0x1000: no such file or directory
```

regtest/docker-compose.yml
danielgranhao: Worked on my machine so it's probably due to a different shell or compose version. Escaping the $ also works on mine. Applied in [06e22a9](https://github.com/breez/breez-sdk-liquid/pull/707/commits/06e22a9abfc961e056783a35ddd1fae696502426)

regtest/docker-compose.yml
danielgranhao: Applied in [06e22a9](https://github.com/breez/breez-sdk-liquid/pull/707/commits/06e22a9abfc961e056783a35ddd1fae696502426)

regtest/docker-compose.yml
danielgranhao: Same here, my compose didn't mind it. Applied your fix in [06e22a9](https://github.com/breez/breez-sdk-liquid/pull/707/commits/06e22a9abfc961e056783a35ddd1fae696502426).

lib/core/src/sdk.rs
danielgranhao: Yes, thanks. Went in by mistake. [0654cda](https://github.com/breez/breez-sdk-liquid/pull/707/commits/0654cda8331f4cd1ab0ac3a63ed615ea02012d35)

lib/core/src/model.rs
danielgranhao: I had thought about it. Agree it would be nice to run the sync service locally as well. I propose we do it in a later PR. I think it's fine to leave the actual sync service configured because it allows us to test sync using the CLI. Once we start running automated tests, then it becomes more important to use a local sync service to avoid excessively polluting the public sync deployment. Wdyt?

hydra-yse: Check out #709 :wink:

danielgranhao: Cool, thank you 😄  Do you mean the env vars for the sync service URL and API key? I don't think we need those, IMO it's fine to have the default regtest config set up for the regtest setup included in the repo.

dangeross (approved):
Tested ACK (needs a rebase).

It would be good to see (in a followup PR) a more detailed guide, maybe running though the life of a payment end-to-end.

hydra-yse (approved):
LGTM

PR #708, opened 2025-02-02
Remove electrum tip query on start

roeierez:
ElecturmClient needs to be initialized with the current tip which forced us to query right on start which caused up to several seconds of sdk initialization.
Since we don't really need lwk ElectrumClient to use electrum this PR shifts to use the underline Client directly and saves this startup time and reduced it to less than a millisecond. 

Edit: It turns out that also Electurm client initialization has some overhead as it tries to connect. I changed the code to use lazy initialization of electrum which causes the startup flow to complete instantly now.

### DIFF
--- a/lib/core/src/chain/bitcoin.rs
+++ b/lib/core/src/chain/bitcoin.rs
@@ -1,4 +1,4 @@
-use std::{sync::Mutex, time::Duration};
+use std::time::Duration;
 
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
@@ -76,7 +76,6 @@ pub trait BitcoinChainService: Send + Sync {
 
 pub(crate) struct HybridBitcoinChainService {
     client: Client,
-    tip: Mutex<HeaderNotification>,
     config: Config,
 }
 impl HybridBitcoinChainService {
@@ -88,14 +87,7 @@ impl HybridBitcoinChainService {
     pub fn with_options(config: Config, options: ElectrumOptions) -> Result<Self, Error> {
         let electrum_url = ElectrumUrl::new(&config.bitcoin_electrum_url, true, true)?;
         let client = electrum_url.build_client(&options)?;
-        let header = client.block_headers_subscribe_raw()?;
-        let tip: HeaderNotification = header.try_into()?;
-
-        Ok(Self {
-            client,
-            tip: Mutex::new(tip),
-            config,
-        })
+        Ok(Self { client, config })
     }
 }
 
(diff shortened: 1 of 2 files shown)


lib/core/src/chain/liquid.rs
```diff
@@ -60,52 +61,117 @@ pub trait LiquidChainService: Send + Sync {
...
+        println!("Fetching block headers");
+        while let Some(header) = client.block_headers_pop_raw()? {
+            maybe_popped_header = Some(header)
+        }
+
+        let new_tip: Option<u32> = match maybe_popped_header {
+            Some(popped_header) => Some(popped_header.height.try_into()?),
+            None => {
+                println!("Fetching block headers none");
+                // https://github.com/bitcoindevkit/rusprintln!("Fetching block headers");t-electrum-client/issues/124
```
dangeross, line 102:
```suggestion
                // https://github.com/bitcoindevkit/rust-electrum-client/issues/124
```

dangeross (commented):
Makes sense to reuse the `block_headers_subscribe_raw()` instead of caching the tip.

There are a couple of `println`'s that need removing

lib/core/src/chain/liquid.rs
roeierez: Done

lib/core/src/chain/liquid.rs
```diff
@@ -60,52 +61,112 @@ pub trait LiquidChainService: Send + Sync {
...
+        let client = self.get_client()?;
+        let mut maybe_popped_header = None;
+        while let Some(header) = client.block_headers_pop_raw()? {
+            maybe_popped_header = Some(header)
+        }
+
+        let new_tip: Option<u32> = match maybe_popped_header {
+            Some(popped_header) => Some(popped_header.height.try_into()?),
+            None => {
+                // https://github.com/bitcoindevkit/rusprintln!("Fetching block headers");t-electrum-client/issues/124
```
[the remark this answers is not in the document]
roeierez: Oh... how did it got there. 
Done

dangeross (approved):
LGTM

lib/core/src/chain/bitcoin.rs
```diff
@@ -114,30 +114,28 @@ impl BitcoinChainService for HybridBitcoinChainService {
...
             }
         };
 
-        let mut tip = self.tip.lock().unwrap();
-        if let Some(new_tip) = new_tip {
-            *tip = new_tip;
-        }
-
-        Ok(tip.clone())
+        let tip = new_tip.ok_or_else(|| anyhow!("Failed to get tip"))?;
```
danielgranhao, line 125: Nit: the `new_tip` name no longer makes sense. Maybe rename to `maybe_tip`? We can even get rid of it and return the tip directly within the `match`.

lib/core/src/chain/liquid.rs
```diff
@@ -60,52 +61,112 @@ pub trait LiquidChainService: Send + Sync {
...
+                // successful retry will prevent us knowing about the reconnect.
+                if let Ok(header) = client.block_headers_subscribe_raw() {
+                    Some(header.height.try_into()?)
+                } else {
+                    None
+                }
+            }
+        };
+
+        let tip: u32 = new_tip.ok_or_else(|| anyhow!("Failed to get tip"))?;
```
danielgranhao, line 112: Nit: Same as above. The name `new_tip` doesn't fit. We can return within the `match`.

danielgranhao (commented):
LGTM. Only a couple nits

danielgranhao (approved):

PR #709, opened 2025-02-02
Add real-time sync to regtest setup

hydra-yse:


danielgranhao (approved):

danielgranhao:
Awesome! Thank you @hydra-yse. Added a commit on top to expose the port and configure the newly added rtsync instance in the default regtest config.

PR #712, opened 2025-02-03
Lazily initialize Boltz client

### LINKED ISSUE #710: Lazy fetch swapper url
Currently we are fetching the swapper url as part of the sdk initialization flow.
We need to make sure it doesn't affect the sdk init and that the fetch is executed lazily when needed.

hydra-yse:
Closes #710.
This PR makes sure the Boltz client (along with its url and referral id) are only initialized when required.

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -113,16 +113,7 @@ impl LiquidSdk {
         req: ConnectWithSignerRequest,
         signer: Box<dyn Signer>,
     ) -> Result<Arc<LiquidSdk>> {
-        let maybe_swapper_proxy_url =
-            match BreezServer::new("https://bs1.breez.technology:443".into(), None) {
-                Ok(breez_server) => breez_server
-                    .fetch_boltz_swapper_urls()
-                    .await
-                    .ok()
-                    .and_then(|swapper_urls| swapper_urls.first().cloned()),
-                Err(_) => None,
-            };
-        let sdk = LiquidSdk::new(req.config, maybe_swapper_proxy_url, Arc::new(signer))?;
+        let sdk = LiquidSdk::new(req.config, Arc::new(signer))?;
         sdk.start()
             .inspect_err(|e| error!("Failed to start an SDK instance: {:?}", e))
             .await?;
@@ -154,11 +145,7 @@ impl LiquidSdk {
         Ok(())
     }
 
-    fn new(
-        config: Config,
-        swapper_proxy_url: Option<String>,
-        signer: Arc<Box<dyn Signer>>,
-    ) -> Result<Arc<Self>> {
+    fn new(config: Config, signer: Arc<Box<dyn Signer>>) -> Result<Arc<Self>> {
         if let Some(breez_api_key) = &config.breez_api_key {
             Self::validate_breez_api_key(breez_api_key)?
         }
@@ -189,11 +176,7 @@ impl LiquidSdk {
         let event_manager = Arc::new(EventManager::new());
         let (shutdown_sender, shutdown_receiver) = watch::channel::<()>(());
 
-        if let Some(swapper_proxy_url) = swapper_proxy_url {
-            persister.set_swapper_proxy_url(swapper_proxy_url)?;
-        }
-        let cached_swapper_proxy_url = persister.get_swapper_proxy_url()?;
-        let swapper = Arc::new(BoltzSwapper::new(config.clone(), cached_swapper_proxy_url));
+        let swapper = Arc::new(BoltzSwapper::new(config.clone(), persister.clone()));
         let status_stream = Arc::<dyn SwapperStatusStream>::from(swapper.create_status_stream()?);
 
         let recoverer = Arc::new(Recoverer::new(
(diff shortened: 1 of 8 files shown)


lib/core/src/swapper/boltz/mod.rs
```diff
@@ -421,24 +460,27 @@ impl Swapper for BoltzSwapper {
         Ok(tx_id)
     }
 
-    fn create_status_stream(&self) -> Box<dyn SwapperStatusStream> {
-        Box::new(BoltzStatusStream::new(&self.boltz_url))
+    fn create_status_stream(&self) -> Result<Box<dyn SwapperStatusStream>> {
+        Ok(Box::new(BoltzStatusStream::new(self.config.clone())))
```
danielgranhao, line 464: Does it need to return a result?

danielgranhao (commented):
Looks good!

dangeross (approved):
LGTM, just @danielgranhao's comment needs addressing

lib/core/src/swapper/boltz/mod.rs
```diff
@@ -421,24 +499,27 @@ impl Swapper for BoltzSwapper {
         Ok(tx_id)
     }
 
-    fn create_status_stream(&self) -> Box<dyn SwapperStatusStream> {
-        Box::new(BoltzStatusStream::new(&self.boltz_url))
+    fn create_status_stream(&self) -> Result<Box<dyn SwapperStatusStream>> {
+        Ok(Box::new(BoltzStatusStream::new(&self.get_client()?.url)))
```
roeierez: This will block on creating the client which miss the goal of not blocking the initialization.
Perhaps we can pass the url later when the status stream connects?

lib/core/src/sdk.rs
```diff
@@ -340,6 +316,50 @@ impl LiquidSdk {
...
+                                .unwrap_or_else(|err| {
+                                    error!("Could not set cached swapper proxy url: {err:?}")
+                                });
+                        }
+                        maybe_swapper_proxy_url
+                    }
+                    Err(_) => cloned.persister.get_swapper_proxy_url().unwrap_or(None),
+                };
+
+            if let Err(err) = cloned.swapper.connect(maybe_swapper_proxy_url.clone()) {
```
roeierez, line 343: The problem here is that by the time the swapper connects it is not available and users may attempt to use the sdk right after initialization receiving an error.
I think we can solve that by passing a closure or trait implementation for fetching the swapper url and let the swapper use that on demand (when creating the client). This is the only way I see we can defer the fetch of the url and maintaining the swapper immediate availability.

roeierez (changes requested):

hydra-yse:
Unfortunately this new impl has had the opposite effect: there's a pretty severe performance impact from ~250ms of initialization to ~1.5 seconds. I doubt it could be related to allocation time but so far that's the primary thing I suspect, as I've tried disabling the background tasks and even avoiding the swapper proxy url fetching altogether, with no improvement in startup time at all (meaning, it's related to the `new` method, which constructs all the services). I'll continue investigating tomorrow.

lib/core/src/swapper/boltz/mod.rs
```diff
@@ -28,48 +34,28 @@ pub(crate) mod bitcoin;
...
-    client: BoltzApiClientV2,
-    boltz_url: String,
+pub(crate) struct BoltzClient {
+    url: String,
     referral_id: Option<String>,
+    inner: BoltzApiClientV2,
+}
+
+pub(crate) type FetchProxyUrlFn =
+    dyn Fn() -> Pin<Box<dyn Future<Output = Result<Option<String>>> + Send>> + Send + Sync;
```
danielgranhao, line 44: I get the feeling this (i.e. the generalization of the URL fetching function) might be slightly unnecessarily complex. Could you elaborate on the requirement that justifies it?

danielgranhao (commented):
Nice that this leads to a faster initialization of the SDK, but do I get it right that the swapper client is lazily initialized? i.e. only after it is first needed?

If so, it would nice to try to try to initialize it in the background right after connecting the SDK, such that it's most likely ready when it's needed.

lib/core/src/swapper/boltz/mod.rs
hydra-yse: Sorry for the long block of text :pray:, hopefully this will cover all the considerations that have been made so far:

I understand it may be complex, but after some discussions with @roeierez we figured that it would be the best way to handle the issue. The requirements are:
- Being able to fetch the swapper proxy url from the persister if needed, without exposing it to the swapper
- Having an initial fetch of the proxy url from the `BreezServer`, without exposing it to the swapper
- Being able to retrieve this information at a later time than initialization of the swapper (so, on-demand)
- Being able to pass this information to the status stream when requested, specifically in the background thread that handles the ws connection

The only way to defer this computation to a secondary thread is to pass a method that allows retrieving it from that context. Statically passing won't do as it will slow down the startup. Other alternatives would involve spawning another thread and then using channels to communicate with the Swapper/SwapperStatusStream contexts, though I figured that would be even more complex than the current implementation. 

A way to simplify our current approach would be to pass a synchronous fn rather than an async, as it previously was (so avoiding pinning the future, and having that nasty type with the trait requirements). Though, as you also pointed out in a previous comment, that would need a workaround since the Breez server's fetching method is inherently non-blocking/async, and we'd need a way to return its output in a sync context.

Overall, I don't see many other alternatives that don't involve creating unnecessary structures for this single operation, but if you have any other ideas I'd be more than happy to hear them out (spent way too long on this, lol)

hydra-yse: No, it sounds good 👍

danielgranhao (approved):

roeierez (approved):
LGTM

PR #713, opened 2025-02-04
Cache last known tip

roeierez:
Until we have proper redundancy of chain services the tip endpoint that is used frequently may fail if electrum is not responsive or disconnected which is likely to happen in mobile environment.
This PR brings back the previous behavior where we keep the last known tip and return it on error.

### DIFF
--- a/lib/core/src/chain/liquid.rs
+++ b/lib/core/src/chain/liquid.rs
@@ -1,4 +1,4 @@
-use std::sync::OnceLock;
+use std::sync::{Mutex, OnceLock};
 use std::time::Duration;
 
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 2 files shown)


dangeross (approved):
Tested ACK

lib/core/src/chain/liquid.rs
```diff
@@ -109,7 +111,15 @@ impl LiquidChainService for HybridLiquidChainService {
             }
         };
 
-        new_tip.ok_or_else(|| anyhow!("Failed to get tip"))
+        let mut last_tip: std::sync::MutexGuard<'_, Option<u32>> =
+            self.last_known_tip.lock().unwrap();
+        match new_tip {
+            Some(header) => {
+                *last_tip = Some(header);
+                Ok(header)
```
danielgranhao, line 119: Nit:
```suggestion
            Some(height) => {
                *last_tip = Some(height);
                Ok(height)
```

danielgranhao (approved):

lib/core/src/chain/liquid.rs
roeierez: Done thanks.

PR #718, opened 2025-02-05
Persist and expose BIP353 addresses

### LINKED ISSUE #714: Persist BIP 353 Address on Payment Details

danielgranhao:
Resolves #714.

Depends on sdk-common PR: https://github.com/breez/breez-sdk-greenlight/pull/1164

This PR adds the bip353 address to `PaymentTxDetails`, persisting it locally and syncing it to other devices. The address is then exposed as an additional field of `PaymentDetails::Lightning`. Technically, bip353 can also be used to fetch onchain addresses, but that isn't currently supported. We can add it to `PaymentDetails::Bitcoin` once it is.

#### Test notes

Paying to a BIP353 address that resolves to a BOLT12 or an LNURL Pay results in a payment entry that shows the bip353 that was used. Note that this will only be testable once Misty exposes the new field.


lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -67,12 +67,19 @@ interface InputType {
     Bolt12Offer(LNOffer offer);
     NodeId(string node_id);
     Url(string url);
+    Bip353Address(ResolvedBip353 resolved_bip353);
```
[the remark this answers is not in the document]
erdemyerebasmaz: @roeierez After the integration, I didn't see much benefit to having the wrapper `Bip353Address` as an `InputType`. In fact, it made things a bit more complicated. It's better to add an optional `bip353_address` field to `Bolt12Offer` & `LnUrlPay` as such:
```
Bolt12Offer(LNOffer offer, string? bip353_address);
LnUrlPay(LnUrlPayRequestData request_data, string? bip353_address);
```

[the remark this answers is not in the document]
danielgranhao: I applied @erdemyerebasmaz's suggestion. I didn't pursue it initially because it felt weird, but indeed it's simpler.

lib/core/src/model.rs
```diff
@@ -1926,8 +1933,10 @@ pub struct PrepareLnUrlPayResponse {
     pub destination: SendDestination,
     /// The fees in satoshis to send the payment
     pub fees_sat: u64,
-    /// The [LnUrlPayRequestData] returned by [crate::input_parser::parse]
+    /// The [LnUrlPayRequestData] returned by [parse]
     pub data: LnUrlPayRequestData,
+    /// A BIP353 address, in case one was used in order to fetch the LNURL Pay request data.
+    pub bip353_address: Option<String>,
```
dangeross, line 1939: I didn't dig too deep, but is there a reason we need the `bip353_address` here and in the SendDestination? 

I'm wondering if it's enough add `bip353_address` to the Bolt11 variant also and to have `prepare_lnurl_pay()` fill in the `bip353_address` in the SendDestination response from its call to `prepare_send_payment()`

danielgranhao: I didn't consider that before, but I like that it improves consistency. See [d004300](https://github.com/breez/breez-sdk-liquid/pull/718/commits/d00430006353a09348e8f7f79d8f34012d64ce6c).

lib/core/src/sdk.rs
```diff
@@ -1243,19 +1253,34 @@ impl LiquidSdk {
                 self.pay_liquid(liquid_address_data.clone(), amount_sat, *fees_sat, true)
                     .await
             }
-            SendDestination::Bolt11 { invoice } => {
+            SendDestination::Bolt11 { invoice, .. } => {
                 self.pay_bolt11_invoice(&invoice.bolt11, *fees_sat).await
```
dangeross, line 1257: Would it be an option to insert the payment details with the `bip353_address` here (if set), then we don't have to deal with extracting the `bip353_address` from the SendDestination in `lnurl_pay()`?

lib/core/src/sdk.rs
```diff
@@ -3164,8 +3194,20 @@ impl LiquidSdk {
...
 
+                let destination =
+                    if let SendDestination::Bolt11 { invoice, .. } = prepare_response.destination {
+                        SendDestination::Bolt11 {
+                            invoice,
+                            bip353_address: req.bip353_address,
+                        }
+                    } else {
+                        return Err(LnUrlPayError::Generic {
+                            err: "SendDestination for LNURL Pay  is not BOLT11 invoice".to_string(),
```
dangeross, line 3205:
```suggestion
                            err: "SendDestination for LNURL Pay is not BOLT11 invoice".to_string(),
```

lib/core/src/sdk.rs
danielgranhao: Good idea! [10b31e3](https://github.com/breez/breez-sdk-liquid/pull/718/commits/10b31e3970c6aceb8fc3a8a1322a0e502c7bf722)

lib/core/src/sdk.rs
danielgranhao: Fixed

lib/core/src/sdk.rs
```diff
@@ -1243,19 +1253,50 @@ impl LiquidSdk {
...
+                {
+                    self.persister
+                        .insert_or_update_payment_details(PaymentTxDetails {
+                            tx_id: tx_id.clone(),
+                            destination: destination.clone(),
+                            description: None,
+                            lnurl_info: None,
+                            bip353_address: bip353_address.clone(),
+                        })?;
+                }
```
dangeross, line 1272: I think there's no point inserting if the bip353_address is not set, same for the bolt12

dangeross (approved):
LGTM

erdemyerebasmaz (approved):
LGTM

Issue #729, opened 2025-02-10
Synced sending payments shows wrong value temporarily


roeierez:
When sending a lightning payment it shows immediately on the other device but the amount is wrong, only after a while it is shown with the correct amount. Perhaps the transaction is considered first as liquid payment? Not sure it is the case because it takes some times until the amount correct itself and the sync happens immediately. 

Edit: It seems the amounts stay incorrect on the other device.

erdemyerebasmaz:
Hey @dangeross,

Can we close this now that
- https://github.com/breez/breez-sdk-liquid/pull/730

is merged?

PR #730, opened 2025-02-10
Send Swap recovery fixes

dangeross:
This PR fixes some issues found in send swap recovery:
1.  Fixes an issue where the lockup tx is seen as the claim tx if the non-local wallet has not yet seen the lockup tx (its not in the tx_map)
2. Fixes an issue where the non local swap has not yet had time to recover (is in state Created) and the SDK is restarted (set to TimedOut and no longer monitored)

**Testing notes:**
1.  Not directly testable, nothing occurs in the UI
2.  Previously the issue is seen on the non-local device where the send swap is always shown as a liquid payment. 
    - Send a lightning payment on the local device
    - On the non-local device wait for the liquid payment
    - Force quit the app
    - Reopen and check if the payment swap data recovers

### DIFF
--- a/lib/core/src/persist/send.rs
+++ b/lib/core/src/persist/send.rs
@@ -123,10 +123,11 @@ impl Persister {
         let con = self.get_connection()?;
         let mut where_clauses = vec!["state = :from_state".to_string()];
         if let Some(is_local) = is_local {
-            where_clauses.push(format!(
-                "(sync_state.is_local = {} OR sync_state.is_local IS NULL)",
-                is_local as i8
-            ));
+            let mut where_is_local = format!("sync_state.is_local = {}", is_local as u8);
+            if is_local {
+                where_is_local = format!("({} OR sync_state.is_local IS NULL)", where_is_local);
+            }
+            where_clauses.push(where_is_local);
         }
 
         let where_clause_str = where_clauses_to_string(where_clauses);
--- a/lib/core/src/persist/send.rs
+++ b/lib/core/src/persist/send.rs
@@ -11,6 +11,8 @@ use crate::sync::model::data::SendSyncData;
 use crate::sync::model::RecordType;
 use crate::{ensure_sdk, get_updated_fields};
 
+use super::where_clauses_to_string;
+
 impl Persister {
     pub(crate) fn insert_or_update_send_swap_inner(
         con: &Connection,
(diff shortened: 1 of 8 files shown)


lib/core/src/persist/send.rs
```diff
@@ -116,20 +116,36 @@ impl Persister {
...
+                is_local as i8
+            ));
+        }
+
+        let ids = self
+            .list_send_swaps_where(&tx, where_clauses)?
+            .iter()
+            .map(|s| s.id.clone())
+            .collect::<Vec<_>>()
+            .join(", ");
```
roeierez, line 136: Do we need this extra query or we can pass the where clause directly to the update statement and remove the need for a transaction?

roeierez (changes requested):
Looks good. One small comment

lib/core/src/persist/send.rs
dangeross: Fixed [ac6754d](https://github.com/breez/breez-sdk-liquid/pull/730/commits/ac6754d8e7bee538626e8b99af7bc7798ba4f7ee)

lib/core/src/persist/send.rs
```diff
@@ -116,15 +118,34 @@ impl Persister {
         &self,
         from_state: PaymentState,
         to_state: PaymentState,
+        is_local: Option<bool>,
     ) -> Result<()> {
         let con = self.get_connection()?;
+        let mut where_clauses = vec!["state = :from_state".to_string()];
+        if let Some(is_local) = is_local {
+            where_clauses.push(format!(
+                "(sync_state.is_local = {} OR sync_state.is_local IS NULL)",
```
danielgranhao, line 127: I realized now that `is_local` is not nullable. I think I was the one to start with also checking for a null value, but unless I'm missing something, it seems it isn't needed.

dangeross: There might not be a row for the swap_state, as it's either not arrived yet, or the rt sync is disabled

danielgranhao: Ahh right, exactly, now I remember. This brings me to another issue: wouldn't a null `is_local` mean it is local?

dangeross: In our case we only want to set the local ones

danielgranhao: True, but the method interface allows for is_local to be `Some(false)`, and in that case, it could behave unexpectedly.

danielgranhao (approved):
LGTM

dangeross:
Fixes #729

Issue #725, opened 2025-02-06
Recoverer/claim broadcasting race causes permanent duplicate payment in local instance


danielgranhao:
The issue occurs as follows:

1. A`track_new_blocks` loop iteration starts, and does a full scan of the liquid onchain wallet
2. A receive swap claim tx is broadcasted (and the claim tx id is set in the db)
3. The same `track_new_blocks` iteration continues and tries to recover data from the swap. 2 txs are found in the lockup script history, both pending (due to 0-conf), but neither are known local wallet txs (wallet is out of date), so we see the log: `Found unconfirmed lockup and refund txs while recovering data for Receive Swap Jusa6EMuaCRd`. End result is that the swap gets both its lockup and claim tx id cleared.
4. Swapper sends duplicate update `transaction.mempool` (see #704). Given the claim tx id was erased, we proceed with broadcasting a new claim tx
5. Only one of the claim txs confirms, but the other still gets added to the `payment_tx_data` table. There is no implemented mechanism to clean this table, causing a permanent pending payment to appear in the payments list. 

We should:

- [ ] Review tx recovery logic to see if it's possible to tolerate an out-of-date wallet tx set. If not, we need to drop reliance on the wallet txs - #751 
- [x] Prevent the race issue between the claiming and the recoverer tasks - #733 
- [x] As a fallback, add some mechanism for cleaning the `payment_tx_data` table as part of the recovery - #731

roeierez:
Great analysis @danielgranhao !

PR #731, opened 2025-02-10
Clean payment txs data on sync

danielgranhao:
This addresses part of #725 by adding a stale payment tx data cleanup. 

The idea is to look for pending payment txs that the liquid wallet does not recognize and delete them if they are older than a certain grace period (to allow for tx propagation through the network).

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -387,10 +387,6 @@ impl LiquidSdk {
                             .unwrap_or_else(|err| warn!("Could not update local tips: {err:?}"));
                         };
 
-                        if let Err(err) = cloned.onchain_wallet.full_scan().await {
-                          error!("Failed to scan wallet: {err:?}");
-                        }
-
                         // Only partial sync when there are no new Liquid or Bitcoin blocks
                         let partial_sync = (is_new_liquid_block || is_new_bitcoin_block).not();
                         _ = cloned.sync(partial_sync).await;
@@ -3020,6 +3016,11 @@ impl LiquidSdk {
         self.ensure_is_started().await?;
 
         let t0 = Instant::now();
+
+        if let Err(err) = self.onchain_wallet.full_scan().await {
+            error!("Failed to scan wallet: {err:?}");
+        }
+
         let is_first_sync = !self
             .persister
             .get_is_first_sync_complete()?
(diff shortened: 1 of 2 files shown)


lib/core/src/sdk.rs
```diff
@@ -2771,6 +2774,22 @@ impl LiquidSdk {
...
+            }) {
+                self.persister
+                    .delete_payment_tx_data(&unknown_unconfirmed_tx_data.tx_id)?;
+                info!(
+                    "Found an unknown unconfirmed tx and deleted it. Txid: {}",
+                    unknown_unconfirmed_tx_data.tx_id
+                );
+            } else {
+                debug!("Found an unknown unconfirmed tx. Keeping it to allow propagation through the network. Txid: {}", unknown_unconfirmed_tx_data.tx_id)
+            }
```
roeierez, line 2792: Do you think it is better to run it in the track_new_blocks loop right after the wallet full_scan ?
This way we are sure the wallet is synced before we delete the transaction while here if the full_scan haven't been running for a while (network issues) we could end up deleting an existing transaction.

Also in this case we may reduce the grace period even to 20-30 seconds IMO since it is only the time for the sdk to see that tx in the mempool.

roeierez (changes requested):

lib/core/src/sdk.rs
danielgranhao: Maybe I misunderstood your comment, but it already runs in that exact place: `sync_payments_with_chain_data` < `sync` < `track_new_blocks`.

In my experiments, the wallet is not seeing the tx until it confirms, so 20-30 seconds wouldn't be enough. I proposed 3 minutes because it's large enough that accidental deletions are fully prevented. We could reduce it, but IMO it's fine to have a slightly higher delay given the low likelihood of this correction being actually needed.

roeierez (approved):
LGTM

lib/core/src/sdk.rs
danielgranhao: After discussing out-of-band with @roeierez, we concluded there was a potential issue where if `sync()` is called by the user and if the last full scan is old (e.g. after the app is backgrounded for a while), a tx may be deleted unintentionally. Moving the full scan into the sync method prevents this issue: [12ad903](https://github.com/breez/breez-sdk-liquid/pull/731/commits/12ad903d77d82e2589fd46ae38560275d99ff118)

lib/core/src/sdk.rs
```diff
@@ -2771,6 +2770,22 @@ impl LiquidSdk {
                 // An unconfirmed tx that was not found in the payments table
                 self.persister.insert_or_update_payment_with_wallet_tx(tx)?;
             }
+            unconfirmed_txs_by_id.remove(&tx_id);
```
dangeross, line 2773: So we are removing it if its still seen by the wallet, not necessarily confirmed?

PR #732, opened 2025-02-11
Refactor sync notifications loop

roeierez:
Before this PR both local_sync_trigger and remot_sync_trigger were wrapped by Mutex and the reconnect was executed in a separate task resulting in deadlock after several disconnections.
This PR removes the locking and uses the main loop to reconnect at the same thread.
Also we no longer need to save the local_sync_trigger as part of the Sync object as we use it locally in the loop.

### DIFF
--- a/lib/core/src/sync/mod.rs
+++ b/lib/core/src/sync/mod.rs
@@ -4,9 +4,9 @@ use std::time::Duration;
 
 use anyhow::{anyhow, Result};
 use futures_util::TryFutureExt;
-use log::trace;
+use log::{info, trace, warn};
 use tokio::sync::mpsc::Receiver;
-use tokio::sync::{watch, Mutex};
+use tokio::sync::watch;
 use tokio::time::sleep;
 use tokio_stream::StreamExt as _;
 use tonic::Streaming;
@@ -36,10 +36,7 @@ pub(crate) struct SyncService {
     recoverer: Arc<Recoverer>,
     signer: Arc<Box<dyn Signer>>,
     client: Box<dyn SyncerClient>,
-    sync_trigger: Mutex<Receiver<()>>,
 }
-
-type RemoteStream = Option<Streaming<Notification>>;
 
 impl SyncService {
     fn set_sync_trigger(persister: Arc<Persister>) -> Receiver<()> {
@@ -56,14 +53,12 @@ impl SyncService {
         signer: Arc<Box<dyn Signer>>,
         client: Box<dyn SyncerClient>,
     ) -> Self {
-        let sync_trigger_rx = Self::set_sync_trigger(persister.clone());
         Self {
             remote_url,
             persister,
             recoverer,
             signer,
             client,
-            sync_trigger: Mutex::new(sync_trigger_rx),
         }
     }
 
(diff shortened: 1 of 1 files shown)


lib/core/src/sync/mod.rs
```diff
@@ -80,91 +75,64 @@ impl SyncService {
...
-                            self.clone().handle_reconnect(remote_sync_trigger.clone(), shutdown.clone());
+                loop {
+                    log::info!("realtime-sync: before tokio_select");
+                    tokio::select! {
+                        Some(_) = local_sync_trigger.recv() => self.run_event_loop().await,
+                        Some(msg) = remote_sync_trigger.next() => match msg {
+                            Ok(_) => self.run_event_loop().await,
+                            Err(err) => {
+                                log::warn!("realtime-sync: Received status {} from remote, attempting to reconnect.", err.message());
+                                break; // break the inner loop which will start the main loop by reconnecting
```
dangeross, line 126: Do we want to have a delay here too before the reconnect attempt?

lib/core/src/persist/sync.rs
```diff
@@ -246,10 +246,6 @@ impl Persister {
         record_type: RecordType,
         updated_fields: Option<Vec<String>>,
     ) -> Result<()> {
-        if self.sync_trigger.try_read().is_ok_and(|t| t.is_none()) {
-            return Ok(());
-        }
-
```
dangeross, line 252: If we remove this arn't we then filling up the sync tables even if there is no rt-sync activated?

roeierez: We do. Isn't it necessary to keep track for later when we activate the sync again? cc @hydra-yse

lib/core/src/sync/mod.rs
```diff
@@ -80,91 +75,64 @@ impl SyncService {
...
-                        // tokio_select! branch is never picked
-                        None => std::future::pending().await,
+                if shutdown.has_changed().unwrap_or(true) {
+                    return;
+                }
+                let mut remote_sync_trigger = match self.new_listener().await {
+                    Ok(trigger) => trigger,
+                    Err(e) => {
+                        warn!(
+                            "realtime-sync: new_listener returned error: {:?} waiting one second",
```
danielgranhao, line 111:
```suggestion
                            "realtime-sync: new_listener returned error: {:?} waiting 3 seconds",
```

danielgranhao (commented):
LGTM - just found a minor issue with one of the logs

lib/core/src/sync/mod.rs
roeierez: fixed.

danielgranhao (approved):

dangeross (approved):
LGTM, just need feedback from @hydra-yse

lib/core/src/persist/sync.rs
hydra-yse: You're both right, I think we should go for a fix here.
This line shouldn't be removed, as we don't want to clutter the db when rt-sync is unset. The behavior to choose now is what to do when it's switching between activated and de-activated.

The fix: we should clean the db and unset the cached remote url every time rt-sync has been deactivated (this will make it so that the records are pushed out on re-activation as if it were a new remote; if there already are records present, they will be skipped/merged as needed). Even if there are pending changes missed those will still be pushed out as long as a new remote url is set.

lib/core/src/sync/mod.rs
```diff
@@ -80,91 +75,64 @@ impl SyncService {
...
-                    let mut lock = remote_sync_trigger.lock().await;
-                    match lock.as_mut() {
-                        Some(trigger) => trigger.next().await,
-                        // Future hangs in case of a missing initial remote trigger,
-                        // tokio_select! branch is never picked
-                        None => std::future::pending().await,
+                if shutdown.has_changed().unwrap_or(true) {
+                    return;
+                }
+                let mut remote_sync_trigger = match self.new_listener().await {
```
hydra-yse, line 107: I like the inner/outer loop approach (which was also suggested by @dangeross). Just want to make sure we're fine with the fact that we won't be able to push records out until we can connect to the stream.

hydra-yse (commented):
Looks good, just two things:

hydra-yse (approved):
LGTM

lib/core/src/sync/mod.rs
roeierez: I think it is OK. We are also using the same underline channel so if it doesn't function well on the notification subscription it won't function also when pushing.

lib/core/src/persist/sync.rs
roeierez: Addressed. Moved to broadcast sender/receiver to allow lock free event broadcasting, setting the sync_enabled flag to the persister on construction time.

PR #733, opened 2025-02-11
Add tx propagation grace period to swap recovery

### LINKED ISSUE #725: Recoverer/claim broadcasting race causes permanent duplicate payment in local instance

danielgranhao:
Partly addresses #725.

This PR prevents clearing swap tx ids that we potentially broadcast for a certain grace period after the last update, only for local swaps.

Changes:

- Encapsulated metadata fields like the existing `index` and the new `last_updated_at` in a new `SwapMetadata` struct
- Added the `is_local` boolean to the metadata struct, removing the need to query it separately 
- When within the grace period, prevent clearing tx id of txs we potentially broadcasted
- Dropped the tx propagation grace period from 3 minutes to 30 secs since our electrum no longer has issues providing txs that are in the mempool

### DIFF
--- a/lib/core/src/recover/recoverer.rs
+++ b/lib/core/src/recover/recoverer.rs
@@ -234,7 +234,8 @@ impl Recoverer {
                     if is_local_within_grace_period && (lockup_is_cleared || refund_is_cleared) {
                         warn!(
                             "Local send swap {swap_id} was updated recently - skipping recovery \
-                        as it would clear a tx that may have been broadcasted by us"
+                        as it would clear a tx that may have been broadcasted by us. Lockup clear: \
+                        {lockup_is_cleared} - Refund clear: {refund_is_cleared}"
                         );
                         continue;
                     }
@@ -282,7 +283,7 @@ impl Recoverer {
                     if is_local_within_grace_period && claim_is_cleared {
                         warn!(
                             "Local receive swap {swap_id} was updated recently - skipping recovery \
-                        as it would clear a tx that may have been broadcasted by us"
+                        as it would clear a tx that may have been broadcasted by us (claim)"
                         );
                         continue;
                     }
@@ -318,10 +319,13 @@ impl Recoverer {
 
                         let claim_is_cleared = chain_swap.claim_tx_id.is_some()
                             && recovered_data.lbtc_claim_tx_id.is_none();
-                        if is_local_within_grace_period && claim_is_cleared {
+                        let refund_is_cleared = chain_swap.refund_tx_id.is_some()
+                            && recovered_data.btc_refund_tx_id.is_none();
+                        if is_local_within_grace_period && (claim_is_cleared || refund_is_cleared) {
                             warn!(
                             "Local incoming chain swap {swap_id} was updated recently - skipping recovery \
-                        as it would clear a tx that may have been broadcasted by us"
+                        as it would clear a tx that may have been broadcasted by us. Claim clear: \
+                        {claim_is_cleared} - Refund clear: {refund_is_cleared}"
                         );
                             continue;
                         }
@@ -375,11 +379,15 @@ impl Recoverer {
                             && recovered_data.lbtc_user_lockup_tx_id.is_none();
                         let refund_is_cleared = chain_swap.refund_tx_id.is_some()
                             && recovered_data.lbtc_refund_tx_id.is_none();
-                        if is_local_within_grace_period && (lockup_is_cleared || refund_is_cleared)
+                        let claim_is_cleared = chain_swap.claim_tx_id.is_some()
+                            && recovered_data.btc_claim_tx_id.is_none();
+                        if is_local_within_grace_period
+                            && (lockup_is_cleared || refund_is_cleared || claim_is_cleared)
                         {
                             warn!(
                             "Local outgoing chain swap {swap_id} was updated recently - skipping recovery \
-                        as it would clear a tx that may have been broadcasted by us"
+                        as it would clear a tx that may have been broadcasted by us. Lockup clear: \
+                        {lockup_is_cleared} - Refund clear: {refund_is_cleared} - Claim clear: {claim_is_cleared}"
                         );
                             continue;
                         }
--- a/lib/core/src/recover/recoverer.rs
+++ b/lib/core/src/recover/recoverer.rs
@@ -14,6 +14,7 @@ use lwk_wollet::WalletTx;
 use super::model::*;
 
 use crate::prelude::{Direction, Swap};
+use crate::sdk::NETWORK_PROPAGATION_GRACE_PERIOD;
 use crate::swapper::Swapper;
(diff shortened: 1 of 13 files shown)


lib/core/src/recover/recoverer.rs
```diff
@@ -302,51 +320,55 @@ impl Recoverer {
...
                     }
                     Direction::Outgoing => {
                         let Some(recovered_data) = recovered_chain_send_data.get(swap_id) else {
-                            log::warn!("Could not apply recovered data for outgoing Chain swap {swap_id}: recovery data not found");
+                            warn!("Could not apply recovered data for outgoing Chain swap {swap_id}: recovery data not found");
                             continue;
                         };
                         let is_expired = liquid_tip >= chain_swap.timeout_block_height;
                         if let Some(new_state) = recovered_data.derive_partial_state(is_expired) {
                             chain_swap.state = new_state;
```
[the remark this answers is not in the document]
danielgranhao: Oops, I overlooked that issue. I'll have to think about how to address it. Thanks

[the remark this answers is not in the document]
danielgranhao: I've prevented state updates if the recovered data is inconsistent with the existing swap data in [70e4bef](https://github.com/breez/breez-sdk-liquid/pull/733/commits/70e4befe47f684af3294a55164a24b161940e30d), which should address the issue.

dangeross (commented):
Did you consider filtering out the monitored swaps within their grace period before running them through the recoverer? 

Seems the cleaner approach in most cases (unless I'm missing something), though receive swaps are a special case because it's how we determine the MRH address is paid to.

danielgranhao:
> Did you consider filtering out the monitored swaps within their grace period before running them through the recoverer?
> 
> Seems the cleaner approach in most cases (unless I'm missing something), though receive swaps are a special case because it's how we determine the MRH address is paid to.

As discussed out-of-band, that affects UX as it delays payment confirmation. Also, there is the problem you mention regarding MRH.

I've updated the PR. Now, when within the grace period, we simply skip recovery if it would clear a tx that we might have just broadcasted, which is much cleaner than what I was going for before.

lib/core/src/recover/recoverer.rs
```diff
@@ -268,10 +300,6 @@ impl Recoverer {
                         .mrh_tx_id
                         .clone()
                         .map(|history_tx_id| history_tx_id.txid.to_string());
-                    receive_swap.lockup_tx_id = recovered_data
-                        .lockup_tx_id
-                        .clone()
-                        .map(|history_tx_id| history_tx_id.txid.to_string());
```
dangeross, line 274: Should this be re-added?

danielgranhao: Thanks! Missed that. Fixed

lib/core/src/recover/recoverer.rs
```diff
@@ -280,9 +312,20 @@ impl Recoverer {
...
                     Direction::Incoming => {
                         let Some(recovered_data) = recovered_chain_receive_data.get(swap_id) else {
-                            log::warn!("Could not apply recovered data for incoming Chain swap {swap_id}: recovery data not found");
+                            warn!("Could not apply recovered data for incoming Chain swap {swap_id}: recovery data not found");
                             continue;
                         };
+
+                        let claim_is_cleared = chain_swap.claim_tx_id.is_some()
+                            && recovered_data.lbtc_claim_tx_id.is_none();
+                        if is_local_within_grace_period && claim_is_cleared {
```
dangeross, line 321: It could perhaps clear or reset (address reuse) the refund_tx_id here also

danielgranhao: Thanks. I was missing the btc part on both directions of chain swap. I've added that as well as the claim for outgoing chain swaps in [914ed5c](https://github.com/breez/breez-sdk-liquid/pull/733/commits/914ed5cf4f7a045c3a5be9a22469d3f732013a50).

lib/core/src/recover/recoverer.rs
```diff
@@ -280,9 +313,23 @@ impl Recoverer {
...
-                            log::warn!("Could not apply recovered data for incoming Chain swap {swap_id}: recovery data not found");
+                            warn!("Could not apply recovered data for incoming Chain swap {swap_id}: recovery data not found");
                             continue;
                         };
+
+                        let claim_is_cleared = chain_swap.claim_tx_id.is_some()
+                            && recovered_data.lbtc_claim_tx_id.is_none();
+                        let refund_is_cleared = chain_swap.refund_tx_id.is_some()
+                            && recovered_data.btc_refund_tx_id.is_none();
+                        if is_local_within_grace_period && (claim_is_cleared || refund_is_cleared) {
```
dangeross, line 324: I still think it might miss the case where we just broadcast a second refund but only the first is seen by the recoverer (It will revert to the first refund_tx_id and state Failed briefly).

roeierez: @dangeross wouldn't this be a temporary situation? Wouldn't it be fixed as soon as the previous (overridden)  refund will get out of the mempool?

dangeross: Yeah, it's temporary. It might just be for one sync cycle

danielgranhao: > I still think it might miss the case where we just broadcast a second refund but only the first is seen by the recoverer (It will revert to the first refund_tx_id and state Failed briefly).

Yeah, I agree that might happen. Maybe it's fine for now. I don't see a simple way of addressing that issue. Maybe the solution is having a more complete and complex model that tracks multiple refunds separately

dangeross (approved):
LGTM

PR #736, opened 2025-02-12
Deduct send amount if tx not seen yet

### LINKED ISSUE #723: Sending a lightning payment doesn't deduct the balance immediately
Probably because the wallet will only detect that on next sync.
I think we can deduct it directly from the persisted balance which will be overridden on next sync.

dangeross:
This PR when updating the wallet info, if a send payment tx_id is not in the wallet txs then deduct it from the balance

```
[2025-02-12 16:41:57.961 DEBUG breez_sdk_liquid::sdk:2945] Onchain wallet balance: 32776 sats
...
[2025-02-12 16:42:07.396 DEBUG breez_sdk_liquid::chain_swap:768] Successfully broadcast lockup transaction for Chain Swap M2AAgpQ4nYyA. Lockup tx id: edd94496fa499cb6385b58df2ba4fcd3842d5f21dbbbe1156e7f91ca1eac57ed
[2025-02-12 16:42:07.473 DEBUG breez_sdk_liquid::sdk:2934] Deducting 26142 sats from balance
[2025-02-12 16:42:07.473 DEBUG breez_sdk_liquid::sdk:2945] Onchain wallet balance: 6634 sats
...
[2025-02-12 16:42:17.156 DEBUG breez_sdk_liquid::sdk:2830] New tx detected in the last 30 secs. Txid: edd94496fa499cb6385b58df2ba4fcd3842d5f21dbbbe1156e7f91ca1eac57ed
[2025-02-12 16:42:17.204 DEBUG breez_sdk_liquid::sdk:2945] Onchain wallet balance: 6634 sats

```
Fixes #723

**Testing notes:**
- After sending a payment the sent amount should be deducted from the wallet balance within ~10 secs (before the "Processing payment" modal is popped in Misty)

### DIFF
--- a/lib/core/src/chain_swap.rs
+++ b/lib/core/src/chain_swap.rs
@@ -781,9 +781,10 @@ impl ChainSwapHandler {
     }
 
     // Updates the swap without state transition validation
-    pub(crate) fn update_swap(&self, updated_swap: ChainSwap) -> Result<(), PaymentError> {
+    pub(crate) fn update_swap(&self, updated_swap: ChainSwap) -> Result<bool, PaymentError> {
         let swap = self.fetch_chain_swap_by_id(&updated_swap.id)?;
-        if updated_swap != swap {
+        let is_updated = updated_swap != swap;
+        if is_updated {
             info!(
                 "Updating Chain swap {} to {:?} (user_lockup_tx_id = {:?}, server_lockup_tx_id = {:?}, claim_tx_id = {:?}, refund_tx_id = {:?})",
                 updated_swap.id,
@@ -796,7 +797,7 @@ impl ChainSwapHandler {
             self.persister.insert_or_update_chain_swap(&updated_swap)?;
             let _ = self.subscription_notifier.send(updated_swap.id);
         }
-        Ok(())
+        Ok(is_updated)
     }
 
     // Updates the swap state with validation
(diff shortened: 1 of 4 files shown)


lib/core/src/chain_swap.rs
```diff
@@ -781,9 +781,10 @@ impl ChainSwapHandler {
     }
 
     // Updates the swap without state transition validation
-    pub(crate) fn update_swap(&self, updated_swap: ChainSwap) -> Result<(), PaymentError> {
+    pub(crate) fn update_swap(&self, updated_swap: ChainSwap) -> Result<bool, PaymentError> {
```
roeierez, line 784: Instead of changing the update_swaps functions  did you consider only trigger the update_wallet_info on the event of PendingPayment ?

dangeross: I can test this. I think it would be good to cover all payment related events

danielgranhao (approved):
LGTM

lib/core/src/sdk.rs
```diff
@@ -2838,7 +2838,9 @@ impl LiquidSdk {
             }
         }
 
-        self.update_wallet_info().await?;
+        if !partial_sync {
+            self.update_wallet_info().await?;
+        }
```
danielgranhao, line 2843: Do we want to update the wallet info only when there are new blocks?

roeierez (approved):
LGTM

danielgranhao (approved):

PR #740, opened 2025-02-14
Extract `description` from `metadata_str` for LNURL payments

### LINKED ISSUE #741: Ensure LNURL payment descriptions are persisted correctly
Currently, the correct LNURL payment description is available and saved in `PaymentTxDetails`, but there's mismatch between `Payment` data.

Instead of storing the actual LNURL payment description, the placeholder "Lightning payment" is being used by the `Persister`.

For LNURL payments, the short mandatory description from the metadata string should be used as payment description.

erdemyerebasmaz:
Fixes #741 

This PR 
- Uses the invoice's `description` first; if it's empty, falls back to extracting the `description` from `metadata_str` as the `description` value saved to `PaymentTxDetails`.
- Fixes how `description` is persisted for LNURLp payments, as the `description` value saved to `PaymentTxDetails` was previously ignored by the persister.

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -3349,6 +3349,8 @@ impl LiquidSdk {
             None => None,
         };
 
+        let description = extract_description_from_metadata(&prepare_response.data);
+
         let lnurl_pay_domain = match prepare_response.data.ln_address {
             Some(_) => None,
             None => Some(prepare_response.data.domain),
@@ -3360,7 +3362,7 @@ impl LiquidSdk {
                 .insert_or_update_payment_details(PaymentTxDetails {
                     tx_id,
                     destination,
-                    description: prepare_response.comment.clone(),
+                    description,
                     lnurl_info: Some(LnUrlInfo {
                         ln_address: prepare_response.data.ln_address,
                         lnurl_pay_comment: prepare_response.comment,
@@ -3569,6 +3571,18 @@ impl LiquidSdk {
     pub fn init_logging(log_dir: &str, app_logger: Option<Box<dyn log::Log>>) -> Result<()> {
         crate::logger::init_logging(log_dir, app_logger)
     }
+}
+
+/// Extracts `description` from `metadata_str`
+fn extract_description_from_metadata(request_data: &LnUrlPayRequestData) -> Option<String> {
+    let metadata = request_data.metadata_vec().ok()?;
+    metadata
+        .iter()
+        .find(|item| item.key == "text/plain")
+        .map(|item| {
+            info!("Extracted payment description: '{}'", item.value);
+            item.value.clone()
+        })
 }
 
 #[cfg(test)]
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -3349,7 +3349,10 @@ impl LiquidSdk {
             None => None,
         };
 
-        let description = extract_description_from_metadata(&prepare_response.data);
+        let description = payment
+            .details
+            .get_description()
+            .or_else(|| extract_description_from_metadata(&prepare_response.data));
 
         let lnurl_pay_domain = match prepare_response.data.ln_address {
             Some(_) => None,
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -1681,6 +1681,14 @@ impl PaymentDetails {
         }
     }
 
+    pub(crate) fn get_description(&self) -> Option<String> {
+        match self {
+            Self::Lightning { description, .. }
+            | Self::Bitcoin { description, .. }
+            | Self::Liquid { description, .. } => Some(description.clone()),
+        }
+    }
+
     pub(crate) fn is_lbtc_asset_id(&self, network: LiquidNetwork) -> bool {
         match self {
             Self::Liquid { asset_id, .. } => {
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -19,6 +19,7 @@ use crate::sync::model::RecordType;
 use crate::{get_invoice_description, utils};
 use anyhow::{anyhow, Result};
 use boltz_client::boltz::{ChainPair, ReversePair, SubmarinePair};
+use log::warn;
 use lwk_wollet::WalletTx;
 use migrations::current_migrations;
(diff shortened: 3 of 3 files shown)


lib/core/src/sdk.rs
```diff
@@ -3349,20 +3349,34 @@ impl LiquidSdk {
...
-            Some(_) => None,
-            None => Some(prepare_response.data.domain),
-        };
         if let (Some(tx_id), Some(destination)) =
             (payment.tx_id.clone(), payment.destination.clone())
         {
+            // Uses existing `ln_address` if it's available, otherwise fallback to `ln_address` from `metadata_str`
+            let ln_address = derive_ln_address(
+                prepare_response.data.ln_address.clone(),
+                &prepare_response.data.metadata_str,
```
dangeross, line 3358: You can use `data.metadata_vec()` to parse this
https://github.com/breez/breez-sdk-greenlight/blob/bb68587b221cd1c4a0471b7bc98b3cd0082144f7/libs/sdk-common/src/input_parser.rs#L804-L808

lib/core/src/sdk.rs
```diff
@@ -3571,6 +3585,138 @@ impl LiquidSdk {
...
+            info!("Using existing LN address: {}", addr);
+            Some(addr)
+        }
+        None => match extract_ln_address_from_metadata(metadata_str, domain) {
+            Some(addr) => {
+                info!("Extracted LN address from metadata_str: {}", addr);
+                Some(addr)
+            }
+            None => {
+                warn!(
```
dangeross, line 3605: It's optional so it's enough as debug/info

lib/core/src/sdk.rs
```diff
@@ -3571,6 +3585,138 @@ impl LiquidSdk {
...
+                None
+            }
+        },
+    }
+}
+
+/// Extracts `ln_address` from `metadata_str` and validates it against given `domain`
+fn extract_ln_address_from_metadata(metadata_str: &str, expected_domain: &str) -> Option<String> {
+    if metadata_str.is_empty() {
+        warn!("Metadata string is empty, no LN address available.");
```
dangeross, line 3618: We already log above when none is found

lib/core/src/sdk.rs
```diff
@@ -3571,6 +3585,138 @@ impl LiquidSdk {
...
+                                identifier, expected_domain
+                            );
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    warn!("No valid LN address found in metadata_str.");
```
dangeross, line 3657: We already log above when none is found

lib/core/src/sdk.rs
```diff
@@ -3571,6 +3585,138 @@ impl LiquidSdk {
...
+            warn!("No valid description found");
+            None
+        }
+    }
+}
+
+/// Extracts `description` from `metadata_str`
+fn extract_description_from_metadata(metadata_str: &str) -> Option<String> {
+    if metadata_str.is_empty() {
+        warn!("Metadata string is empty, no description available.");
```
dangeross, line 3691: We already log above when none is found

lib/core/src/sdk.rs
```diff
@@ -3571,6 +3585,138 @@ impl LiquidSdk {
...
+                    if let Some(description) = item.get(1).and_then(|v| v.as_str()) {
+                        info!("Extracted payment description: '{}'", description);
+                        return Some(description.to_string());
+                    }
+                }
+            }
+        }
+    }
+
+    warn!("No valid description found in metadata_str.");
```
dangeross, line 3716: We already log above when none is found

lib/core/src/sdk.rs
erdemyerebasmaz: @dangeross Thanks for the review! Most of the logs will be removed later on.

I'm aware of the redundancies, they were helpful for tracking the code without needing to rebuild. I'm keeping them for debugging purposes for now. Redundant ones will be removed when PR is marked for review.

lib/core/src/sdk.rs
```diff
@@ -3349,20 +3349,34 @@ impl LiquidSdk {
...
 
-        let lnurl_pay_domain = match prepare_response.data.ln_address {
-            Some(_) => None,
-            None => Some(prepare_response.data.domain),
-        };
         if let (Some(tx_id), Some(destination)) =
             (payment.tx_id.clone(), payment.destination.clone())
         {
+            // Uses existing `ln_address` if it's available, otherwise fallback to `ln_address` from `metadata_str`
+            let ln_address = derive_ln_address(
```
roeierez, line 3356: What is the use case for searching in the metadata? Isn't data.ln_address enough?

erdemyerebasmaz: I can't reproduce it but there was a case where `data.ln_address` was missing, but the [internet identifier](https://github.com/lnurl/luds/blob/luds/16.md) was available on the `metadata_str` on a payment to WoS ln address.

Either way, this looks like something to be handled on the input parser itself.

erdemyerebasmaz: Removed and reduced the scope of the PR to only extract `description` from `metadata_str`.

lib/core/src/sdk.rs
```diff
@@ -3360,7 +3362,7 @@ impl LiquidSdk {
                 .insert_or_update_payment_details(PaymentTxDetails {
                     tx_id,
                     destination,
-                    description: prepare_response.comment.clone(),
+                    description,
```
roeierez, line 3365: What exactly do we consider as description for lnurlpay payments? I can see several options but just want to be accurate at what we want. Also is it aligned with the description we show in greenlight sdk?

erdemyerebasmaz: On app level, we show `text/long-desc` if it's available and fallback to mandatory `text/plain` field on `metadata_str` for `description`.

Currently, both SDK's do not persist this value as `description` for LNURLp payments.

roeierez: @erdemyerebasmaz  I see. Here we only take the text/plain ignoring the text/long-desc is that OK?

roeierez: Why not taking the description from payment.description? It should contain the invoice description shouldn't it?

roeierez (approved):
LGTM

PR #747, opened 2025-02-17
Real-time Sync: Avoid running event loop on local changes

hydra-yse:
This PR makes sure that we're not running the event loop when we're the broadcasters of a record.

**Note:** If the record is pulled (non-local), even if we're the ones broadcasting the changes on top we'll still execute the flow. This only applies when we're the initiators.

@roeierez **Explanation:** The reason for omitting `new_revision` from the `Notification` object is that we run into a race condition between the remote listener and push threads. The notification from the remote may be received before we persist the pushed record (and its state), so we need to `lock_outgoing_record` before pushing, using a mock revision. Since we then have a mock revision there would be no need to compare that with the notification's `new_revision` field.

### DIFF
--- a/lib/core/src/sync/mod.rs
+++ b/lib/core/src/sync/mod.rs
@@ -31,6 +31,7 @@ pub(crate) mod model;
 
 pub(crate) struct SyncService {
     remote_url: String,
+    client_id: String,
     persister: Arc<Persister>,
     recoverer: Arc<Recoverer>,
     signer: Arc<Box<dyn Signer>>,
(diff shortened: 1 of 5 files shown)


lib/core/src/sync/mod.rs
```diff
@@ -435,27 +440,38 @@ impl SyncService {
...
+            self.persister.set_sync_state(SyncState {
+                data_id: data_id.to_string(),
+                record_id: record_id.to_string(),
+                record_revision: reply.new_revision,
+                is_local,
+            })?;
+
+            Ok(())
+        } {
+            self.persister.unlock_outgoing_record(record_id)?;
```
roeierez, line 471: What happen if we exit before executing this line? Seems like the record is locked forever?

roeierez (changes requested):

roeierez (approved):
LGTM

danielgranhao (approved):
LGTM

PR #751, opened 2025-02-18
Tolerate older wallet tx map during tx recovery

### LINKED ISSUE #725: Recoverer/claim broadcasting race causes permanent duplicate payment in local instance

danielgranhao:
Resolves #725.

This PR adds to the tx recovery methods the assumption that the wallet tx map might be out of date (in practice, this means that wallet txs might be included in the swap script histories, but not in the wallet tx map iself).

I've reviewed the current tx recovery implementations, and they mostly conform to this assumption. The main issue I found was that incoming chain swaps didn't consider the possibility of a swapper refund (which was an issue regardless of the new assumptions).

### DIFF
--- a/lib/core/src/recover/recoverer.rs
+++ b/lib/core/src/recover/recoverer.rs
@@ -545,6 +545,10 @@ impl Recoverer {
     }
 
     /// Reconstruct Send Swap tx IDs from the onchain data and the immutable data
+    ///
+    /// The implementation tolerates a `tx_map` that is older than the `send_histories_by_swap_id`
+    /// in the sense that no incorrect data is recovered. Transactions that are missing from `tx_map`
+    /// are simply not recovered.
     fn recover_send_swap_tx_ids(
         &self,
         tx_map: &TxMap,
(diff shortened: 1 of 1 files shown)


dangeross (approved):
LGTM

roeierez (approved):
LGTM

PR #759, opened 2025-02-26
Fix payment tx data cleanup logic

danielgranhao:
This PR fixes an issue introduced in #731.

Unknown unconfirmed txs were previously obtained by filtering `tx_map` txs out of `unconfirmed_txs_by_id`. The issue was that `tx_map` already had all swap-related txs removed, causing some swap transactions to be treated as unknown and potentially deleted (once the grace period elapsed).


dangeross (approved):
LGTM

I understand this as unconfirmed swap related txs could potentially be removed even if the wallet still sees them, so now it only removes unconfirmed txs the wallet doesn't see (after grace period)

roeierez (approved):

PR #760, opened 2025-02-27
Subscribe incoming synced records to swapper updates

roeierez:
This PR ensures the incoming synced records are subscribed (if needed) to swapper status stream updates. Previously it was only executed on restart.


dangeross (approved):
Looks good

lib/core/src/swapper/reconnect_handler.rs
danielgranhao, line 1: Nit: should we update the file name accordingly?

danielgranhao (approved):
LGTM - just a nit

lib/core/src/swapper/reconnect_handler.rs
roeierez: Sure. done.

PR #761, opened 2025-03-03
Add Esplora client to chain services

hydra-yse:
## Overview
This PR adds support for the Esplora client by allowing the user to choose a `BlockchainBackend` through the configuration. The backend is then picked up by the chain handlers and is used to query the blockchain.

## Todo
- [ ] Add support for batch Bitcoin Esplora methods (`get_transactions`, `get_scripts_history` and `get_scripts_balance`)
- [x] Add support for missing Bitcoin Esplora methods (`get_scripts_balance`)
- [x] Feature-lock Electrum branch (WASM)


dangeross (commented):
I feel like it would be cleaner to put the BitcoinClient / LiquidClient behind traits and have an implementation each for Electrum and Esplora in separate files (locking out Electrum for wasm). Then use a factory/fn to select them

lib/core/src/chain/bitcoin/client.rs
```diff
@@ -0,0 +1,205 @@
...
+use anyhow::{bail, Context, Result};
+
+use crate::prelude::*;
+use bitcoin::{Script, Transaction, Txid};
+#[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+use {electrum_client::ElectrumApi as _, lwk_wollet::ElectrumOptions};
+
+use crate::model::{BlockchainExplorer, LiquidNetwork, RecommendedFees};
+
+pub(crate) enum BitcoinClient {
```
roeierez, line 12: Currently we have several abstractions:
BitcoinClient (uses electrum/esplora underneath) and BitcoinChainService. 
Given that the BitcoinClient only proxies the request to the underline electrum/esplora I am not sure we need this extra layer. I find myself a bit confused traveling between the implementations.
Why not use the electrum/esplora directly under the current abstractions of BitcoinChainService & LiquidChainService ?

lib/core/src/chain/bitcoin/service.rs
```diff
@@ -82,147 +76,130 @@ pub trait BitcoinChainService: Send + Sync {
...
+    async fn set_client(&self) -> Result<()> {
+        let lock = self.client.read().await;
+        if let Some(client) = lock.as_ref() {
+            if client.is_available() {
+                return Ok(());
+            }
+        }
+
+        let mut lock = self.client.write().await;
+        *lock = Some(self.init_client()?);
```
roeierez, line 113: This introduces network request inside the lock which will cause it to hang. I suggest to remove the is_available for now.

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -325,10 +325,15 @@ enum PaymentError {
...
+    Electrum(string url);
+    Esplora(string url, boolean use_waterfalls);
+};
+
 dictionary Config {
-    string liquid_electrum_url;
-    string bitcoin_electrum_url;
-    string mempoolspace_url;
+    sequence<BlockchainExplorer> liquid_explorers;
+    sequence<BlockchainExplorer> bitcoin_explorers;
```
roeierez, line 336: I suggest to start simple without redundancy. one explorer for bitcoin, one for liquid.

roeierez (changes requested):

lib/core/src/chain/bitcoin/client.rs
hydra-yse: Tbf it's just to avoid cluttering the chain services implementation with logic about switching clients, but I'm not against removing it

lib/core/src/chain/bitcoin/mod.rs
```diff
@@ -0,0 +1,91 @@
...
+        tx_id: &str,
+        tx_hex: &str,
+        verify_confirmation: bool,
+    ) -> Result<Transaction>;
+
+    /// Get the recommended fees, in sat/vbyte
+    async fn recommended_fees(&self) -> Result<RecommendedFees>;
+}
+
+pub(crate) struct HybridBitcoinChainService<C> {
```
roeierez, line 77: Now that we have a trait is there a reason for the generics here? If no it seems better to have two non generic concrete implementations EsporaChainService ElectrumChainService.

roeierez (changes requested):
Looks good! dropped one comment.

lib/core/src/recover/handlers/handle_chain_receive_swap.rs
```diff
@@ -282,19 +281,19 @@ impl ChainReceiveSwapHandler {
 
 pub(crate) struct RecoveredOnchainDataChainReceive {
     /// LBTC tx locking up funds by the swapper
-    pub(crate) lbtc_server_lockup_tx_id: Option<HistoryTxId>,
+    pub(crate) lbtc_server_lockup_tx_id: Option<History<elements::Txid>>,
```
dangeross, line 284: Could potentially have the type `LBtcHistory` as
```
pub(crate) type LBtcHistory = History<elements::Txid>;
```
and use `Vec<LBtcHistory>` explicitly elsewhere. Same for `BtcHistory`

lib/core/src/swapper/boltz/mod.rs
```diff
@@ -59,14 +59,14 @@ impl<P: ProxyUrlFetcher> BoltzSwapper<P> {
             config: config.clone(),
             liquid_electrum_client: ElectrumLiquidClient::new(
                 config.network.into(),
-                &config.liquid_electrum_url,
+                config.liquid_explorer.url(),
```
dangeross, line 62: The use of Esplora for Boltz will be addressed in a followup PR right? (cc @danielgranhao)

dangeross (commented):
Looking good!

There seems to be an issue with the generated RN code

lib/core/src/model.rs
```diff
@@ -189,6 +229,30 @@ impl Config {
             LiquidNetwork::Regtest => BOLTZ_REGTEST,
         }
     }
+
+    pub(crate) fn bitcoin_chain_service(&self) -> Arc<dyn BitcoinChainService> {
+        match self.bitcoin_explorer {
+            BlockchainExplorer::Esplora { .. } => {
+                Arc::new(EsploraBitcoinChainService::new(self.clone()))
+            }
+            #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
```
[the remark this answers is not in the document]
roeierez: I am also not sure we should exclude here the electrum branch. Now that we have the connect method at the wasm trait perhaps we only need to  validate the config there.

roeierez (approved):
LGTM. I dropped one comment.

lib/core/src/model.rs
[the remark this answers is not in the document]
hydra-yse: My understanding was that by using the cfg flag we would disallow compiling the Electrum client only on the wasm toolchain, as it's unsupported. So 1) it won't show in the enum options 2) it will cause a compilation issue if you attempt to use it. Am I missing something or is there a better approach?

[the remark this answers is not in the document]
roeierez: I don't think you are missing. We just think out loud if we need it or not. In general I am a fan of reducing the amount of compilation flags as much as we can but TBH I am not strongly opinionated here.

lib/core/src/recover/handlers/handle_chain_receive_swap.rs
hydra-yse: Yes, I was actually debating whether to use one of the other. I think I'll go for what you suggested

lib/core/src/swapper/boltz/mod.rs
hydra-yse: Yes

lib/core/src/model.rs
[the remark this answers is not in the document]
dangeross: It won't include this block at all for wasm compilation, so the compiler will complain the `BlockchainExplorer::Electrum` branch is missing for the `match`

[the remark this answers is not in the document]
hydra-yse: You're right, my bad, I did not put the feature flag on the main enum. I'll remove it altogether

hydra-yse:
Flutter CI seems to be failing due to FRB not compiling. It can be excluded for now

lib/core/src/chain/bitcoin/esplora.rs
```diff
@@ -0,0 +1,347 @@
...
+                    })
+                    .collect()
+            })
+            .collect();
+        Ok(scripts_utxos)
+    }
+
+    async fn script_get_balance(&self, script: &Script) -> Result<BtcScriptBalance> {
+        let client = self.get_client()?;
+        let utxos = client.scripthash_outputs(script).await?;
```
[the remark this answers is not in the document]
hydra-yse: I'll test for this, though the Boltz side needs to be completed first (CLI won't run otherwise). My intuition for now is that the output value is always in absolute value therefore it will always be positive, even for an unconfirmed spend. The electrum client docs state that some servers use negative values. Which standard should we follow?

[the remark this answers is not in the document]
dangeross: I always thought this is electrum's way of showing (unconfirmed) utxos - unconfirmed txos, therefore can be negative.
 
(cc @roeierez)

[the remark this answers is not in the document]
dangeross: Looking through the code, I don't think it has a huge impact. It's primary used in the amountless swap validation looking for a positive unconfirmed balance
https://github.com/breez/breez-sdk-liquid/blob/fb232f070793dba59f101ea2e3378fadd43d54f5/lib/core/src/chain_swap.rs#L469-L470

dangeross (approved):
LGTM (pending script balance check)

lib/core/src/chain/bitcoin/esplora.rs
```diff
@@ -0,0 +1,347 @@
...
+                false => break,
+            }
+        }
+        Ok(script_history)
+    }
+
+    async fn get_script_utxos(&self, script: &Script) -> Result<Vec<Utxo>> {
+        let client = self.get_client()?;
+        let utxos = client
+            .scripthash_outputs(script)
```
danielgranhao, line 147: Maybe this isn't an issue, but do I get it right that `get_script_utxos()` by relying on `scripthash_outputs()`, will exclude outputs that have pending spends, but `get_scripts_utxos()` will do the opposite? If so, is that intentional?

lib/core/src/chain/bitcoin/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+use crate::{
+    bitcoin,
+    model::{BtcHistory, BtcScriptBalance, RecommendedFees, Utxo},
+};
+use bitcoin::{Address, Script, Transaction, Txid};
+
+pub(crate) type History = BtcHistory;
+
+/// Trait implemented by types that can fetch data from a blockchain data source.
+#[allow(dead_code)]
```
danielgranhao, line 15: Is this needed?

lib/core/src/chain/bitcoin/esplora.rs
```diff
@@ -0,0 +1,347 @@
...
+                        value: Amount::from_sat(out.value),
+                        script_pubkey: script.into(),
+                    },
+                ))
+            })
+            .collect();
+        Ok(utxos)
+    }
+
+    async fn get_scripts_utxos(&self, scripts: &[&Script]) -> Result<Vec<Vec<Utxo>>> {
```
danielgranhao, line 166: It would be nice to have some tests for the logic in this method, given its intricacies.

lib/core/src/recover/handlers/tests/mod.rs
```diff
@@ -9,70 +9,73 @@ pub mod handle_receive_swap_tests_integration;
 pub mod handle_send_swap_tests;
 pub mod handle_send_swap_tests_integration;
 
-// Helper function to create a HistoryTxId for testing
-mod test {
-    use std::{collections::BTreeMap, str::FromStr};
```
danielgranhao, line 14: Why expose the test helper functions outside of tests?

lib/core/src/recover/model.rs
```diff
@@ -63,7 +30,7 @@ impl TxMap {
     }
 }
 
-/// Swap list containing all swap data indexed by swap ID
+/// Swap immutable data
```
danielgranhao, line 33: Is this correct?

danielgranhao (commented):
Awesome! Just a few comments

lib/core/src/model.rs
```diff
@@ -81,9 +112,12 @@ pub struct Config {
 impl Config {
     pub fn mainnet(breez_api_key: Option<String>) -> Self {
         Config {
-            liquid_electrum_url: "elements-mainnet.breez.technology:50002".to_string(),
-            bitcoin_electrum_url: "bitcoin-mainnet.blockstream.info:50002".to_string(),
-            mempoolspace_url: "https://mempool.space/api".to_string(),
```
danielgranhao, line 86: Since we don't depend on mempool anymore, we can get rid of the mempool instance on the regtest setup. It's best to rebase first (there are regtest changes to be taken in).

lib/core/src/chain/bitcoin/esplora.rs
hydra-yse: The `scripthash_outputs` only includes the unspent outputs (not sure whether or not it considers pending spends, the API reference does not give any hints on that regard). The best way is to test it out in a refund for example, and see whether the pending utxo is getting picked up or not.
 
Note: Aboute the name, I've had to add this method myself, as it was missing from the official `esplora_client` crate. Should I rename it to `scripthash_utxos` for clarity?

lib/core/src/recover/handlers/tests/mod.rs
hydra-yse: It's already locked by tests, the whole file is under a cfg flag. The mod was namespacing something that didn't need to be.

lib/core/src/recover/model.rs
hydra-yse: Slipped from the rebase, thanks

lib/core/src/chain/bitcoin/esplora.rs
danielgranhao: I believe it considers pending spends, and if it does, it won't be possible to RBF refund txs. I see that `get_scripts_utxos` does implement the logic to not consider pending spends. Why not reuse the method by providing a single script like we do on electrum?

Regarding the name, I do think `scripthash_utxos` would be clearer.

danielgranhao (commented):
My last worry is regarding refund RBF, otherwise LGTM. Left a nit as well.

lib/core/src/chain/bitcoin/esplora.rs
danielgranhao: Also, if it doesn't consider pending spends, we can greatly simplify the `get_scripts_utxos` implementation.

lib/core/src/model.rs
```diff
@@ -2177,6 +2239,67 @@ pub struct AcceptPaymentProposedFeesRequest {
...
+    /// Confirmation height of txid
+    ///
+    /// -1 means unconfirmed with unconfirmed parents
+    ///  0 means unconfirmed with confirmed parents
+    pub height: i32,
+}
+pub(crate) type LBtcHistory = History<elements::Txid>;
+pub(crate) type BtcHistory = History<bitcoin::Txid>;
+
+impl<Txid> History<Txid> {
```
danielgranhao, line 2254:
```suggestion
impl<T> History<T> {
```

danielgranhao (approved):
LGTM!

PR #762, opened 2025-03-03
Recovery: Ensure wallet and chain service tips are within leeway

hydra-yse:
This PR:
- Ensures the wallet and chain service tips are aligned (with 3 block leeway) when calling the `recover_from_onchain` method
- Ensure the error is handled properly depending on the caller:
  1. Real-time sync: We want to throw an error and retry on next event loop
  2. SDK sync loop: We want to handle the error gracefully and ensure we are not disabling event notifications

### DIFF
--- a/lib/core/src/recover/recoverer.rs
+++ b/lib/core/src/recover/recoverer.rs
@@ -173,7 +173,7 @@ impl Recoverer {
         &self,
         swaps: &mut [Swap],
     ) -> Result<HashMap<Txid, WalletTx>> {
-        let wallet_tip = self.onchain_wallet.tip().await.height();
+        let wallet_tip = self.onchain_wallet.tip().await;
         let liquid_tip = self.liquid_chain_service.tip().await?;
         if wallet_tip.abs_diff(liquid_tip) > LIQUID_TIP_LEEWAY {
             let msg = format!("Skipping recovery: wallet and liquid chain service tips are too far apart - chain_service_tip {liquid_tip} wallet_tip {wallet_tip}");
--- a/lib/core/src/recover/recoverer.rs
+++ b/lib/core/src/recover/recoverer.rs
@@ -176,9 +176,8 @@ impl Recoverer {
         let wallet_tip = self.onchain_wallet.tip().await;
         let liquid_tip = self.liquid_chain_service.tip().await?;
         if wallet_tip.abs_diff(liquid_tip) > LIQUID_TIP_LEEWAY {
-            let msg = format!("Skipping recovery: wallet and liquid chain service tips are too far apart - chain_service_tip {liquid_tip} wallet_tip {wallet_tip}");
-            log::debug!("{msg}");
-            anyhow::bail!(msg)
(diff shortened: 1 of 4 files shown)


lib/core/src/recover/recoverer.rs
```diff
@@ -171,6 +173,14 @@ impl Recoverer {
         &self,
         swaps: &mut [Swap],
     ) -> Result<HashMap<Txid, WalletTx>> {
+        let wallet_tip = self.onchain_wallet.tip().await;
+        let liquid_tip = self.liquid_chain_service.tip().await?;
+        if wallet_tip.abs_diff(liquid_tip) > LIQUID_TIP_LEEWAY {
```
roeierez, line 178: I am ok with this as a safety measure since it is better not to run the sync if the tips are not in sync but it seems that right now on restoring an old node the recovery will be skipped leaving the records without the swaps information.
Shouldn't we also ensure the sync is running after first_sync_completed?

lib/core/src/sdk.rs
```diff
@@ -3138,21 +3138,34 @@ impl LiquidSdk {
...
+            match self.sync_payments_with_chain_data(partial_sync).await {
+                Ok(_) => {
+                    if is_first_sync {
+                        self.persister.set_is_first_sync_complete(true)?;
+                    }
+                    Result::<()>::Ok(())
+                }
+                Err(err) => anyhow::bail!("Could not sync payments with chain data: {err:?}"),
+            }
+        };
```
roeierez, line 3152: is this async code necessary? My understanding is the only change we need here is to not set the first sync as completed in case of error.

roeierez (changes requested):

lib/core/src/recover/recoverer.rs
dangeross: We only do a partial sync every 10 seconds (receive swaps), so in the negative flow we wait until the next block to do the full sync. Perhaps we should always full sync until the first sync is complete?

roeierez: I think if we do what Ross suggests then it is the only change necessary for this PR and we can end up with a tiny diff. I am in favor.

roeierez: @hydra-yse @dangeross  Going over this again and giving some more thinking I think the first sync is a private case. For example if the user opens a syncing wallet that was closed for a long time we can fall into the same issue.
What do you think about adding a wallet sync in the recovery only in the case we have a big diff (more than LIQUID_TIP_LEEWAY) in the blokchain tips?

roeierez: The LIQUID_TIP_LEEWAY is 3 blocks so I think even if I open a device after 24 hours of shutdown (which is totally reasonable) we will fall into this issue. That's why I suggested to run full sync in the recoverer in case we have a gap. I think now that we don't have performance/locking issues it should be fine.

roeierez (approved):
LGTM

Nackoo2000:
verified

PR #765, opened 2025-03-05
Handle transition error on TimedOut

dangeross:
It can be the case that the sending payment timeout occurs before we receive the payment pending event, causing an erorr when trying to transition the payment to TimedOut.

This PR:
- Handle transition error on TimedOut when sending payment (by fetching and returning the payment on error)
- Also remove LwkError from the interface errors as it's an internal error


danielgranhao (approved):
LGTM

erdemyerebasmaz (approved):
LGTM

Please remove LwkError exception handler on Misty Breez.

lib/core/src/sdk.rs
```diff
@@ -1961,16 +1961,23 @@ impl LiquidSdk {
...
-                            _ => ()
+                                }),
+                            _ => Ok(())
+                        };
+                        return match update_res {
+                            Ok(_) => Err(PaymentError::PaymentTimeout),
+                            Err(_) => {
+                                // Not able to transition the payment state to TimedOut, which means the payment
+                                // state progressed but we didn't see the event before the timeout
+                                self.persister.get_payment(&expected_swap_id).ok().flatten().ok_or(PaymentError::generic("Payment not found"))
```
roeierez, line 1978: So in that case the user gets a payment in created state? Shouldn't we allow the transition to timeout and fail the payment instead?

roeierez (approved):

PR #768, opened 2025-03-10
Wallet: Switch clients on full sync failure

hydra-yse:
## Overview

This PR adds the ability for the wallet to switch between available Electrum/Esplora clients when performing a full sync. In case its current client is not available, the sync iterates cyclically over the available clients, ultimately failing if none succeed. Once an available client is found, it will be reused on the next iteration.

## Notes

### DIFF
--- a/lib/core/src/wallet.rs
+++ b/lib/core/src/wallet.rs
@@ -5,23 +5,23 @@ use std::path::PathBuf;
 use std::time::Instant;
 use std::{path::Path, str::FromStr, sync::Arc};
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use async_trait::async_trait;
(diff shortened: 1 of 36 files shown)


lib/core/src/wallet.rs
```diff
@@ -181,6 +206,20 @@ impl LiquidOnchainWallet {
         .map_err(|e| anyhow!("Invalid descriptor: {e}"))?;
         Ok(descriptor_str.parse()?)
     }
+
+    async fn execute_scan(&self, client: &mut LiquidClient, index: u32) -> Result<()> {
```
dangeross, line 210: Maybe change this to derivation_index too just for clarity
```suggestion
    async fn execute_scan(&self, client: &mut LiquidClient, derivation_index: u32) -> Result<()> {
```

PR #771, opened 2025-03-12
Fix payment amount/fee for overpaid non-amountless chain swap

dangeross:
For an overpaid non-amountless chain swap we are displaying the following when sending 59999sat instead of 25000sat:
```
BreezSDKLiquid.Payment(
    timestamp: 1741719704, 
    amountSat: 24173, 
    feesSat: 35826, 
    ...
```
In this case it's taking the actual payer amount (59999) and receiver amount (24173) and calculating the fee as 59999 - 24173 = 35826. We should only take the actual payer amount if it's an amountless swap, as any other paid amount results in a refund.

**Testing notes:**
- When receiving a fixed BTC amount and overpaying it, the amount and fee in the payment item should remain the same when refundable

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -656,20 +656,26 @@ impl Persister {
                 }
                 None => match maybe_chain_swap_id {
                     Some(chain_swap_id) => {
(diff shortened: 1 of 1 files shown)


danielgranhao (approved):
LGTM

roeierez (approved):
LGTM

PR #775, opened 2025-03-13
Fix swap recovery before wallet scan

danielgranhao:
This PR fixes an issue where swaps would be incorrectly recovered due to the local wallet tx map missing some txs. 

On a fresh instance that begins recovery, there is no cached derivation index, causing the wallet full scan to be done to index 0. We didn't notice this before because it requires having 20 consecutive addresses unused.

The fix proposed here is to start by applying the derivation index records, then do a wallet full scan, and only afterward try to process swap records. This does add an additional delay to the recovery.


lib/core/src/recover/recoverer.rs
```diff
@@ -142,7 +142,7 @@ impl Recoverer {
             .ok_or_else(|| anyhow!("Found no input for claim tx"))?;
 
         let script_witness_bytes = input.clone().witness.script_witness;
-        log::info!("Found Send Swap {swap_id} claim tx witness: {script_witness_bytes:?}");
+        info!("Found Send Swap {swap_id} claim tx witness: {script_witness_bytes:?}");
```
roeierez, line 145: It seems that If we can compare here the cached index with the wallet index then we can run a scan here and we won't need the full scan in the recovery

lib/core/src/sync/mod.rs
```diff
@@ -414,33 +418,45 @@ impl SyncService {
...
+
+        // Step 5: Apply derivation index updates and sync wallet (step 6 requires an updated wallet)
+        for decryption_info in decrypted_last_derivation_index_info {
+            if let Err(e) = self.commit_record(&decryption_info, None) {
+                warn!("Could not commit last derivation index record: {e:?}");
+                continue;
+            }
+            succeded.push(decryption_info.record.id);
+        }
+        self.onchain_wallet.full_scan().await?;
```
dangeross, line 439: Maybe check there is a derivation index update before triggering, otherwise we do it every pull. Could also check there are swaps to recover, else there is no point scanning

danielgranhao: Good point. Done in [932bf1f](https://github.com/breez/breez-sdk-liquid/pull/775/commits/932bf1fb2875a9d0894c66daf547904f3c587bc2)

danielgranhao: But this still doesn't prevent the initial pull on a fresh instance from being significantly delayed. But I'm not sure we can get around doing the full scan, at least for as long as we continue to depend on the wallet tx map.

dangeross: This only effects an edge case though right, where the first batch of 20 addresses are unused?

I imagine if there is any large gap in the addresses and the device comes online after a while, there will be a delay until we sync the up-to-date derivation index

roeierez: I see now that I my comment above should have been located here.
I meant that if we are able to compare the derivation index to the one already scanned by the wallet at this point then we can trigger a full scan and remove the added full scan during sync.

danielgranhao: Thanks @roeierez. Indeed having the scan in `recover_from_onchain` prevents an unnecessary additional scan. Refactored in [0687df9](https://github.com/breez/breez-sdk-liquid/pull/775/commits/0687df9acc19c44eba588cce64e409f632e32f51)

roeierez (approved):
LGTM

lib/core/src/wallet.rs
```diff
@@ -407,6 +407,9 @@ impl OnchainWallet for LiquidOnchainWallet {
                 }
                 Err(e) => Err(e.into()),
             };
+
+        self.persister.set_last_scanned_derivation_index(index)?;
```
dangeross, line 411: This is going to store last_derivation_index + 5, meaning if the data-sync receives an increased index by 1-5 it won't force the sync

danielgranhao: Yes, but I think that is fine. The stored index is the one we did end up scanning to. Do you see this causing any issues?

dangeross: My point is an incremented derivation index received by the data-sync will not trigger a full scan. Which I think is half the reason for the PR, no?

roeierez: @danielgranhao I think what Ross means is that we should save the index we get from the persister here: https://github.com/breez/breez-sdk-liquid/pull/775/files#diff-6b184d7d7439883bb4f686575e75810a483be9041940320c97a094c81f0b077cR386
and not the one we increment by 5 (to scan a bit forward). Because eventually we compare the persister value for the decision to scan: https://github.com/breez/breez-sdk-liquid/pull/775/files#diff-28794968ead61e86665263d494740e1a8fbe62fcfed42ed568bcf77fd51ca31dR438

danielgranhao: I've changed the persisted index to not include the buffer in [de5bbf0](https://github.com/breez/breez-sdk-liquid/pull/775/commits/de5bbf05eefb965ad9c622ba03f0bdb159b2c2e7). Let's be on the safe side.

Nackoo2000:
verified

PR #776, opened 2025-03-14
Fix LNURL drain and payment info

dangeross:
This PR fixes a couple of LNURL related issues:
- Fixes a regression where LNURL-pay drain no longer worked (failed in the prepare step)
- Adds the LNURL info and BIP353 address to Liquid payment details so they are also available when paying via MRH

See https://github.com/breez/misty-breez/issues/422

**Testing notes:**
- LNURL-pay drain should work
- LNURL info should be displayed in the payment when paying via LNURL-pay / lightning address to another Misty/SDK app

**Not-testable testing notes:**
- BIP353 address (as Lightning Address) should be displayed in the payment when paying to another Misty/SDK app and the BIP353 address links to a LNURL-pay.

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -3479,9 +3479,8 @@ impl LiquidSdk {
                             bip353_address: req.bip353_address,
                         }
                     } else {
-                        return Err(LnUrlPayError::Generic {
-                            err: "SendDestination for LNURL Pay is not BOLT11 invoice".to_string(),
-                        });
+                        // Destination is a Liquid address for the drain request
+                        prepare_response.destination
                     };
 
                 Ok(PrepareLnUrlPayResponse {
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -1457,7 +1457,7 @@ impl LiquidSdk {
                 invoice,
                 bip353_address,
             } => {
-                let response = self.pay_bolt11_invoice(&invoice.bolt11, *fees_sat).await?;
+                let mut response = self.pay_bolt11_invoice(&invoice.bolt11, *fees_sat).await?;
                 if bip353_address.is_some() {
                     if let (Some(tx_id), Some(destination)) =
                         (&response.payment.tx_id, &response.payment.destination)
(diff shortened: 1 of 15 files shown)


danielgranhao (approved):
LGTM

lib/core/src/sdk.rs
```diff
@@ -3479,9 +3489,8 @@ impl LiquidSdk {
                             bip353_address: req.bip353_address,
                         }
                     } else {
-                        return Err(LnUrlPayError::Generic {
-                            err: "SendDestination for LNURL Pay is not BOLT11 invoice".to_string(),
-                        });
+                        // Destination is a Liquid address for the drain request
+                        prepare_response.destination
```
erdemyerebasmaz, line 3493: Wouldn't it be more robust to use pattern matching on `prepare_response.destination`?

```
let destination = match prepare_response.destination {
    SendDestination::Bolt11 { invoice, .. } => ...,
    SendDestination::LiquidAddress { .. } => ...,
    ...
```

erdemyerebasmaz: Does this also imply we should add the optional `bip353_address` to `SendDestination::LiquidAddress` for LNURL-pay drain's via BIP 353?

dangeross: Good catch, added [c4f1559](https://github.com/breez/breez-sdk-liquid/pull/776/commits/c4f15590e8a2d7b5f12551155359c63883d5a0e2)

erdemyerebasmaz (approved):
LGTM

danielgranhao (approved):

Nackoo2000:
The issue above happens when trying to empty an iOS device to an LNURL on an Android device.
But when I tried the reverse on those same devices it worked - Android to iOS
[l-breez.logs (71).zip](https://github.com/user-attachments/files/19314708/l-breez.logs.71.zip)

dangeross:
@Nackoo2000 Only the Android build in [this comment](https://github.com/breez/misty-breez/pull/426#issuecomment-2729732185) included the fix. So it is expected. I can make an iOS build for you also if you want

Nackoo2000:
Yes please make iOS builds every time because I am testing both

PR #790, opened 2025-03-19
WASM: sync grpc client

danielgranhao:
This PR mimics the approach from https://github.com/breez/breez-sdk-greenlight/pull/1170 for using `tonic` on WASM. That is:

* Use `tonic_web_wasm_client` on wasm builds
* Do not build tonic transport features on wasm builds

It depends on https://github.com/breez/breez-sdk-greenlight/pull/1183

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -349,7 +349,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3b829e4e32b91e643de6eafe82b1d90675f5874230191a4ffbc1b336dec4d6bf"
 dependencies = [
  "async-trait",
- "axum-core 0.3.4",
+ "axum-core",
  "bitflags 1.3.2",
  "bytes",
  "futures-util",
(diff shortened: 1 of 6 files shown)


dangeross (commented):
Is it possible to expose either GrpcClient/Transport from sdk_common?

danielgranhao:
> Is it possible to expose either GrpcClient/Transport from sdk_common?

Good point. I'll do that

dangeross (approved):
LGTM

roeierez (approved):
Nice!

danielgranhao:
@dangeross @roeierez Thanks for the review. Triggered new review requests because I had to downgrade tonic and prost in order to avoid compatibility issues with `sdk-common`. Unfortunately, for now, `sdk-common` is stuck on tonic `0.8` due to gl-client. I'll see how easy it is to upgrade tonic on gl-client, but in the meantime, I think downgrading here should be fine.

danielgranhao:
Ok, I had a look, and it isn't so easy to upgrade tonic on gl-client. It is also locked into 0.8 due to `cln-grpc`, which has later releases with more recent versions of tonic but with API changes.

roeierez (approved):

dangeross (approved):

hydra-yse (approved):

PR #791, opened 2025-03-19
WASM: lock frb and bindings modules

danielgranhao:
This locks out the flutter-specific modules on WASM builds


dangeross (commented):
I'm fine with this approach. Maybe in a follow up PR it would make sense to remove `frb` from the default feature and only add it when generating the Flutter bindings. We just need to check if this feature also needs to be active when building the android/swift libraries for Flutter.

dangeross (approved):

PR #792, opened 2025-03-19
WASM: prevent filesystem access

danielgranhao:
This PR skips all fs access on WASM builds. We may later want to use [`FileSystem`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.FileSystem.html), but as a start, we should be fine without any persistence.

Additionally, it also proposes exposing `LiquidSdkBuilder` and starting to use it on the wasm crate to provide wasm-compatible components. This is applied to the `OnchainWallet` implementation, which on wasm must only cache data in memory. This approach helps minimize the use of conditional compilation within our core logic.


lib/core/src/logger.rs
```diff
@@ -34,40 +32,48 @@ impl log::Log for GlobalSdkLogger {
 }
 
 pub(super) fn init_logging(log_dir: &str, app_logger: Option<Box<dyn log::Log>>) -> Result<()> {
-    let target_log_file = Box::new(
```
roeierez, line 37: As far as I understand this method was only created to easy the usage for rust callers. It is not exposed to the bindings and we only use it from the cli.
Perhaps then it make sense to exclude it completely for wasm?
If we think later we need something specific for wasm perhaps we can add it in the wasm crate?

roeierez (changes requested):

lib/core/src/logger.rs
danielgranhao: Thanks. I didn't have in mind that it's only used on the rust interface. I completely excluded it for wasm in [101751f](https://github.com/breez/breez-sdk-liquid/pull/792/commits/101751fbd4eb19702ac68fa3426419ccf547565e)

lib/core/src/test_utils/persist.rs
```diff
@@ -123,12 +123,26 @@ pub(crate) fn new_receive_swap(
...
+            let res: String = rand::thread_rng()
+                .sample_iter(&rand::distributions::Alphanumeric)
+                .take(16)
+                .map(char::from)
+                .collect();
+            res
+        };
+
+        #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+        let temp_dir_path = tempdir::TempDir::new("liquid-sdk")?
```
roeierez, line 138: @danielgranhao currently we are including/excluding wasm related code using macros. I think on cases where we have a clear trait and implementation we can utilize the approach of passing the implementation from outside instead.
We already having the LiquidSdkBuilder which builds the sdk instance and has the ability to config services such persister, onchain wallet, etc..
We can implement the connect there and even if it means some code duplication it might be a cleaner approach.
LMK what you think.

lib/core/src/sdk.rs
```diff
@@ -142,6 +142,18 @@ impl LiquidSdkBuilder {
         self
     }
 
+    pub fn default_persister(&self, sync_enabled: bool) -> Result<Arc<Persister>> {
```
roeierez, line 145: Perhaps it makes sense to move this function to the Persister struct?
We can have two methods:
```
new_fs_persister(workingDir, syncEnabled)
new_memory_persisiter(syncEnabled)
```

It will also allow us to move the calls of `get_working_dir` and `std::fs::create_dir_all(&main_db_dir)` to the corresponding function and exclude/include the whole functions based on wasm build or not.

lib/core/src/sdk.rs
```diff
@@ -162,17 +174,22 @@ impl LiquidSdkBuilder {
...
+    }
+
+    pub(crate) fn build(&self) -> Result<Arc<LiquidSdk>> {
         if let Some(breez_api_key) = &self.config.breez_api_key {
             LiquidSdk::validate_breez_api_key(breez_api_key)?
         }
 
-        fs::create_dir_all(&self.config.working_dir)?;
+        #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+        std::fs::create_dir_all(&self.config.working_dir)?;
```
roeierez, line 190: I think it would also be nice to move this code to an upper layer. Perhaps in the connect_with_signer ?

roeierez (changes requested):
I very like this approach! Dropped a few comments.

lib/core/src/sdk.rs
danielgranhao: Done in [7e85942](https://github.com/breez/breez-sdk-liquid/pull/792/commits/7e859427927bdbb2c9c6ee94531af9c855ff3e31). I also moved the log print header as it was not being called when integrators provided their own signer.

lib/core/src/sdk.rs
danielgranhao: I wasn't keen on having a specific in-memory constructor for the persister because of us not having control over when it's in-memory or not, but excluding the constructors on the appropriate environment solves that issue, thanks!

I applied this on [ac846b4](https://github.com/breez/breez-sdk-liquid/pull/792/commits/ac846b4d4f84baa214cb4a66bb085bfd5b2f8dc2). Still kept using the path for the in-memory version because `sqlite-wasm-rs` allows using the same in-memory db if we provide the same path. Not sure if we have an immediate need for this, but it's there.

lib/core/src/sdk.rs
```diff
@@ -345,6 +337,12 @@ impl LiquidSdkBuilder {
         });
         Ok(sdk)
     }
+
+    pub async fn connect(&self) -> Result<Arc<LiquidSdk>> {
+        let sdk = self.build()?;
+        sdk.start().await?;
```
roeierez, line 343: What is the reason for moving the start here? I prefer to keep the LiquidSdkBuilder as a builder for sdk object.

roeierez (changes requested):
One last NIT, otherwise LGTM.

lib/core/src/sdk.rs
danielgranhao: Fixed in [2971128](https://github.com/breez/breez-sdk-liquid/pull/792/commits/2971128c0eb7673c11ff8b322a8b9e2e130567ab)

roeierez (approved):
LGTM!

lib/core/src/sdk.rs
```diff
@@ -181,25 +184,14 @@ impl LiquidSdkBuilder {
...
-                let persister = Arc::new(Persister::new(
-                    &working_dir,
-                    self.config.network,
-                    sync_enabled,
-                )?);
-                persister.init()?;
-                persister.replace_asset_metadata(self.config.asset_metadata.clone())?;
-                persister
-            }
+            None => Arc::new(Persister::new_using_fs(
```
dangeross, line 189: Won't using `Persister::new_using_fs()` cause compilation issues in WASM?

lib/core/src/sdk.rs
```diff
@@ -3304,13 +3301,14 @@ impl LiquidSdk {
     }
 
     /// Empties the Liquid Wallet cache for the [Config::network].
+    #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
```
dangeross, line 3304: Does this need removing from the wasm crate interface also?

lib/core/src/sdk.rs
danielgranhao: Yes, it isn't even included in wasm builds. The idea is that wasm builds will always use `new_in_memory()`.

lib/core/src/sdk.rs
danielgranhao: Nice catch! Removed in [bbde082](https://github.com/breez/breez-sdk-liquid/pull/792/commits/bbde082f304a4da3034749c6ab1091011992e0e1).

lib/core/src/sdk.rs
dangeross: How is this going to compile then if wasm calls the `build()` fn?

danielgranhao: You're right. I'll fix it later today or tomorrow. Thanks for spotting that. There might be some more similar issues with WASM builds on these PRs I've prepared. WASM can only be built when the esplora PR and some of them are put together, so I haven't built them and was aware I could be missing things like this.

As long as native builds are good, we should be fine for now with potential WASM issues, which we'll fix once we have all PRs merged.

danielgranhao: I added conditional compilation imposing the use of a custom persister on WASM builds to address the issue.

lib/core/src/sdk.rs
```diff
@@ -181,24 +184,19 @@ impl LiquidSdkBuilder {
...
+                if cfg!(not(all(target_family = "wasm", target_os = "unknown"))) {
+                    Arc::new(Persister::new_using_fs(
+                        &self.get_working_dir()?,
+                        self.config.network,
+                        self.config.sync_enabled(),
+                        self.config.asset_metadata.clone(),
+                    )?)
+                } else {
+                    panic!("Must provide a WASM-compatible persister on WASM builds")
+                }
```
dangeross, line 199: I don't think this will compile as the `cfg!` macro does not conditionally compile blocks/statements. Maybe:

```suggestion
                #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
                Arc::new(Persister::new_using_fs(
                    &self.get_working_dir()?,
                    self.config.network,
                    self.config.sync_enabled(),
                    self.config.asset_metadata.clone(),
                )?)
                #[cfg(all(target_family = "wasm", target_os = "unknown"))]
                return Err(anyhow!("Must provide a WASM-compatible persister on WASM builds"));
```

danielgranhao: Thanks. Applied in [074b2ec](https://github.com/breez/breez-sdk-liquid/pull/792/commits/074b2ec2ac5070bd5368ab209a9cc721067c0283)

dangeross (approved):
LGTM

PR #795, opened 2025-03-20
WASM: getting the current time

danielgranhao:
This PR introduces the `web-time` crate for getting the current time in a wasm-compatible way. It's essentially a wrapper that on native builds just re-exports `std::time::{Instant, SystemTime}` but on wasm replaces those with implementations that rely on browser APIs.

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -764,6 +764,7 @@ dependencies = [
  "tonic-build 0.12.3",
  "url",
  "uuid",
+ "web-time",
  "x509-parser",
  "zbase32",
 ]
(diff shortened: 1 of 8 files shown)


dangeross (approved):
Looks good. Do we know what kind of effect [this might](https://docs.rs/web-time/latest/web_time/#ticking-during-sleep) have? I guess we won't really know until we start testing

danielgranhao:
> Looks good. Do we know what kind of effect [this might](https://docs.rs/web-time/latest/web_time/#ticking-during-sleep) have? I guess we won't really know until we start testing

Indeed we'll only know better once we start testing, but at least for now, we only use Instant for logging execution times, so even if it isn't reliable on the browser, it isn't critical.

roeierez (approved):
LGTM

PR #796, opened 2025-03-20
WASM: async sleep

danielgranhao:
This PR proposes using [`setTimeout()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout) to implement async sleep on WASM builds.

This approach does have issues, such as [this](https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#timeouts_in_inactive_tabs), and serves just a first step. Once we have a WASM-compatible build, we'll need to test it, and we'll figure out then if we to improve it.

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -736,6 +736,7 @@ dependencies = [
  "futures-util",
  "glob",
  "hex",
+ "js-sys",
  "lazy_static",
  "log",
  "lwk_common",
@@ -764,6 +765,8 @@ dependencies = [
  "tonic-build 0.12.3",
  "url",
  "uuid",
+ "wasm-bindgen-futures",
+ "web-sys",
  "x509-parser",
  "zbase32",
 ]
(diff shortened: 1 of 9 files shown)


dangeross (commented):
Did you consider using `tokio_with_wasm` for your timeout and spawn PRs?

roeierez:
> Did you consider using `tokio_with_wasm` for your timeout and spawn PRs?

According to the docs tokio time feature is supported in wasm and it has `tokio::time::sleep` function.

danielgranhao:
> > Did you consider using `tokio_with_wasm` for your timeout and spawn PRs?
> 
> According to the docs tokio time feature is supported in wasm and it has `tokio::time::sleep` function. https://docs.rs/tokio/latest/tokio/#wasm-support

@roeierez Great catch! I could've sworn I did see tokio's sleep causing issues when doing some experiments on WASM, but I just tried it out and it does indeed work. Most likely I was influenced by the solution I saw on lwk_wollet 🤦 

Closing this as for now it looks completely unnecessary. We'll figure out reliability of the sleeps once we start testing.

PR #797, opened 2025-03-21
WASM: spawn task

danielgranhao:
This PR proposes using [`tokio_with_wasm`](https://crates.io/crates/tokio_with_wasm) to spawn async tasks on wasm. 

Something to keep in mind is https://crates.io/crates/tokio_with_wasm#caution. The most obvious and immediate effect is that our tests that rely on spawning a task ( the update tracking tests on `sdk.rs`), if they fail on WASM, they do with a very non-descriptive error. A failing assert in just one of the tests causes the following output. We don't even get to see which test failed. We might deem this fine if we think it's unlikely for wasm tests to fail differently from native tests.

```
Failed to detect test as having been run. It might have timed out.
output div contained:
    running 38 tests
    test sdk::tests::test_zero_amount_chain_swap_with_leeway ... ok
    test sdk::tests::test_zero_amount_chain_swap_zero_leeway ... ok

driver status: signal: 9 (SIGKILL)                
driver stdout:
    1742562140558       geckodriver     INFO    Listening on 127.0.0.1:56493
    1742562140658       mozrunner::runner       INFO    Running command: MOZ_CRASHREPORTER="1" MOZ_CRASHREPORTER_NO_REPORT="1" MOZ_CRASHREPORTER_SHUTDOWN="1" "/Applications/Firefox. ... e" "-headless" "-foreground" "-no-remote" "-profile" "/var/folders/j7/9m9jk4bs61s_r3vcrdtkjdyh0000gn/T/rust_mozprofileDqrTcJ"
    console.warn: services.settings: Ignoring preference override of remote settings server
    console.warn: services.settings: Allow by setting MOZ_REMOTE_SETTINGS_DEVTOOLS=1 in the environment
    1742562143312       Marionette      INFO    Marionette enabled
    1742562144188       Marionette      INFO    Listening on port 56503
    [GFX1-]: RenderCompositorSWGL failed mapping default framebuffer, no dt
    1742562166332       Marionette      INFO    Stopped listening on port 56503

driver stderr:
    *** You are running in headless mode.
    JavaScript error: http://127.0.0.1:56492/wasm-bindgen-test_bg.wasm, line 27874641: RuntimeError: unreachable executed

```

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -760,6 +760,7 @@ dependencies = [
  "thiserror 1.0.69",
  "tokio",
  "tokio-stream",
+ "tokio_with_wasm",
  "tonic 0.12.3",
  "tonic-build 0.12.3",
  "url",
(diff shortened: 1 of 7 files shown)


danielgranhao:
@dangeross Thanks for proposing [`tokio_with_wasm`](https://crates.io/crates/tokio_with_wasm). I'll update the PR.

hydra-yse:
I agree with the choice here. Given the negatives I suggest we postpone this to after we've successfully debugged any testing errors

danielgranhao:
> I agree with the choice here. Given the negatives I suggest we postpone this to after we've successfully debugged any testing errors

Maybe I missed your point but we'll only be able to test WASM once we have this. I don't see the value in postponing. If we start having issues with tests failing just on WASM, then we'll probably need to refactor the test to avoid needing to spawn tasks.

roeierez (approved):
LGTM

Issue #742, opened 2025-02-14
Persist LNURL metadata when receiving a payment via Notification Extension


erdemyerebasmaz:
The metadata we pass to `ReceivePaymentRequest` as `description` is not being persisted when we create & update a `ReceiveSwap`.

**Non-verified, possible culprits:**
- [ ] The `description` we use on `CreateReverseRequest` is not reflected on a `SignedRawBolt11Invoice`

erdemyerebasmaz:
> [@erdemyerebasmaz](https://github.com/erdemyerebasmaz) what exactly the user issue here?

@roeierez As far as I can recall, the optional comments(notes) the sender is setting are not reflected on the payment item.

PR #798, opened 2025-03-21
Store a clone of the description before moving it in CreateReverseReq…

vkprogrammer-001:
Fix: #742

hey @erdemyerebasmaz. i have solved this issue. can you please review this PR.


dangeross:
Thanks for the PR @vkprogrammer-001! Can we test that the description is set correctly in the created invoice? If not we should raise an issue with Boltz

danielgranhao:
Thanks @vkprogrammer-001. I think what @dangeross meant is that we should check why we're not getting the correct invoice in the response returned by Boltz. If Boltz is not setting the description we provide, we should open an issue there. Your fix does address the persisted description but the created invoice could still have an empty description.

vkprogrammer-001:
Thanks for the feedback, @danielgranhao! How we can check why we're not getting the correct invoice in the response returned by Boltz?

PR #799, opened 2025-03-21
WASM: run core crate tests on wasm

danielgranhao:
This adds a makefile to the `core` crate with helper targets for running clippy and tests. We already have something similar on the wasm crate, but running tests there won't run the tests in the core crate. I also adjusted the main makefile to run wasm clippy and tests on both crates.


roeierez (approved):
LGTM

dangeross (approved):
LGTM

PR #800, opened 2025-03-21
Expose real time synced event

### LINKED ISSUE #778: Expose in-progress sync
At the start of an instance that needs to pull remote records, it may be difficult for integrators to communicate to users that some background job is happening, only after which the payments list will be up-to-date.

I want to open the discussion for exposing info about these background jobs, at least their first run. 

Maybe it's enough to add a new event `DataSynced` or `RealtimeDataSynced` (like the existing `Synced`) in `SdkEvent` that would be emitted when `SyncService::run_event_loop()` finishes. Then integrators could show some loading indicator on startup until they get the first instance of this event.

danielgranhao:
Resolves #778.

This exposes a new event for when a real-time data sync occurs. The primary motivation is to allow showing users some visual indication on startup that this sync is occurring (so payments might be missing). Integrators know when the rt sync is complete through the new event. Also included in the event indication if new data was pulled or not, which may allow integrators to provide users with feedback that data is coming in from other devices.

Misc: renamed the existing `Synced` event to `ChainSynced` to allow better distinction between both synced events.

### DIFF
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -690,6 +690,7 @@ interface SdkEvent {
     PaymentWaitingConfirmation(Payment details);
     PaymentWaitingFeeAcceptance(Payment details);
     Synced();
+    DataSynced(boolean did_pull_new_records);
 };
 
 callback interface EventListener {    
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -295,15 +295,37 @@ pub trait EventListener: MaybeSend + MaybeSync {
 /// to listen for emitted events.
 #[derive(Clone, Debug, PartialEq)]
 pub enum SdkEvent {
-    PaymentFailed { details: Payment },
-    PaymentPending { details: Payment },
-    PaymentRefundable { details: Payment },
-    PaymentRefunded { details: Payment },
-    PaymentRefundPending { details: Payment },
-    PaymentSucceeded { details: Payment },
-    PaymentWaitingConfirmation { details: Payment },
-    PaymentWaitingFeeAcceptance { details: Payment },
+    PaymentFailed {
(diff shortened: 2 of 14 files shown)


lib/core/src/model.rs
```diff
@@ -295,15 +295,37 @@ pub trait EventListener: MaybeSend + MaybeSync {
...
+    PaymentWaitingConfirmation {
+        details: Payment,
+    },
+    PaymentWaitingFeeAcceptance {
+        details: Payment,
+    },
+    /// Synced with mempool and onchain data
+    ChainSynced,
+    /// Synced with real-time data sync
+    RtDataSynced {
```
dangeross, line 325: I'd rather call it `RealTimeDataSynced` or simply `DataSynced` than have to explain what Rt means

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -689,7 +689,8 @@ interface SdkEvent {
     PaymentSucceeded(Payment details);
     PaymentWaitingConfirmation(Payment details);
     PaymentWaitingFeeAcceptance(Payment details);
-    Synced();
+    ChainSynced();
```
roeierez, line 692: NIT: Are we sure we want to rename it? If so shouldn't we prefer something more abstract that will not lock us for the chain semantic? Perhaps WalletSynced? TBH I am not sure we need to rename it.

roeierez (commented):
Looks good! Only one concern regarding the rename.

lib/core/src/model.rs
danielgranhao: Renamed to `DataSynced` in [207b692](https://github.com/breez/breez-sdk-liquid/pull/800/commits/207b692ef79e3d4080f8eca4da94d12a24498144)

dangeross (approved):

Issue #793, opened 2025-03-19
Drop `frb` feature from default features


danielgranhao:
Right now, the `frb` feature is a default feature. We should remove it from the default features and only enable it when necessary (i.e. building flutter bindings).

vkprogrammer-001:
hey @danielgranhao. can I solve this issue, please?

rajanarahul93:
hey @danielgranhao , as the issue is not solved yet can I solve it?

Bansalayush247:
@danielgranhao 

## Follow-up on removing `frb` from default features (Issue #793)

I've made the changes to remove `frb` from default features in the core `Cargo.toml` file:

```toml
[features]
default = []
frb = ["dep:flutter_rust_bridge"]

However, I'm encountering a build error in the WASM module:

error[E0599]: no function or associated item named `new_in_memory` found for struct `breez_sdk_liquid::persist::Persister` in the current scope
  --> wasm/src/lib.rs:53:41
   |
53 |     let persister = Arc::new(Persister::new_in_memory(
   |                                         ^^^^^^^^^^^^^ function or associated item not found in `Persister`

It appears that the WASM module is trying to use functionality that is only available when the frb feature is enabled.

Question
Does the WASM build inherently require the frb feature? If yes, we should explicitly enable it for the WASM module like this:

# In wasm/Cargo.toml
[dependencies]
breez-sdk-liquid = { path = "../core", features = ["frb"] }

Or should we modify the WASM code to work without the frb feature?

I'd appreciate guidance on the intended approach for the WASM build in relation to Flutter Rust Bridge functionality.

PR #801, opened 2025-03-24
Remove frb from default features in Cargo.toml #793

rajanarahul93:
#### Changes:
- Removed `frb` from the `default` feature list in `Cargo.toml`.
- Ensured that `frb` remains optional by keeping it under `[features]`.

#### Reason:
- `frb` should not be enabled by default to provide more flexibility in feature selection.
- This change allows users to explicitly enable `frb` only when needed.

#### Impact:
- The build process will no longer include `flutter_rust_bridge` by default.
- Users who require `frb` need to enable it manually.

solves #793 
@danielgranhao

### DIFF
--- a/lib/core/Cargo.toml
+++ b/lib/core/Cargo.toml
@@ -8,8 +8,8 @@ name = "breez_sdk_liquid"
 crate-type = ["lib", "cdylib", "staticlib"]
 
 [features]
-default = ["frb"]
-frb = ["dep:flutter_rust_bridge"]
+default = []  # Removed "frb" from default features
+frb = ["dep:flutter_rust_bridge"]  # Kept as an optional feature
 # Uniffi features required to build using cargo-lipo
 uniffi-25 = []
 uniffi-28 = []


danielgranhao:
Also, please keep in mind that building flutter bindings is also failing due to a Rust upgrade. Fix is in https://github.com/breez/breez-sdk-liquid/pull/803.

danielgranhao:
But your PR is related to flutter. `frb` stands for "flutter rust bridge," and removing this feature from the default features is expected to cause binding issues as the build will need to enable the feature. As I wrote, flutter is failing anyway due to an issue fixed in #803. Maybe it's best for you to base your PR on top of that one so you can see where any issue could be.

rajanarahul93:
Hi @danielgranhao,  
Since #803  is not merged yet, should I wait for it to be merged before proceeding with my PR,  
or should I temporarily rebase on top of #803 's branch to test the Flutter bindings?  
Thanks!

PR #805, opened 2025-03-24
Integrate Esplora support to Boltz swapper

hydra-yse:


lib/core/src/swapper/boltz/client.rs
```diff
@@ -0,0 +1,142 @@
...
+                    config.network.into(),
+                    url,
+                    BOLTZ_CONNECTION_TIMEOUT as u64,
+                )))
+            }
+        })
+    }
+}
+
+#[async_trait]
```
dangeross, line 47: To make it wasm friendly
```suggestion
#[sdk_macros::async_trait]
```

lib/core/src/swapper/boltz/client.rs
```diff
@@ -0,0 +1,142 @@
...
+                    config.network.as_bitcoin_chain(),
+                    url,
+                    BOLTZ_CONNECTION_TIMEOUT as u64,
+                )))
+            }
+        })
+    }
+}
+
+#[async_trait]
```
dangeross, line 110:
```suggestion
#[sdk_macros::async_trait]
```

dangeross (commented):
Looking good

lib/core/src/swapper/boltz/client.rs
hydra-yse: It's using the `sdk_macros` trait, just implicitly. I'll set it to explicit

dangeross: Ah sorry, I missed it 🙃

dangeross (approved):
LGTM

lib/core/src/swapper/boltz/mod.rs
```diff
@@ -36,41 +38,24 @@ pub(crate) struct BoltzClient {
 pub struct BoltzSwapper<P: ProxyUrlFetcher> {
     config: Config,
     client: OnceLock<BoltzClient>,
-    liquid_electrum_client: ElectrumLiquidClient,
-    bitcoin_electrum_client: ElectrumBitcoinClient,
+    liquid_electrum_client: LiquidClient,
+    bitcoin_electrum_client: BitcoinClient,
```
danielgranhao, line 42:
```suggestion
    liquid_client: LiquidClient,
    bitcoin_client: BitcoinClient,
```

danielgranhao (commented):
I see the issue you mentioned earlier. It's annoying to have to reimplement the boltz traits but at least it solves the issue. Looks good! Just missed renaming these fields

lib/core/src/swapper/boltz/client.rs
```diff
@@ -0,0 +1,142 @@
...
+
+const BOLTZ_CONNECTION_TIMEOUT: u8 = 100;
+
+pub(crate) enum LiquidClient {
+    Electrum(Box<ElectrumLiquidClient>),
+    Esplora(Box<EsploraLiquidClient>),
+}
+
+impl LiquidClient {
+    pub(crate) fn new(config: &Config) -> Result<Self, Error> {
```
roeierez, line 24: Did you consider using the public implementations that already exist in boltz? https://github.com/SatoshiPortal/boltz-rust/blob/trunk/src/network/esplora.rs#L20

roeierez (changes requested):

lib/core/src/swapper/boltz/client.rs
hydra-yse: Yes, it cannot be done as we're matching for the client type at runtime and the library uses static dispatch for its trait coercion

lib/core/src/swapper/boltz/mod.rs
hydra-yse: Thanks :smile:

roeierez (approved):

danielgranhao:
Ah, for wasm to build, we'll need to exclude the electrum client options, right?

lib/core/src/model.rs
```diff
@@ -112,12 +113,25 @@ pub struct Config {
...
+
+            #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
             bitcoin_explorer: BlockchainExplorer::Electrum {
                 url: "bitcoin-mainnet.blockstream.info:50002".to_string(),
             },
+            #[cfg(all(target_family = "wasm", target_os = "unknown"))]
+            bitcoin_explorer: BlockchainExplorer::Esplora {
+                url: "https://blockstream.info/api/".to_string(),
+            },
+
```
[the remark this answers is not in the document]
hydra-yse: I see, makes sense

lib/core/src/chain/bitcoin/electrum.rs
```diff
@@ -46,6 +46,7 @@ impl ElectrumBitcoinChainService {
             LiquidNetwork::Regtest => (false, false),
         };
         let electrum_url = match &self.config.bitcoin_explorer {
+            #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
```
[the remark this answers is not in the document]
danielgranhao: True. I just noticed we already are excluding the entire file on wasm, so this isn't needed

lib/core/src/model.rs
```diff
@@ -132,6 +130,31 @@ impl Config {
         }
     }
 
+    pub fn mainnet_esplora(breez_api_key: Option<String>) -> Self {
```
dangeross, line 133: The use of `mainnet_esplora()` depends Wasm, so why not switch the function depending on Wasm. Same for the other networks
```suggestion
    #[cfg(all(target_family = "wasm", target_os = "unknown"))]
    pub fn mainnet(breez_api_key: Option<String>) -> Self {
```

lib/core/src/swapper/boltz/mod.rs
```diff
@@ -36,41 +38,24 @@ pub(crate) struct BoltzClient {
 pub struct BoltzSwapper<P: ProxyUrlFetcher> {
     config: Config,
     client: OnceLock<BoltzClient>,
-    liquid_electrum_client: ElectrumLiquidClient,
-    bitcoin_electrum_client: ElectrumBitcoinClient,
+    liquid_client: LiquidClient,
+    bitcoin_electrum_client: BitcoinClient,
```
dangeross, line 42: Needs renaming?
```suggestion
    bitcoin_client: BitcoinClient,
```

dangeross (changes requested):

lib/core/src/sdk.rs
```diff
@@ -3763,11 +3763,19 @@ impl LiquidSdk {
...
             LiquidNetwork::Mainnet => Config::mainnet(breez_api_key),
             LiquidNetwork::Testnet => Config::testnet(breez_api_key),
             LiquidNetwork::Regtest => Config::regtest(),
         };
+        #[cfg(all(target_family = "wasm", target_os = "unknown"))]
+        let config = match network {
+            LiquidNetwork::Mainnet => Config::mainnet_esplora(breez_api_key),
+            LiquidNetwork::Testnet => Config::testnet_esplora(breez_api_key),
+            LiquidNetwork::Regtest => Config::regtest_esplora(),
+        };
```
danielgranhao, line 3777: Another option here is to leave the default_config() as it is, just excluding it on Wasm builds. Then have a separate default_esplora_config() method. We can discuss later if we want to expose the new esplora default method on the bindings, but at least in the rust interface, it could be useful to easily start a node with an esplora backend on the CLI or in end-to-end tests

lib/core/src/model.rs
danielgranhao: I think it may be useful to have an esplora default available on native builds as well, so we could even rename the mainnet to mainnet_electrum. If not for anything else, for the CLI and end-to-end tests. But I'm not strongly opinionated here.

hydra-yse: Sorry for not explaining, it was mostly an online conversation with @danielgranhao. Opening up the mainnet_esplora also allows non-wasm clients to select that branch if they wish to

lib/core/src/swapper/boltz/mod.rs
hydra-yse: This seems like the 3rd time I've renamed that field, wth is wrong with my editor

lib/core/src/model.rs
dangeross: No worries 👍

lib/core/src/sdk.rs
hydra-yse: Agreed, I'll just exclude it on WASM for now

dangeross (commented):
Please check the Test Wasm errors: https://github.com/breez/breez-sdk-liquid/actions/runs/14105504544/job/39511043237

The test_utils need adapting
```
error[E0599]: no function or associated item named `testnet` found for struct `model::Config` in the current scope
   --> core/src/test_utils/send_swap.rs:20:26
    |
20  |     let config = Config::testnet(None);
    |                          ^^^^^^^ function or associated item not found in `Config`
```

hydra-yse:
@dangeross addressed in #810 :wink:

dangeross (approved):

danielgranhao (approved):

PR #806, opened 2025-03-24
WASM: Conditionally add MissedTickBehavior

dangeross:
This PR:
- Fixes the proto generation for WASM
- Adds `sync` and `time` features to `tokio_with_wasm`
- Conditionally sets the `MissedTickBehavior` for non-WASM intervals

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -27,8 +27,6 @@ use sdk_common::prelude::{FiatAPI, FiatCurrency, LnUrlPayError, LnUrlWithdrawErr
 use signer::SdkSigner;
 use swapper::boltz::proxy::BoltzProxyFetcher;
 use tokio::sync::{watch, RwLock};
-#[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
-use tokio::time::MissedTickBehavior;
 use tokio_stream::wrappers::BroadcastStream;
 use tokio_with_wasm::alias as tokio;
 use web_time::Instant;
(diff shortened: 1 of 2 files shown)


lib/core/src/sdk.rs
```diff
@@ -556,6 +557,7 @@ impl LiquidSdk {
             let mut current_bitcoin_block: u32 = 0;
             let mut shutdown_receiver = cloned.shutdown_receiver.clone();
             let mut interval = tokio::time::interval(Duration::from_secs(10));
+            #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
             interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
```
danielgranhao, line 561: Is it not possible to use the `interval` from the vanilla tokio so that we can set the missed tick behavior?

dangeross: The `Interval` in this case is specific to WASM, using javascripts setTimeout/clearTimeout internally

danielgranhao, line 561: If we directly reference the type we can get rid of the conditional on the top

```suggestion
            interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
```

dangeross: Fixed [ca33d13](https://github.com/breez/breez-sdk-liquid/pull/806/commits/ca33d131e64ecba5db6338aa16aa8a062dd9ec28)

PR #816, opened 2025-03-26
Add regtest end-to-end tests

### LINKED ISSUE #635: Regtest end-to-end test suite

danielgranhao:
Works towards #635

This PR:

- Adds an initial set of regtest tests that covers the different kinds of payments the SDK supports
- Adds a proxy to the regtest setup to allow control of the nodes when running the tests on Wasm (full wasm support is out of scope for this PR). The proxy is useful to circumvent issues caused by browser restrictions on TLS certificates and CORS policies.
- Sets up CI jobs to run the tests separately

### DIFF
--- a/lib/core/Makefile
+++ b/lib/core/Makefile
@@ -4,6 +4,10 @@ ifeq ($(UNAME), Darwin)
 	CLANG_PREFIX += AR=$(shell brew --prefix llvm)/bin/llvm-ar CC=$(shell brew --prefix llvm)/bin/clang
 endif
 
+LND_MACAROON_HEX=$(shell xxd -p ../../regtest/boltz/data/lnd1/data/chain/bitcoin/regtest/admin.macaroon | tr -d '\n')
+BITCOIND_COOKIE=$(shell cat ../../regtest/boltz/data/bitcoind/regtest/.cookie)
+REGTEST_PREFIX = LND_MACAROON_HEX=$(LND_MACAROON_HEX) BITCOIND_COOKIE=$(BITCOIND_COOKIE)
+
 init:
 	cargo install wasm-pack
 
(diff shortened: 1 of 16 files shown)


lib/core/Makefile
```diff
@@ -4,27 +4,48 @@ ifeq ($(UNAME), Darwin)
 	CLANG_PREFIX += AR=$(shell brew --prefix llvm)/bin/llvm-ar CC=$(shell brew --prefix llvm)/bin/clang
 endif
 
+LND_MACAROON_HEX=$(shell xxd -p ../../regtest/boltz/data/lnd1/data/chain/bitcoin/regtest/admin.macaroon | tr -d '\n')
```
dangeross, line 7: What is the behaviour here if the submodule does not exist? Should we check that before running the tests?

danielgranhao: If the submodule doesn't exist it won't be possible to start the regtest environment. I've added a check so that we get an immediate and helpful error message if the regtest setup isn't running. [8be6360](https://github.com/breez/breez-sdk-liquid/pull/816/commits/8be6360b4f764dcff49db3171d44a8cbc2baeda1)

dangeross (approved):
LGTM

roeierez (approved):
Fantastic!

PR #823, opened 2025-03-27
WASM: fix clippy and tests

danielgranhao:
This PR fixes the last things missing to clean up the Wasm clippy and tests.

On Wasm there were some clippy errors due to having non send and sync `Arc`s. Fixed by creating an Arc alias for Rc in sdk-common: https://github.com/breez/breez-sdk-greenlight/pull/1189

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -4167,7 +4167,7 @@ dependencies = [
 [[package]]
 name = "sdk-common"
 version = "0.6.2"
-source = "git+https://github.com/breez/breez-sdk?rev=34b589583e040ccf773d5b368326856994d005f0#34b589583e040ccf773d5b368326856994d005f0"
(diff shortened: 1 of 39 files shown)


dangeross (commented):
Seems like the only issue is the use of `Arc` in the wasm crate. Can you test with `Rc` instead?

edit: I'm happy keeping it `Arc` and adding an allow for it

danielgranhao:
> ~Seems like the only issue is the use of `Arc` in the wasm crate. Can you test with `Rc` instead?~ seems more complicated than I thought
> 
> edit: I'm happy keeping it `Arc` and adding an allow for it

Ended up going for an alias to Rc on Wasm builds :)

dangeross (approved):
LGTM

hydra-yse (approved):
Great stuff

PR #838, opened 2025-03-31
Resolved balance update lag after PaymentSucceeded event

Keerthi421:
This PR resolves an issue where the wallet balance update was delayed after a payment was marked as successful. The fix ensures that payment data is fully synchronized with the blockchain before updating the wallet info and emitting the PaymentSucceeded event.
Changes Made :
Moved update_wallet_info() to the Complete match arm to ensure it's only triggered at the right stage.
Added self.sync_payments_with_chain_data(false).await? before updating wallet info to ensure accurate data synchronization.
New execution sequence:
First: Sync payment data with the blockchain.
Second: Update wallet info with the latest synchronized data.
Third: Emit the PaymentSucceeded event.

Impact of the Fix
Ensures the wallet balance reflects the latest blockchain state.
Eliminates lag and incorrect balance updates after successful payments.
 Improves reliability of the PaymentSucceeded event.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,7 @@ jobs:
           workspaces: |
             lib -> target
             cli -> target
+          shared-key: "cargo-clippy-${{ hashFiles('**/Cargo.lock') }}"
 
       - name: Install Protoc
         uses: arduino/setup-protoc@v3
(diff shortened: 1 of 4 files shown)


dangeross (changes requested):
Thanks for the PR @Keerthi421. There are lots of unrelated changes in this PR, please remove what is not related

Keerthi421:
@dangeross  I will work on it .

PR #840, opened 2025-03-31
WASM: fix remaining sleep and current time uses

danielgranhao:
This PR fixes some additional uses of tokio sleep and getting the current time that we didn't spot before. It applies the same strategies that we are currently using in the rest of the codebase.

- The esplora clients use a sleep to retry requests after a delay. Fixed using `tokio_with_wasm`.
- The `is_expired()` method of `Bolt11Invoice` internally uses `std::time::SystemTime::now()`. Replaced it with an implementation that relies on `web_time.`


dangeross (approved):
LGTM

dangeross (approved):

PR #845, opened 2025-04-01
Switch Esplora client url to Breez server instance

hydra-yse:
~Waiting for breez/server#67~
This PR:
- Adds API key check for Breez waterfalls url on all client types (chain, wallet, swapper)
- Changes the main Esplora url to Breez instance

## Notes
- Switching to [custom branch](https://github.com/hydra-yse/boltz-rust/tree/esplora-custom-client) until SatoshiPortal/boltz-rust#100 is merged

### DIFF
--- a/lib/core/src/chain/bitcoin/esplora.rs
+++ b/lib/core/src/chain/bitcoin/esplora.rs
@@ -52,7 +52,6 @@ impl EsploraBitcoinChainService {
             .timeout(3)
             .max_retries(10)
             .build_async()?;
-
         let client = self.client.get_or_init(|| client);
         Ok(client)
     }
--- a/lib/core/src/chain/liquid/esplora.rs
+++ b/lib/core/src/chain/liquid/esplora.rs
@@ -10,7 +10,7 @@ use crate::{
     utils,
 };
 
-use log::{info, warn};
+use log::{error, info};
 use lwk_wollet::{
     asyncr::EsploraClientBuilder, clients::asyncr::EsploraClient, elements::hex::FromHex as _,
 };
@@ -50,7 +50,7 @@ impl EsploraLiquidChainService {
                         }
                         None => {
                             let err = "Cannot start Breez Esplora client: Breez API key is not set";
-                            warn!("{err}");
+                            error!("{err}");
                             bail!(err)
                         }
                     };
--- a/lib/core/src/chain/liquid/esplora.rs
+++ b/lib/core/src/chain/liquid/esplora.rs
@@ -1,16 +1,16 @@
 use std::{sync::OnceLock, time::Duration};
 
-use anyhow::{anyhow, Context as _, Result};
+use anyhow::{anyhow, bail, Context as _, Result};
 use tokio::sync::RwLock;
 use tokio_with_wasm::alias as tokio;
 
 use crate::{
     elements::{Address, OutPoint, Script, Transaction, Txid},
-    model::{BlockchainExplorer, Config, Utxo},
+    model::{BlockchainExplorer, Config, Utxo, BREEZ_LIQUID_ESPLORA_URL},
     utils,
 };
 
-use log::info;
+use log::{info, warn};
 use lwk_wollet::{
     asyncr::EsploraClientBuilder, clients::asyncr::EsploraClient, elements::hex::FromHex as _,
 };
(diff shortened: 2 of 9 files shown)


lib/core/src/chain/bitcoin/esplora.rs
```diff
@@ -48,11 +48,20 @@ impl EsploraBitcoinChainService {
                 anyhow::bail!("Cannot start Bitcoin Esplora chain service without an Esplora url")
             }
         };
-        let client = Builder::new(esplora_url)
-            .timeout(3)
-            .max_retries(10)
-            .build_async()?;
+        let mut builder = Builder::new(esplora_url).timeout(3).max_retries(10);
 
+        if esplora_url == BREEZ_LIQUID_ESPLORA_URL {
```
roeierez, line 53: Shouldn't the change be only for liquid esplora? Currently it seems that we change bitcoin esplora.

roeierez (approved):
LGTM

lib/core/src/chain/liquid/esplora.rs
```diff
@@ -40,10 +40,21 @@ impl EsploraLiquidChainService {
...
+            } => {
+                let mut builder = EsploraClientBuilder::new(url, self.config.network.into());
+                if url == BREEZ_LIQUID_ESPLORA_URL {
+                    match &self.config.breez_api_key {
+                        Some(api_key) => {
+                            builder = builder
+                                .header("authorization".to_string(), format!("Bearer {api_key}"));
+                        }
+                        None => {
+                            warn!("Breez API key is not set with Breez Esplora URL: client will not work as expected");
```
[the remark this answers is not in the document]
hydra-yse: I've thought about it, I agree that bailing here is better to avoid unpredictable behavior. The idea was that some endpoints don't need the API key and therefore they will work just fine. But still, we don't want to attempt a request knowing it will fail, so it's better to just cut it all off.

dangeross (approved):

lib/core/src/chain/liquid/esplora.rs
```diff
@@ -40,10 +40,23 @@ impl EsploraLiquidChainService {
...
+                let mut builder = EsploraClientBuilder::new(url, self.config.network.into());
+                if url == BREEZ_LIQUID_ESPLORA_URL {
+                    match &self.config.breez_api_key {
+                        Some(api_key) => {
+                            builder = builder
+                                .header("authorization".to_string(), format!("Bearer {api_key}"));
+                        }
+                        None => {
+                            let err = "Cannot start Breez Esplora client: Breez API key is not set";
+                            warn!("{err}");
```
danielgranhao, line 53: Nit: do we need a log here? If so, `error!` may be more appropriate

danielgranhao (approved):
LGTM. Just a nit

dangeross:
Seems like there are some warning when building wasm from the lwk changes:
```bash
warning: unused import: `std::collections::HashMap`
 --> /Users/rosssavage/Source/Projects/lwk/lwk_wollet/src/pegin.rs:5:5
  |
5 | use std::collections::HashMap;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `Error`
 --> /Users/rosssavage/Source/Projects/lwk/lwk_wollet/src/pegin.rs:9:30
  |
9 | use crate::{ElementsNetwork, Error};
  |                              ^^^^^

warning: variable does not need to be mutable
   --> /Users/rosssavage/Source/Projects/lwk/lwk_wollet/src/clients/asyncr/esplora.rs:637:13
    |
637 |         let mut builder = reqwest::Client::builder().default_headers(headers);
    |             ----^^^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default

warning: method `capabilities` is never used
   --> /Users/rosssavage/Source/Projects/lwk/lwk_wollet/src/clients/asyncr/esplora.rs:575:19
    |
52  | impl EsploraClient {
    | ------------------ method in this implementation
...
575 |     pub(crate) fn capabilities(&self) -> HashSet<Capability> {
    |                   ^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

warning: function `height_with_fed_peg_script` is never used
  --> /Users/rosssavage/Source/Projects/lwk/lwk_wollet/src/pegin.rs:14:4
   |
14 | fn height_with_fed_peg_script(network: ElementsNetwork, current_tip: u32) -> u32 {
   |    ^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: methods `get_script_batch`, `heights`, `paths`, `txs`, `tip`, and `descriptor` are never used
  --> /Users/rosssavage/Source/Projects/lwk/lwk_wollet/src/wollet.rs:59:8
   |
58 | pub trait WolletState {
   |           ----------- methods in this trait
59 |     fn get_script_batch(
   |        ^^^^^^^^^^^^^^^^
...
70 |     fn heights(&self) -> &HashMap<Txid, Option<Height>>;
   |        ^^^^^^^
71 |     fn paths(&self) -> &HashMap<Script, (Chain, ChildNumber)>;
   |        ^^^^^
72 |     fn txs(&self) -> HashSet<Txid>;
   |        ^^^
73 |     fn tip(&self) -> (Height, BlockHash);
   |        ^^^
74 |     fn last_unused(&self) -> LastUnused; // TODO change to &LastUnused when possible
75 |     fn descriptor(&self) -> WolletDescriptor;
   |        ^^^^^^^^^^
```

dangeross:
I get some issues testing this against our endpoint for the wallet scan:
```
 Failed to scan wallet: Generic { err: "Generic(\"404 page not found\\n\")" }
```
I can CURL a request to /v2/waterfalls + header auth and it works

roeierez (changes requested):
It seems that we don't send the headers in Esplora client that is passed to the swapper functions.
@hydra-yse  One way is to add an option to pass request::Client to boltz EsploraLiquidClient and configure the auth header there.

dangeross (approved):
Tested send/receive, chain send/receive

roeierez (approved):
Tested and was able to claim. LGTM

PR #848, opened 2025-04-02
Reserve addresses and release them once the tx is broadcast

dangeross:
This PR helps to address the case if the tx broadcast fails, the address is never reused (and the derivation index spirals). We reserve the claim/refund addresses for 60 blocks, if the tx broadcast fails the address can be reused once the block height is reached. If the tx succeeds we release the reservation.

We also have a case where in the send claim where we use a refund address to form the partial signature of the claim tx. We should reuse this address also.

### DIFF
--- a/lib/core/src/chain_swap.rs
+++ b/lib/core/src/chain_swap.rs
@@ -716,16 +716,18 @@ impl ChainSwapHandler {
                         match swap.user_lockup_tx_id {
                             Some(_) => {
                                 warn!("Chain Swap {id} user lockup tx has been broadcast.");
-                                let refund_tx_id = match self.refund_outgoing_swap(swap, true).await
-                                {
-                                    Ok(refund_tx_id) => Some(refund_tx_id),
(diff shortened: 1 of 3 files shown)


roeierez:
@dangeross I think the idea was to reuse the same address as long as the refund fails, isn't it?
Currently if the refund fails to broadcast we will still ask for another address on the next refund. (that on the original case happened in a loop).

dangeross:
> @dangeross I think the idea was to reuse the same address as long as the refund fails, isn't it? Currently if the refund fails to broadcast we will still ask for another address on the next refund. (that on the original case happened in a loop).

Ok, I thought in the issue we mentioned using the pool. But we can also set the refund address on the swap

roeierez:
> Ok, I thought in the issue we mentioned using the pool. But we can also set the refund address on the swap

Yes, sorry. I meant something like that. Not sure we need the pool at all. Once we pick up a refund address, we can keep it on the swap.

PR #850, opened 2025-04-02
Fix clippy test issues

dangeross:


lib/core/src/payjoin/pset/tests.rs
```diff
@@ -1,3 +1,4 @@
+#[allow(clippy::module_inception)]
 #[cfg(test)]
 mod tests {
```
danielgranhao, line 3: Wdyt of the following alternative? 

```suggestion
#![cfg(test)]
```

danielgranhao (commented):
Left a proposal for avoiding the `allow` attribute. Also, should we adjust our clippy in the makefiles / CI to have `--all-targets`?

lib/core/src/payjoin/pset/tests.rs
dangeross: Good idea, done + added `--all-targets`

danielgranhao (approved):
LGTM

PR #851, opened 2025-04-02
WASM: db backups

### LINKED ISSUE #836: WASM - persist sqlite db.
We are currently running in memory which has a performrance drawback when we reload the page.
I think rather than implementing virtual file system for the web we can achieve a very good performance improvement by persisting the db (backup snapshot) after initial restore only.
99% of the time goes for restoring old transactions via chain service. This will go away when persisting the db and use it as a starting point for the next startup.

danielgranhao:
Resolves #836

This adds a db backup mechanism on WASM by persisting the db on every sync (every 10 secs) and when new records are pulled from the real time data sync. We may reduce the frequency, but it seems to me like the overhead is negligible. May become significant in instances with a very large payment history, but we can adjust it later. I added a log to check how long it takes to create a backup so that we can track it. 

Persistence is done using [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) on the browser and the [fs](https://nodejs.org/api/fs.html) API on node.js. Other runtimes won't support backups, but we may add support in the future.

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -23,6 +23,7 @@ use log::warn;
 use lwk_wollet::WalletTx;
 use migrations::current_migrations;
 use model::PaymentTxDetails;
+use rusqlite::backup::Backup;
 use rusqlite::{
     params, params_from_iter, Connection, OptionalExtension, Row, ToSql, TransactionBehavior,
 };
@@ -72,7 +73,7 @@ impl Persister {
         if !main_db_dir.exists() {
             std::fs::create_dir_all(&main_db_dir)?;
         }
-        Self::new_inner(main_db_dir, network, sync_enabled, asset_metadata)
+        Self::new_inner(main_db_dir, network, sync_enabled, asset_metadata, None)
     }
 
     /// Creates a new Persister that only keeps data in memory.
@@ -85,9 +86,25 @@ impl Persister {
         network: LiquidNetwork,
         sync_enabled: bool,
         asset_metadata: Option<Vec<AssetMetadata>>,
+        backup_bytes: Option<Vec<u8>>,
     ) -> Result<Self> {
         let main_db_dir = PathBuf::from_str(database_id)?;
-        Self::new_inner(main_db_dir, network, sync_enabled, asset_metadata)
+        let backup_con = backup_bytes
+            .map(|data| {
+                let size = data.len();
+                let cursor = std::io::Cursor::new(data);
+                let mut conn = Connection::open_in_memory()?;
+                conn.deserialize_read_exact(rusqlite::DatabaseName::Main, cursor, size, false)?;
+                Ok::<Connection, anyhow::Error>(conn)
+            })
+            .transpose()?;
+        Self::new_inner(
+            main_db_dir,
+            network,
+            sync_enabled,
+            asset_metadata,
+            backup_con,
+        )
     }
 
     fn new_inner(
@@ -95,6 +112,7 @@ impl Persister {
         network: LiquidNetwork,
         sync_enabled: bool,
         asset_metadata: Option<Vec<AssetMetadata>>,
+        backup_con: Option<Connection>,
     ) -> Result<Self> {
         let mut sync_trigger = None;
         if sync_enabled {
@@ -108,13 +126,19 @@ impl Persister {
             sync_trigger,
         };
 
+        if let Some(backup_con) = backup_con {
+            let mut dst_con = persister.get_connection()?;
+            let backup = Backup::new(&backup_con, &mut dst_con)?;
+            backup.step(-1)?;
+        }
+
         persister.init()?;
         persister.replace_asset_metadata(asset_metadata)?;
 
         Ok(persister)
     }
 
-    pub fn get_connection(&self) -> Result<Connection> {
+    pub(crate) fn get_connection(&self) -> Result<Connection> {
         Ok(Connection::open(
             self.main_db_dir.join(DEFAULT_DB_FILENAME),
         )?)
@@ -125,8 +149,15 @@ impl Persister {
         Ok(())
     }
 
+    #[cfg(all(target_family = "wasm", target_os = "unknown"))]
+    pub fn serialize(&self) -> Result<Vec<u8>> {
+        let con = self.get_connection()?;
+        let db_bytes = con.serialize(rusqlite::DatabaseName::Main)?;
+        Ok(db_bytes.to_vec())
+    }
+
     #[cfg(any(test, feature = "test-utils"))]
-    pub fn get_database_dir(&self) -> &PathBuf {
+    pub(crate) fn get_database_dir(&self) -> &PathBuf {
         &self.main_db_dir
     }
 
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -114,7 +114,7 @@ impl Persister {
         Ok(persister)
     }
 
-    pub(crate) fn get_connection(&self) -> Result<Connection> {
+    pub fn get_connection(&self) -> Result<Connection> {
         Ok(Connection::open(
             self.main_db_dir.join(DEFAULT_DB_FILENAME),
         )?)
@@ -125,8 +125,8 @@ impl Persister {
         Ok(())
     }
 
-    #[cfg(test)]
-    pub(crate) fn get_database_dir(&self) -> &PathBuf {
+    #[cfg(any(test, feature = "test-utils"))]
+    pub fn get_database_dir(&self) -> &PathBuf {
         &self.main_db_dir
     }
 
@@ -1131,5 +1131,24 @@ mod tests {
         assert_eq!(storage.list_ongoing_swaps()?.len(), 2);
 
         Ok(())
+    }
+}
+
+#[cfg(feature = "test-utils")]
+pub mod test_helpers {
+    use super::*;
+
+    impl Persister {
+        pub fn test_insert_or_update_send_swap(&self, swap: &SendSwap) -> Result<()> {
+            self.insert_or_update_send_swap(swap)
+        }
+
+        pub fn test_insert_or_update_receive_swap(&self, swap: &ReceiveSwap) -> Result<()> {
+            self.insert_or_update_receive_swap(swap)
+        }
+
+        pub fn test_list_ongoing_swaps(&self) -> Result<Vec<Swap>> {
+            self.list_ongoing_swaps()
+        }
     }
 }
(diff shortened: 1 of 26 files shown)


lib/core/src/sdk.rs
```diff
@@ -453,6 +453,15 @@ impl LiquidSdk {
...
         let mut is_started = self.is_started.write().await;
+
+        #[cfg(all(target_family = "wasm", target_os = "unknown"))]
+        if let Err(e) = self
+            .persister
+            .restore_from_indexed_db_backup(&self.onchain_wallet.fingerprint()?)
+            .await
+        {
+            error!("Failed to restore from IndexedDB backup: {:?}", e);
+        }
```
roeierez, line 464: Since this code is specific to wasm better I think move it to the wasm crate. We already initializing the in memory persister there so it seems naturally belongs there.

lib/core/src/sdk.rs
```diff
@@ -3352,8 +3361,26 @@ impl LiquidSdk {
...
+
+        let t1 = Instant::now();
+        #[cfg(all(target_family = "wasm", target_os = "unknown"))]
+        if let Err(e) = self
+            .persister
+            .backup_to_indexed_db(&self.onchain_wallet.fingerprint()?)
+            .await
+        {
+            error!("Failed to create Indexed DB backup: {e}");
+        }
```
roeierez, line 3373: Also would be great to move this to the wasm crate as well. Perhaps we can register to the sync event in that crate?

roeierez (changes requested):
Looks good! I do think we need to find a way to move the code to the wasm crate.

danielgranhao:
> Looks good! I do think we need to find a way to move the code to the wasm crate.

@roeierez Thanks! I've adjusted the PR.

lib/core/src/persist/backup.rs
```diff
@@ -33,6 +41,92 @@ impl Persister {
 
         Ok(())
     }
+
+    #[cfg(all(target_family = "wasm", target_os = "unknown"))]
+    pub async fn backup_to_indexed_db(&self, db_name: &str) -> Result<()> {
```
roeierez, line 46: Can we move also the backup_to_indexed_db & restore_from_index_db to the wasm crate?

roeierez (changes requested):

dangeross:
As I understand indexedDB is web only (web_sys). So node.js is not supported out-of-the-box. Can we confirm that?

lib/core/src/persist/backup.rs
danielgranhao: Done in [c8bec8f](https://github.com/breez/breez-sdk-liquid/pull/851/commits/c8bec8fd64ce12ee8e9530d67c1234a955b35e8b)

danielgranhao:
> As I understand indexedDB is web only (web_sys). So node.js is not supported out-of-the-box. Can we confirm that?

Good point. I've added support for node.js in [c8bec8f](https://github.com/breez/breez-sdk-liquid/pull/851/commits/c8bec8fd64ce12ee8e9530d67c1234a955b35e8b).

lib/core/src/persist/mod.rs
```diff
@@ -114,7 +114,7 @@ impl Persister {
         Ok(persister)
     }
 
-    pub(crate) fn get_connection(&self) -> Result<Connection> {
+    pub fn get_connection(&self) -> Result<Connection> {
```
roeierez, line 117: How about instead of exposing the get_connection add an optional data parameter to the Persister::new_in_memory constructor? Also we can expose serialize to return the db for backup.
This way the WASM can reload the bytes and pass it there instead of using the connection directly.

lib/core/src/persist/mod.rs
```diff
@@ -125,8 +125,8 @@ impl Persister {
         Ok(())
     }
 
-    #[cfg(test)]
-    pub(crate) fn get_database_dir(&self) -> &PathBuf {
+    #[cfg(any(test, feature = "test-utils"))]
+    pub fn get_database_dir(&self) -> &PathBuf {
```
roeierez, line 129: This doesn't seem to be used outside the crate. Can remove the public ?

roeierez (commented):
Looks good! a few Nit suggestions.

packages/wasm/examples/node/src/cli.js
```diff
@@ -145,7 +145,8 @@ const initSdk = async () => {
     const mnemonic = process.env.MNEMONIC
 
     // Connect using the config
-    const config = defaultConfig('mainnet', breezApiKey)
+    let config = defaultConfig('mainnet', breezApiKey)
+    config.workingDir = "./.data"
```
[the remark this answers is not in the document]
danielgranhao: Added in [0b031a9](https://github.com/breez/breez-sdk-liquid/pull/851/commits/0b031a9a8de9e1fd0320c73d3043543ca91c0cc1)

lib/core/src/persist/mod.rs
danielgranhao: I did dislike exposing that method. Thanks for the proposal. Done in [0b031a9](https://github.com/breez/breez-sdk-liquid/pull/851/commits/0b031a9a8de9e1fd0320c73d3043543ca91c0cc1).

It' a bit annoying that the fingerprint is "needed" for getting a good unique path. The wallet provides the fingerprint method but it can only be created with a persister. I duplicated a bit of code to derive the fingerprint in the Wasm crate as I think that is the simplest solution for now.

danielgranhao: Also, the previous approach had a larger issue: the recovery was made after the db was initialized, meaning that we would be missing migrations. Now, we recover before applying migrations.

lib/core/src/persist/mod.rs
```diff
@@ -85,16 +86,33 @@ impl Persister {
...
+        backup_bytes: Option<Vec<u8>>,
     ) -> Result<Self> {
         let main_db_dir = PathBuf::from_str(database_id)?;
-        Self::new_inner(main_db_dir, network, sync_enabled, asset_metadata)
+        let backup_con = backup_bytes
+            .map(|data| {
+                let size = data.len();
+                let cursor = std::io::Cursor::new(data);
+                let mut conn = Connection::open_in_memory()?;
+                conn.deserialize_read_exact(rusqlite::DatabaseName::Main, cursor, size, false)?;
```
dangeross, line 97: I wonder what we should do in the case the restore of the backup fails (maybe the file/indexedDb get corrupted). Maybe if restore fails we should continue without and re-hydrate via data-sync/recoverer. If we do nothing the SDK is basically bricked until the web user clears the indexedDB table or nodejs user deletes/changes the workingDir

roeierez (approved):
LGTM

dangeross (commented):
Looks good. Just a question about failure handling

lib/core/src/persist/mod.rs
danielgranhao: Good point! Done in [2277efc](https://github.com/breez/breez-sdk-liquid/pull/851/commits/2277efcd81036f9335bd79d211001252f1f307c1)

lib/core/src/persist/mod.rs
```diff
@@ -108,6 +130,17 @@ impl Persister {
             sync_trigger,
         };
 
+        if let Some(backup_con) = backup_con {
+            if let Err(e) = {
```
dangeross, line 134: Doesn't this need to be a closure for this error handling to work on a block level?

danielgranhao: Oops! Fixed in [0832da8](https://github.com/breez/breez-sdk-liquid/pull/851/commits/0832da8fe3b5189723ef48f16a87d8d2a22f4bc8)

dangeross (approved):
LGTM

lib/core/src/persist/mod.rs
```diff
@@ -85,16 +86,37 @@ impl Persister {
...
         asset_metadata: Option<Vec<AssetMetadata>>,
+        backup_bytes: Option<Vec<u8>>,
     ) -> Result<Self> {
         let main_db_dir = PathBuf::from_str(database_id)?;
-        Self::new_inner(main_db_dir, network, sync_enabled, asset_metadata)
+        let backup_con = backup_bytes
+            .map(|data| {
+                let size = data.len();
+                let cursor = std::io::Cursor::new(data);
+                let mut conn = Connection::open_in_memory()?;
```
Spxg, line 96: You can actually `open` directly, and the default is also in memory, and the data will not be lost before refreshing the page. `open_in_memory` uses `:memory:`, and closing the database connection will lose data.

lib/wasm/src/backup/indexed_db.rs
```diff
@@ -0,0 +1,71 @@
...
+use js_sys::{global, Reflect};
+
+const IDB_STORE_NAME: &str = "BREEZ_SDK_LIQUID_DB_BACKUP_STORE";
+
+pub(crate) fn is_indexed_db_supported() -> bool {
+    let global = global();
+    Reflect::get(&global, &"indexedDB".into()).is_ok_and(|v| !v.is_undefined())
+}
+
+pub(crate) async fn backup_to_indexed_db(db_bytes: Vec<u8>, db_name: &str) -> Result<()> {
```
Spxg, line 14: It seems that you achieve persistence by backing up the memory data to indexed db and then loading it out next time you start. Recently, I implemented a new vfs that can help you simplify this step. 
For details, see:
 <https://github.com/Spxg/sqlite-wasm-rs/blob/master/sqlite-wasm-rs/src/vfs/README.md#relaxedidbvfs>
 
<https://github.com/Spxg/sqlite-wasm-rs/blob/master/sqlite-wasm-rs/tests/vfs/relaxed_idb.rs>

By the way, what is the reason why you don’t use opfs-sahpool vfs directly, but use memory? Is it because it is not running on the worker?

Spxg (commented):
Forgive me for coming here uninvited, I noticed that you are using my modified rusqlite and I am curious about how you use it.

lib/core/src/persist/mod.rs
danielgranhao: The idea was to use temporary in-memory connection to deserialize the backup to, only then using `Backup` to move the data to the main db instance (that uses sqlite-wasm-rs). [`deserialize`](https://www.sqlite.org/c3ref/deserialize.html) opens a temporary in-memory connection anyway.

lib/wasm/src/backup/indexed_db.rs
danielgranhao: > It seems that you achieve persistence by backing up the memory data to indexed db and then loading it out next time you start.

Exactly.

> Recently, I implemented a new vfs that can help you simplify this step.

Interesting! We'll definitely take a look into this

> By the way, what is the reason why you don’t use opfs-sahpool vfs directly, but use memory? Is it because it is not running on the worker?

Yes, that was the main reason.

danielgranhao (commented):
> Forgive me for coming here uninvited, I noticed that you are using my modified rusqlite and I am curious about how you use it.

@Spxg Not a problem at all. We value your work on `rusqlite`/`sqlite-wasm-rs` and your feedback here is appreciated.

PR #858, opened 2025-04-04
Waterfalls script history

### LINKED ISSUE #855: Use waterfall to get lbtc recoverer script history
waterfalls has a way to pass a list of addresses and return their script history as a batch. We can use that in the recoverer.

dangeross:
Updates LWK to use wateralls to fetch scripts history (See PR https://github.com/Blockstream/lwk/pull/80).

~Calling `/v2/waterfalls?addresses=` suffers the same issue as calling `/address/<addr>/txs` in waterfalls, in that it only returns a subset of the results esplora does (missing txs in script history). So setting this PR to draft while we determine if that is the intended behavior.~

Fixes #855

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -2727,7 +2727,7 @@ dependencies = [
 [[package]]
 name = "lwk_wollet"
 version = "0.9.0"
-source = "git+https://github.com/breez/lwk?rev=f56909c67be069dd99876a0a72de342d1dd7e334#f56909c67be069dd99876a0a72de342d1dd7e334"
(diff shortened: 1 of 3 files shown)


roeierez:
> Calling `/v2/waterfalls?addresses=` suffers the same issue as calling `/address/<addr>/txs` in waterfalls

Do you see this on our instance too? Asking because there is not difference from this call to calling with descriptor regarding the way waterfall server fetches the script history. For me using descriptor works, that's why I am asking.

roeierez (approved):
LGTM

danielgranhao (approved):

PR #860, opened 2025-04-04
Wasm: Reset interval after sync

### LINKED ISSUE #857: Skip missing interval doesn't work in wasm
I have noticed that if sync takes a lot of time (first sync) then immediately after it finished the sync runs several times in a row.
I think this indicates that the missing intervals were queued.

dangeross:
As Interval MissedTickBehavior is not available for Wasm, after the sync we reset the interval to ensure there is a gap before the next sync.

This PR also updates the lightning dependency to 0.1.2 

Fixes #857

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -625,6 +625,9 @@ impl LiquidSdk {
                         let partial_sync = (is_new_liquid_block || is_new_bitcoin_block).not();
                         _ = cloned.sync(partial_sync).await;
 
+                        #[cfg(all(target_family = "wasm", target_os = "unknown"))]
+                        interval.reset();
+
                         // Update swap handlers
                         if is_new_liquid_block {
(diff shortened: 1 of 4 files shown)


lib/core/src/sdk.rs
```diff
@@ -625,6 +625,9 @@ impl LiquidSdk {
                         let partial_sync = (is_new_liquid_block || is_new_bitcoin_block).not();
                         _ = cloned.sync(partial_sync).await;
 
+                        #[cfg(all(target_family = "wasm", target_os = "unknown"))]
```
roeierez, line 628: NIT: Perhaps it is better to put this line at the end of the code block for clarity?

roeierez (commented):
Just a small NIT.

roeierez (approved):
LGTM

danielgranhao (approved):

PR #863, opened 2025-04-07
Handle send swap minimal_batched

dangeross:
This PR incorporates the [minimal_batched submarine pair limit](https://github.com/SatoshiPortal/boltz-rust/pull/99) to allow batched send swap payments, where in the case of a batched payment there is no need to cooperatively claim.

**Testing notes:**
This lowers the minimum send payment amount to around 21 sats. 
- Pay invoices (from non-SDK apps) between 21 - 999 sats. Should complete and have a valid preimage.
- Pay invoices (from non-SDK apps) 1000 + sats. Should complete and have a valid preimage.

### DIFF
--- a/lib/core/src/recover/handlers/handle_send_swap.rs
+++ b/lib/core/src/recover/handlers/handle_send_swap.rs
@@ -62,13 +62,10 @@ impl SendSwapHandler {
 
         // Recover preimage if needed
         if recovered_data.lockup_tx_id.is_some() && send_swap.preimage.is_none() {
-            let pair_limits = send_swap.get_boltz_pair()?.limits;
-            // If the swap is batched or we have a claim tx id, we can attempt
-            // to recover the preimage cooperatively. If we cannot recover it
-            // cooperatively, we can try to recover it from the claim tx.
-            if send_swap.receiver_amount_sat < pair_limits.minimal
-                || recovered_data.claim_tx_id.is_some()
-            {
+            // We can attempt to recover the preimage cooperatively after we know the
+            // lockup tx was broadcast. If we cannot recover it cooperatively,
+            // we can try to recover it from the claim tx.
+            if recovered_data.claim_tx_id.is_some() {
                 match Self::recover_preimage(
                     context,
                     recovered_data.claim_tx_id.clone(),
--- a/lib/core/src/recover/handlers/handle_send_swap.rs
+++ b/lib/core/src/recover/handlers/handle_send_swap.rs
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use boltz_client::boltz::SubmarinePairLimits;
 use boltz_client::ToHex;
 use log::{debug, error, warn};
 use lwk_wollet::elements::Txid;
@@ -130,7 +131,12 @@ impl SendSwapHandler {
         // Update state based on recovered data
         let timeout_block_height = send_swap.timeout_block_height as u32;
         let is_expired = current_block_height >= timeout_block_height;
-        if let Some(new_state) = recovered_data.derive_partial_state(is_expired) {
+        let pair_limits = send_swap.get_boltz_pair()?.limits;
+        if let Some(new_state) = recovered_data.derive_partial_state(
+            send_swap.receiver_amount_sat,
+            pair_limits,
+            is_expired,
+        ) {
             send_swap.state = new_state;
         }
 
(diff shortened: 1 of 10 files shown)


lib/core/src/send_swap.rs
```diff
@@ -93,14 +94,24 @@ impl SendSwapHandler {
...
             SubSwapStates::TransactionClaimPending => {
                 if swap.metadata.is_local {
-                    self.cooperate_claim(&swap).await.map_err(|e| {
-                        error!("Could not cooperate Send Swap {id} claim: {e}");
-                        anyhow!("Could not post claim details. Err: {e:?}")
-                    })?;
+                    let pair = swap.get_boltz_pair()?;
+                    if swap.receiver_amount_sat < pair.limits.minimal {
+                        // The swap is to be batched and does not need to be cooperatively claimed
+                        self.get_claim_details_and_validate_preimage(&swap)
```
roeierez, line 104: This one (although has a "get" notation) updates the swap state which might make it a bit difficult to follow. 
How about calling this method anyway (also on the cooperative claim), updating the swap state (using the result) and also pass the result to cooperative_claim function?

roeierez (changes requested):
Looks good. Some thoughts about re-arranging the methods.

roeierez (approved):
LGTM

danielgranhao (approved):
LGTM

danielgranhao (approved):
LGTM

lib/core/src/recover/handlers/handle_send_swap.rs
```diff
@@ -61,25 +61,33 @@ impl SendSwapHandler {
...
-                &swap_id,
-                context.swapper.clone(),
-            )
-            .await
+        if recovered_data.lockup_tx_id.is_some() && send_swap.preimage.is_none() {
+            let pair_limits = send_swap.get_boltz_pair()?.limits;
+            // If the swap is batched or we have a claim tx id, we can attempt
+            // to recover the preimage cooperatively. If we cannot recover it
+            // cooperatively, we can try to recover it from the claim tx.
+            if send_swap.receiver_amount_sat < pair_limits.minimal
```
roeierez, line 69: I am not sure we need to check if it is under the limit. I think botz expose the preimage on TransactionClaimPending pending state. I asked Michael.

lib/core/src/send_swap.rs
```diff
@@ -93,14 +94,29 @@ impl SendSwapHandler {
...
-            // the claim by doing so cooperatively
+            // Boltz has detected the lockup in the mempool. If the swap is not to be batched
+            // we can speed up the claim by doing so cooperatively.
             SubSwapStates::TransactionClaimPending => {
                 if swap.metadata.is_local {
-                    self.cooperate_claim(&swap).await.map_err(|e| {
-                        error!("Could not cooperate Send Swap {id} claim: {e}");
-                        anyhow!("Could not post claim details. Err: {e:?}")
-                    })?;
+                    let claim_tx_response = self.swapper.get_send_claim_tx_details(&swap).await?;
```
roeierez, line 101: Just verifying this will not return error in case of batched claim (since the swapper won't want to cooperate on claim here).

roeierez (commented):
Looks correct now with the preimage. Small comments dropped.

lib/core/src/recover/handlers/handle_send_swap.rs
dangeross: Removed [4208d70](https://github.com/breez/breez-sdk-liquid/pull/863/commits/4208d701b7f527440b438c8a8259259a6531550f)

lib/core/src/send_swap.rs
dangeross: It was returning an error, fixed [4208d70](https://github.com/breez/breez-sdk-liquid/pull/863/commits/4208d701b7f527440b438c8a8259259a6531550f)

lib/core/src/recover/handlers/handle_send_swap.rs
```diff
@@ -61,25 +61,30 @@ impl SendSwapHandler {
...
-                    recovered_data.claim_tx_id = None;
-                }
-                Err(e) => {
-                    error!("Failed to recover preimage for swap {swap_id}: {e}");
-                    recovered_data.claim_tx_id = None
+        if recovered_data.lockup_tx_id.is_some() && send_swap.preimage.is_none() {
+            // We can attempt to recover the preimage cooperatively after we know the
+            // lockup tx was broadcast. If we cannot recover it cooperatively,
+            // we can try to recover it from the claim tx.
+            if recovered_data.claim_tx_id.is_some() {
```
roeierez, line 68: This will not recover the preimage from the boltz endpoint in case of under the limit. Is that what we want?

lib/core/src/send_swap.rs
```diff
@@ -93,14 +94,35 @@ impl SendSwapHandler {
...
+            // Boltz has detected the lockup in the mempool. If the swap is not to be batched
+            // we can speed up the claim by doing so cooperatively.
             SubSwapStates::TransactionClaimPending => {
                 if swap.metadata.is_local {
-                    self.cooperate_claim(&swap).await.map_err(|e| {
-                        error!("Could not cooperate Send Swap {id} claim: {e}");
-                        anyhow!("Could not post claim details. Err: {e:?}")
-                    })?;
+                    let pair = swap.get_boltz_pair()?;
+                    if swap.receiver_amount_sat < pair.limits.minimal {
```
roeierez, line 102: Sorry for the addition here but I think that we shouldn't count on the minimal limits on whether to cooperate with the claim or not since it is up to the swapper on whether to cooperate or not.
I think we should:
1. Try to cooperative claim anyway and ignore the failure.
2 If we failed to cooperate (still don't have the preimage) use the other endpoint to get it.

roeierez (changes requested):

lib/core/src/recover/handlers/handle_send_swap.rs
dangeross: No, fixed [4caafc3](https://github.com/breez/breez-sdk-liquid/pull/863/commits/4caafc32fd5133b8cf6b306632679a70a69d055b)

lib/core/src/send_swap.rs
dangeross: Done [4caafc3](https://github.com/breez/breez-sdk-liquid/pull/863/commits/4caafc32fd5133b8cf6b306632679a70a69d055b)

roeierez (approved):
LGTM

PR #867, opened 2025-04-08
Set correct realtime sync url for wasm

roeierez:


dangeross (approved):
LGTM. It still concerns me that we use the same data-sync URL for all networks...

roeierez:
> LGTM. It still concerns me that we use the same data-sync URL for all networks...

absolutely. Let's handle that out of this PR.

PR #871, opened 2025-04-10
Add failsafe mechanism for real-time sync

hydra-yse:
This PR:
- Improves logging for the real-time sync service (prefixing, log level and messages)
- Adds a failsafe mechanism in case of consecutive errors while running the event loop (> 5). After too many failed attempts, it will wipe the sync tables (incoming, outgoing and state) and re-enqueue all local swaps for push.
- Improves error handling with new `PushError` and `PullError` variants (reflected in logs)

### DIFF
--- a/lib/core/src/sync/mod.rs
+++ b/lib/core/src/sync/mod.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 use std::time::Duration;
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Result};
 use log::{info, trace, warn};
 use sdk_common::utils::Arc;
 use tokio::sync::{broadcast, watch};
@@ -18,8 +18,7 @@ use crate::recover::recoverer::Recoverer;
 use crate::sync::model::data::{
     ChainSyncData, PaymentDetailsSyncData, ReceiveSyncData, SendSyncData,
 };
-use crate::sync::model::DecryptionInfo;
-use crate::sync::model::{Record, SetRecordRequest, SetRecordStatus};
+use crate::sync::model::{DecryptionInfo, Record, SetRecordRequest, SetRecordStatus};
 use crate::utils;
 use crate::{
     persist::{cache::KEY_LAST_DERIVATION_INDEX, Persister},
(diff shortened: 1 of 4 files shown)


danielgranhao (commented):
Looks good! Just a couple of nits

lib/core/src/sync/mod.rs
```diff
@@ -87,15 +94,34 @@ impl SyncService {
         }
     }
 
-    async fn run_event_loop(&self) {
+    async fn run_event_loop(&self, failed_sync_counter: &mut u8) {
```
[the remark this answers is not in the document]
hydra-yse: Yes, but it would have to be under a mutex, and then we'd have to handle the set/increment both inside the event loop and on the reconnect handler. This just seemed like the easier option

danielgranhao (approved):

dangeross (approved):
LGTM. Can you let @roeierez review it also?

Were you able to test this against my old db state and see if it recovered ok?

hydra-yse:
@dangeross Yes, it wiped the database and was able to pull the records again

lib/core/src/sync/mod.rs
```diff
@@ -87,15 +86,34 @@ impl SyncService {
...
+                return;
+            }
+            *failed_sync_counter = 0;
+        }
+
+        let pulled_records_count = match self.pull().await {
+            Ok(pulled_records_count) => pulled_records_count,
+            Err(err) => {
+                warn!("realtime-sync: Could not pull records - {err:?}");
+                *failed_sync_counter += 1;
```
roeierez, line 105: We should be more accurate IMO with regards to the specific error.
I would only consider remote conflict error in the count as we don't want other errors (such as network) to affect the db wipe decision.

roeierez (changes requested):

lib/core/src/sync/mod.rs
hydra-yse: Done in 2c5bea1

lib/core/src/sync/mod.rs
```diff
@@ -87,15 +87,34 @@ impl SyncService {
...
```
roeierez, line 106: Why do we increment the fail counter here? We can still get here arbitrary errors such as network errors where we don't want to reset the sync.

lib/core/src/sync/mod.rs
```diff
@@ -87,15 +87,34 @@ impl SyncService {
...
         };
-        let Ok(pushed_records_count) = self.push().await else {
-            warn!("realtime-sync: Could not push records");
-            return;
+
+        let pushed_records_count = match self.push().await {
+            Ok(pushed_records_count) => pushed_records_count,
+            Err(err) => {
+                warn!("realtime-sync: Could not push records - {err:?}");
+                *failed_sync_counter += 1;
```
roeierez, line 115: Also here I don't see how this only addresses conflict errors.

roeierez (changes requested):

lib/core/src/sync/mod.rs
hydra-yse: You're right, we're also making a network request in the pull case so I'll exclude that from the error

lib/core/src/sync/mod.rs
hydra-yse: That's because the only way we can return error here is if:
1. There's a problem with the persister (e.g. `get_sync_outgoing_changes` fails)
2. Some records have a conflict
But I agree maybe I should make the error more explicit throughout the whole call chain, it makes more sense.

lib/core/src/sync/mod.rs
```diff
@@ -446,102 +499,140 @@ impl SyncService {
...
-        if !succeded.is_empty() {
+        let succeded_len = succeded.len();
+        if succeded_len > 0 {
             self.persister
-                .remove_sync_outgoing_changes(succeded.clone())?;
+                .remove_sync_outgoing_changes(succeded.clone())
+                .map_err(PushError::persister)?;
+        }
+
+        if succeded_len != total_changes - recoverable_errors {
```
roeierez, line 627: Not sure I understand this.
If we have total of 10 records, 6 succeeded and 4 ended with conflict then we don't return the ExcessiveRecordConflicts error?

roeierez (changes requested):

lib/core/src/sync/mod.rs
hydra-yse: If we have 10 records, 6 succeded and 4 with error, these 4 errors may be of different kinds. If the error is not of type "RecordConflict", then it's marked as recoverable (i.e. we don't return error). If it is of type "RecordConflict" this condition will error out with ExcessiveRecordConflicts

PR #872, opened 2025-04-10
Wasm: Run tests on node.js

### LINKED ISSUE #859: WASM: run tests on node.js
We currently always run `wasm-pack test` in the browser. Given that we would also like to support node.js, we should also run them on node. https://github.com/breez/breez-sdk-liquid/pull/851 shows how we can configure the tests to run on either runtime (using a feature) on the wasm create. We can do the same on `core` and `sdk-common`.

danielgranhao:
Resolves #859 

This PR sets up Wasm tests to run on node.js (in addition to the existing ones running in a headless browser)


roeierez (approved):
LGTM

PR #874, opened 2025-04-11
Attempt UTXO selection when building a wallet tx

dangeross:
This PR improves the fees when sending payments by selecting only the needed wallet UTXOs instead of allowing LWK to use all UTXOs. This enables consistent 26 sat fee using 1 UTXO with change and 20 sat fee using 1 UTXO with no change (receiver amount + fee = UTXO value). If no optimal UTXO selection can be determined, we fallback to LWK using all UTXOs.

- [Updates LWK](https://github.com/breez/lwk/commit/663ec5531b6be617eee3836e0f2c988d56f60acb) to allow finishing a TxBuilder with no change output
- Move most of the payjoin UTXO selection and network fee calculation modules to the wallet module
- Performs UTXO selection to find the best UTXO set to fit the fee using `utxo_select_dynamic()`, where the fee is dynamically calculated depending on utxo + output count


danielgranhao (approved):
Other than the potential complexity issue pointed out by @roeierez, LGTM!

dangeross:
I'll do some benchmarks. Maybe we can make it opt-in from the config instead?

roeierez:
> I'll do some benchmarks. Maybe we can make it opt-in from the config instead?

Perhaps we just need to put a suspender somewhere that limits the iterations. At some point if we didn't find a match we can fallback to the current approach that uses all utxos.

roeierez (approved):

hydra-yse (approved):
Looks good!

PR #875, opened 2025-04-11
Wasm: prevent IndexedDb wallet cache corruption and gracefully handle it

danielgranhao:
This PR addresses an issue where loading wallet cache updates on the browser would fail with `Missing update on index <index>`, which could happen if there are concurrent SDK instances. If one of the instances cleared the cache, the other would not notice and would continue to persist updates, thus introducing gaps.

This adds prevention by making sure an update only persists if there are no gaps in the persisted updates and also gracefully handles the gaps if, for any reason, they still happen.


roeierez (approved):
LGTM

dangeross (approved):
LGTM

Issue #828, opened 2025-03-28
Balance updates sometimes lag after `PaymentSucceeded` event


danielgranhao:
Occasionally, when running the end-to-end tests proposed in #816, the balance isn't updated after receiving the `PaymentSucceeded` event, only updating later.

Keerthi421:
@danielgranhao  I would like to work on this issue

PR #879, opened 2025-04-13
Fix Balance Update Lag After PaymentSucceeded Event

Keerthi421:
Issue:
Balance updates were lagging after a PaymentSucceeded event due to a race condition between wallet updates and event emissions, causing clients to receive outdated balance information.
Changes Made:
Enhanced emit_payment_updated Function:
->Re-fetch payment data after updating wallet info for transactions with a Complete status.
->Ensures that the PaymentSucceeded event contains the most up-to-date balance information.
Improved sync Method:
->Added an explicit call to update_wallet_info() at the end of the sync process.
->Ensures the wallet balance is always refreshed during sync operations, keeping the information up-to-date.
Why These Changes Work:
->These changes address the race condition by ensuring thay:The PaymentSucceeded event contains the latest payment data with accurate balance updates.
The sync method guarantees that wallet info is always refreshed during synchronization.
Fixes issue #828

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -741,6 +741,20 @@ impl LiquidSdk {
             match self.persister.get_payment(&id)? {
                 Some(payment) => {
                     self.update_wallet_info().await?;
+                    
+                    // For Complete status, re-fetch the payment to ensure we have the most up-to-date data
+                    // after the wallet_info update, which may have updated balances
+                    let payment = match payment.status {
+                        Complete => {
+                            // Try to get the refreshed payment data
+                            match self.persister.get_payment(&id)? {
+                                Some(refreshed_payment) => refreshed_payment,
+                                None => payment // Fall back to original payment data if not found
+                            }
+                        },
+                        _ => payment
+                    };
+                    
                     match payment.status {
                         Complete => {
                             self.notify_event_listeners(SdkEvent::PaymentSucceeded {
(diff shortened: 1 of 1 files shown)


lib/core/src/sdk.rs
```diff
@@ -741,6 +741,20 @@ impl LiquidSdk {
...
+                    let payment = match payment.status {
+                        Complete => {
+                            // Try to get the refreshed payment data
+                            match self.persister.get_payment(&id)? {
+                                Some(refreshed_payment) => refreshed_payment,
+                                None => payment // Fall back to original payment data if not found
+                            }
+                        },
+                        _ => payment
+                    };
```
roeierez, line 756: We already fetch the payment above. To save the extra fetch (and code) perhaps it is better to move the call to update_wallet_info right before the first fetch?

lib/core/src/sdk.rs
```diff
@@ -3513,6 +3527,10 @@ impl LiquidSdk {
                 self.sync_payments_with_chain_data(partial_sync).await?;
             }
         }
+        
+        // Explicitly update wallet info at the end of sync to ensure balance is up-to-date
+        self.update_wallet_info().await?;
```
roeierez, line 3532: This doesn't seem necessary as the sync_payments_with_chain_data already updates the wallet info.

roeierez (changes requested):

lib/core/src/sdk.rs
Keerthi421: @roeierez  I have made changes , Please review

roeierez: @Keerthi421 I don't see any changes in your last commit apart from a comment change.
I have a suspicion that you are pushing your changes via some AI tool without really going over them or understanding what you are doing.
I am basing my comment here on other PRs of yours that the team has reviewed.
Can you elaborate?

PR #880, opened 2025-04-14
fix: Interval not compatible with node.js

danielgranhao:
The `Interval` from [`tokio_with_wasm`](https://docs.rs/tokio_with_wasm/latest/tokio_with_wasm/) causes a panic on node.js. 

This PR proposes fixing it by simply not using Interval and using sleep instead. I did try [`wasmtimer`](https://github.com/whizsid/wasmtimer-rs), which at first seemed like it provided a good Interval implementation, but I later realized it was buggy: using it prevented the shutdown branches from ever being picked up.

I'm of course open to ideas/alternatives where tasks are still started at regular intervals

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -574,12 +574,9 @@ impl LiquidSdk {
             let mut current_liquid_block: u32 = 0;
             let mut current_bitcoin_block: u32 = 0;
             let mut shutdown_receiver = cloned.shutdown_receiver.clone();
-            let mut interval = tokio::time::interval(Duration::from_secs(10));
-            #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
-            interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
             loop {
                 tokio::select! {
-                    _ = interval.tick() => {
+                    _ = tokio::time::sleep(Duration::from_secs(10)) => {
                         info!("Track blocks loop ticked");
                         // Get the Liquid tip and process a new block
                         let t0 = Instant::now();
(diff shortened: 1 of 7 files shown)


dangeross (approved):
utACK

lib/core/src/sdk.rs
```diff
@@ -574,12 +574,9 @@ impl LiquidSdk {
             let mut current_liquid_block: u32 = 0;
             let mut current_bitcoin_block: u32 = 0;
             let mut shutdown_receiver = cloned.shutdown_receiver.clone();
-            let mut interval = tokio::time::interval(Duration::from_secs(10));
-            #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
-            interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
             loop {
                 tokio::select! {
-                    _ = interval.tick() => {
+                    _ = tokio::time::sleep(Duration::from_secs(10)) => {
```
roeierez, line 579: The previous interval implementation first tick was triggered immediately and here we are first waiting 10 seconds. Perhaps it is better to keep the current working behavior?

roeierez (changes requested):

lib/core/src/sdk.rs
danielgranhao: Agreed. Done in [a64c3e7](https://github.com/breez/breez-sdk-liquid/pull/880/commits/a64c3e700dddb9fa105f48ec97b3363f81491d14) for the track new blocks task. IMO it isn't needed on the status stream ping.

I also adjusted the E2E tests to wait for the first sync before starting. The delayed first sync was causing the tests to fail. Now there is no delay any more but it's a good idea to wait for sync anyway.

dangeross (approved):

roeierez (approved):

PR #883, opened 2025-04-16
Prevent concurrent claim attempts

danielgranhao:
This PR addresses a race condition where concurrent attempts to claim the same swap (both Chain Swaps and Receive Swaps) could lead to failures when interacting with the Boltz backend. The server initiates a session when each claim is started, so when a second claim is initiated in quick succession (before the signature is submitted), the first session is dropped, causing the submitted partial signature to be considered invalid.

The proposed fix is to introduce in-memory locking. There is an edge case where a panic may cause a lock to be held until restart, but this is unlikely (a panic on either the sync or the status stream tasks would be a big issue itself), so IMO it's acceptable.

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -637,7 +637,7 @@ checksum = "829a082bd3761fde7476dc2ed85ca56c11628948460ece621e4f56fef5046567"
 [[package]]
 name = "boltz-client"
 version = "0.3.0"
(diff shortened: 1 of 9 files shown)


danielgranhao:
> LGTM. Let's just verify we have a reasonable timeout for the swapper API so we don't risk in keeping the swap locked for claim.

Good point. Adding it to the boltz client in https://github.com/SatoshiPortal/boltz-rust/pull/102

dangeross (approved):

dangeross (approved):

PR #891, opened 2025-04-21
Drop unnecessary tip requests

### LINKED ISSUE #884: Every sync invokes 3 tip requests
In the browser it can be seen clearly that we invoke one waterfall request but 3 times tip. Let's try to optimize it and call it only once per wallet sync.

danielgranhao:
Resolves #884

### DIFF
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -833,6 +833,12 @@ pub struct BlockchainInfo {
     pub bitcoin_tip: u32,
 }
 
+#[derive(Copy, Clone)]
+pub(crate) struct ChainTips {
+    pub liquid_tip: u32,
+    pub bitcoin_tip: u32,
+}
+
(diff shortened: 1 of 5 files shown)


lib/core/src/model.rs
```diff
@@ -833,6 +833,12 @@ pub struct BlockchainInfo {
     pub bitcoin_tip: u32,
 }
 
+#[derive(Copy, Clone)]
+pub(crate) struct ChainTips {
```
dangeross, line 837: Why not reuse `BlockchainInfo`?

danielgranhao: The concepts seemed different to me, even though currently they hold the same data. We may want to add non-tip-related data to `BlockchainInfo`. That said, I'm also fine with using the same struct for now and splitting only in the future if needed.

roeierez (approved):
LGTM

dangeross (approved):

PR #897, opened 2025-04-25
Side-Swap: Add base communication layer and endpoints

hydra-yse:
This PR:
- Adds a base WebSocket layer to allow connections and exchange messages with the SideSwap API
- Adds `sideswap_api` as a dependency for the API types (currently pointing to local crate)

## To-dos
- [X] Add WebSocket start on-demand (swap lifetime only)
- [x] Fix WASM incompatibilities
- [x] Address `sideswap_api` crate issues or switch to custom branch

### DIFF
--- a/lib/core/src/side_swap/api/mod.rs
+++ b/lib/core/src/side_swap/api/mod.rs
@@ -1,0 +1,338 @@
+use anyhow::{anyhow, bail, Result};
+use base64::Engine;
+use futures_util::{SinkExt as _, StreamExt};
+use log::{debug, error, info, warn};
+use request_handler::SideSwapRequestHandler;
+use response_handler::SideSwapResponseHandler;
+use sdk_common::bitcoin::hashes::hex::ToHex as _;
+use sdk_common::prelude::{parse_json, RestClient};
+use sdk_common::utils::Arc;
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+use sideswap_api::http_rpc::{SwapSignRequest, SwapStartRequest};
+use sideswap_api::{StartSwapWebRequest, SubscribePriceStreamRequest, Utxo};
+use std::collections::HashMap;
+use std::str::FromStr;
+use std::time::Duration;
+use tokio::sync::{watch, OnceCell};
+use tokio::task::JoinHandle;
+use tokio::time::sleep;
+use tokio_with_wasm::alias as tokio;
+
+use boltz_client::boltz::tokio_tungstenite_wasm;
+use tokio_tungstenite_wasm::{connect, Message};
+
+use model::*;
+
+use crate::bitcoin::base64;
+use crate::elements::{self, pset::PartiallySignedTransaction, Address, AssetId};
+use crate::error::PaymentError;
+use crate::model::Config;
+use crate::wallet::OnchainWallet;
+use crate::{ensure_sdk, utils};
+
+pub(crate) mod model;
+mod request_handler;
+mod response_handler;
+
+pub const SIDESWAP_MAINNET_URL: &str = "wss://api.sideswap.io/json-rpc-ws";
+pub const SIDESWAP_TESTNET_URL: &str = "wss://api-testnet.sideswap.io/json-rpc-ws";
+
+pub(crate) struct SideSwapService {
+    config: Config,
+    rest_client: Arc<dyn RestClient>,
+    onchain_wallet: Arc<dyn OnchainWallet>,
+    request_handler: SideSwapRequestHandler,
+    response_handler: SideSwapResponseHandler,
+    event_loop_handle: OnceCell<JoinHandle<()>>,
+}
+
+impl SideSwapService {
+    pub(crate) fn from_sdk(sdk: &crate::sdk::LiquidSdk) -> Arc<Self> {
+        Self::new(
+            sdk.config.clone(),
+            sdk.rest_client.clone(),
+            sdk.onchain_wallet.clone(),
+            sdk.shutdown_receiver.clone(),
+        )
+    }
+
+    pub(crate) fn new(
+        config: Config,
+        rest_client: Arc<dyn RestClient>,
+        onchain_wallet: Arc<dyn OnchainWallet>,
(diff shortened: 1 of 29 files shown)


lib/core/src/side_swap/api/notifications.rs
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    pub(crate) async fn handle_notification(&self, notif: Notification) {
+        match notif {
+            Notification::Market(MarketNotification::Quote(new_quote)) => {
+                let mut quotes = self.quotes.write().await;
+                quotes.insert(new_quote.quote_sub_id, new_quote);
+            }
+            _ => todo!(),
```
roeierez, line 26: Should we ignore here?

roeierez (commented):
Looks good to me at this point.
Moving forward I think ideally we would have the sideswap session live only for the duration of each API call.

lib/core/src/side_swap/api/notifications.rs
hydra-yse: Yes, just throwing a debug message

hydra-yse:
@roeierez I think that rather than per API call it's better to let the caller decide when to start/stop the service. So ideally we want to keep the ws running for a whole swap flow (or at least, partly doing prepare, and partly during the execution part) and then close on completion. Wdyt?

roeierez:
How do you see the API from the user perspective?

hydra-yse:
@roeierez I'm thinking of the usual prepare/execute flow. The user calls prepare and gets fees, if they accept they execute the payment (ensuring nothing has changed). So we could have it live for the whole prepare step, and then for the whole execute step.

roeierez:
@hydra-yse approach sounds good to me and OK to move on with the implementation.

lib/core/src/side_swap/api/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+            Ok(res) => anyhow!("Received invalid response from the server: {res:?}"),
+            Err(err) => anyhow!("Received error response from the server: {err:?}"),
+        }
+    }
+}
+
+#[sdk_macros::async_trait]
+impl SideSwapService for HybridSideSwapService {
+    async fn start(self: Arc<Self>) -> Result<()> {
+        let ws = connect(&self.url).await?;
```
dangeross, line 146: Shouldn't we check if it's already started?

lib/core/src/side_swap/api/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+        let keep_alive_ping_interval = Duration::from_secs(15);
+
+        let cloned = Arc::clone(&self);
+        tokio::spawn(async move {
+            let mut shutdown = cloned.shutdown.receiver.lock().await;
+            loop {
+                tokio::select! {
+                    _ = shutdown.recv() => {
+                        info!("Received shutdown signal, exiting SideSwap service loop");
+                        cloned.set_started(false).await;
```
dangeross, line 159: Just to check, the `ongoing_swap` should be kept set if the service is stopped?

dangeross (commented):
Looking good

lib/core/src/side_swap/api/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+use crate::elements::{self, pset::PartiallySignedTransaction, AssetId, Txid};
+use crate::model::AssetSwap;
+use crate::wallet::OnchainWallet;
+
+mod request_handler;
+mod response_handler;
+
+pub const SIDESWAP_MAINNET_URL: &str = "wss://api.sideswap.io/json-rpc-ws";
+pub const SIDESWAP_TESTNET_URL: &str = "wss://api-testnet.sideswap.io/json-rpc-ws";
+pub const SIDESWAP_REGTEST_URL: &str = "wss://api-regtest.sideswap.io/json-rpc-ws";
```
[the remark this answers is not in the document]
hydra-yse: No, it seems they have their own regtest network. Probably unnecessary to specify it then.

lib/core/src/side_swap/api/mod.rs
```diff
@@ -0,0 +1,422 @@
...
+                asset: wallet_utxo.unblinded.asset,
+                asset_bf: wallet_utxo.unblinded.asset_bf,
+                value: wallet_utxo.unblinded.value,
+                value_bf: wallet_utxo.unblinded.value_bf,
+                redeem_script: None,
+            });
+            if send_value >= start_res.send_amount {
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
hydra-yse: Yes, I've actually taken inspiration from the one in payjoin. I'll see if I can reuse it here.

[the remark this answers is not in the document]
dangeross: I have PR #874 that exposes UTXO selection more, so that it can be shared between the wallet and payjoin. It would probably help here too (once merged)

lib/core/src/side_swap/api/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+            )
+            .await?;
+
+        if !reqwest::StatusCode::from_u16(status)?.is_success() {
+            bail!("Received error status code when executing swap: {status}, body: {raw_body}");
+        }
+
+        let HttpResponse::SwapStart(response) = serde_json::from_str(&raw_body)? else {
+            bail!("Expected `SwapStart` response from server, received: {raw_body}");
+        };
```
dangeross, line 271: It might be nice to add a json fn to the RestClient to handle this serialize/deserialize pattern similar to the [payjoin post_request fn](https://github.com/breez/breez-sdk-liquid/blob/f48f71fd72acad3cd82c775ff99091292eb45c79/lib/core/src/payjoin/side_swap.rs#L88-L104)

Just thinking out loud, not requesting it now

dangeross (approved):
Looks good

lib/core/src/side_swap/api/mod.rs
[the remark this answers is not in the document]
hydra-yse: Sure thing, maybe I can make another PR on top to address this once merged?

PR #902, opened 2025-04-29
Notification Plugin: Add Cache-Control header for `lnurlpay_info` response

dangeross:
This PR sets the `Cache-Control` header to `max-age=86400` for LNURL-Pay info requests. This will allow the LNURL-service to cache this static information to minimize the requests back/forth to the device when receiving via LNURL-Pay.


lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/LnurlPayVerify.kt
```diff
@@ -75,7 +75,8 @@ class LnurlPayVerifyJob(
                 throw InvalidLnurlPayException("Not found")
             }
 
-            val success = replyServer(Json.encodeToString(response), request.replyURL)
+            val maxAge = if (response.settled) 604800 else 0
```
erdemyerebasmaz, line 78: Should we have day/week constants for these values?

erdemyerebasmaz (approved):

erdemyerebasmaz (approved):

roeierez (approved):

PR #907, opened 2025-05-01
BOLT12 receive updates

dangeross:
This PR:
- Sends the BOLT12 invoice request error immediately back via websocket or webhook response
- Removes setting the BOLT12 invoice blinded payment path to the MRH short channel id
- Uses the MRH returned alongside fetching the BOLT12 invoice to send directly via Liquid tx
- Optimizes requests to Boltz during invoice creation (remove request to get CLN node public key, parellelize requests to get BOLT12 params and reverse swap pairs)

**Testing notes:**
- When paying under the minimum receive amount (< 100sat) from Phoenix, you should receive a payment failure error faster. Test Misty when both in foreground and background

Builds:
- Android 432: https://github.com/breez/misty-breez/actions/runs/14836581663
- iOS TestFlight 6418.1: https://github.com/breez/misty-breez/actions/runs/14836943107

### DIFF
--- a/.github/workflows/build-bindings-android.yml
+++ b/.github/workflows/build-bindings-android.yml
@@ -177,7 +177,7 @@ jobs:
     - setup
     - build-dummies
     runs-on: ubuntu-latest
-    name: build jniLibs dummy ${{ matrix.target }}${{ matrix.uniffi }}
+    name: build jniLibs dummy ${{ matrix.uniffi }}
     strategy:
       matrix:
         uniffi: ${{ fromJson(needs.setup.outputs.uniffi-matrix) }}
--- a/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/InvoiceRequest.kt
+++ b/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/InvoiceRequest.kt
@@ -65,30 +65,33 @@ class InvoiceRequestJob(
                     if (success) DEFAULT_INVOICE_REQUEST_NOTIFICATION_TITLE else DEFAULT_INVOICE_REQUEST_NOTIFICATION_FAILURE_TITLE,
                 ),
             )
+        } catch (e: PaymentException) {
+            logger.log(TAG, "Failed to process invoice request: ${e.message}", "WARN")
+            notifyError(request, e.message ?: "Failed to create invoice")
         } catch (e: Exception) {
             logger.log(TAG, "Failed to process invoice request: ${e.message}", "WARN")
-            if (request != null) {
-                val error = when (e) {
-                    is PaymentException.AmountOutOfRange -> e.message ?: "Amount out of range"
-                    is PaymentException.AmountMissing -> "Amount missing in invoice request"
-                    else -> "Failed to create invoice"
-                }
-                val response = InvoiceErrorResponse(error)
-                replyServer(Json.encodeToString(response), request.replyURL)
-            }
-            notifyChannel(
-                context,
-                NOTIFICATION_CHANNEL_REPLACEABLE,
-                getString(
-                    context,
-                    INVOICE_REQUEST_NOTIFICATION_FAILURE_TITLE,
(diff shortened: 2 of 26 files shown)


lib/core/src/sdk.rs
```diff
@@ -1621,12 +1616,12 @@ impl LiquidSdk {
                 bip353_address,
             } => {
                 let fees_sat = fees_sat.ok_or(PaymentError::InsufficientFunds)?;
-                let bolt12_invoice = self
+                let bolt12_info = self
                     .swapper
                     .get_bolt12_invoice(&offer.offer, *receiver_amount_sat)
```
danielgranhao, line 1621: Nit: should we rename the method to e.g. `get_bolt12_info`? I'm also fine with keeping as is

.github/workflows/build-bindings-android.yml
```diff
@@ -177,7 +177,7 @@ jobs:
     - setup
     - build-dummies
     runs-on: ubuntu-latest
-    name: build jniLibs dummy ${{ matrix.target }}${{ matrix.uniffi }}
+    name: build jniLibs dummy ${{ matrix.uniffi }}
```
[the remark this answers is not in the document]
dangeross: Yes, there is no matrix target strategy here

lib/core/src/swapper/boltz/status_stream.rs
```diff
@@ -195,6 +195,16 @@ impl<P: ProxyUrlFetcher> SwapperStatusStream for BoltzSwapper<P> {
...
     }
 
+    fn send_invoice_error(&self, id: &str, error: &str) -> Result<()> {
+        let _ = self
+            .request_notifier
+            .send(WsRequest::InvoiceError(boltz::InvoiceError {
+                id: id.to_string(),
+                error: error.to_string(),
+            }));
+        Ok(())
```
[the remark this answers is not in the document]
dangeross: Fixed [2de8ff1](https://github.com/breez/breez-sdk-liquid/pull/907/commits/2de8ff19a09d1c8b36b93413b86638d8b2938f40)

lib/core/src/sdk.rs
dangeross: Fixed [2de8ff1](https://github.com/breez/breez-sdk-liquid/pull/907/commits/2de8ff19a09d1c8b36b93413b86638d8b2938f40)

danielgranhao (approved):

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/InvoiceRequest.kt
```diff
@@ -59,22 +65,34 @@ class InvoiceRequestJob(
                     if (success) DEFAULT_INVOICE_REQUEST_NOTIFICATION_TITLE else DEFAULT_INVOICE_REQUEST_NOTIFICATION_FAILURE_TITLE,
                 ),
             )
+        } catch (e: PaymentException) {
+            logger.log(TAG, "Failed to process invoice request: ${e.message}", "WARN")
```
roeierez, line 69: IIUC we handle the bolt12 invoice request both on the plugin and in the sdk core.
Does this mean when the app is opened it is likely we create two invoices? (two mrh addresses, two reverse swaps to subscribe, etc..).

roeierez (commented):
Looks good. One concern.

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/InvoiceRequest.kt
dangeross: If the app is already open (there is a WS connection and we are subscribed to the offer invoice requests) then Boltz only sends the invoice request to the WS (no webhook request). If there is no WS subscription, then it is only sent via webhook. If there are two instances active and both have subscribed to the offer invoice requests, Boltz only keeps the latest connection (last subscription gets the invoice request).

roeierez (approved):
LGTM

PR #910, opened 2025-05-05
Fix ignored payment events on failed wallet sync

hydra-yse:
This PR:
- Increases the `NETWORK_PROPAGATION_GRACE_PERIOD` to 1 minute, to allow unconfirmed Send swaps to complete before deleting their associated `PaymentTxData`
- Aborts the `sync_inner` method when the wallet scan fails, so that operations that depend on the scan (e.g. recovery) won't execute unpredictably


roeierez (approved):
LGTM

danielgranhao (approved):
LGTM

Issue #852, opened 2025-04-02
Support LNURL-Verify


kingonly:
https://github.com/lnurl/luds/issues/182

dangeross:
I've mostly implemented this @kilrau, but I have an open question to you or Michael about payment verification if the MRH route is taken instead of Lightning. Maybe I can catch you later to talk about it?

kilrau:
Awesome! How can we help to push this over the finish line?

john-zaprite:
Sounds great. Let me know if we can help with anything on our side.

PR #911, opened 2025-05-08
Add LNURL-verify notification plugin tasks

dangeross:
This PR:
- Updates the LNURL Invoice task to fill the verify URL with the payment hash if its set in the request
- Adds the LNURL Verify task to get the payment and check if its settled.

LUD-21: https://github.com/lnurl/luds/blob/luds/21.md

LNURL service PR: https://github.com/breez/breez-lnurl/pull/20
Notify service PR: https://github.com/breez/notify/pull/17

Fixes: #852


roeierez:
@dangeross  I wonder if the preimage should be registered at the lnurl service after successful payment or cached after the first request to save the notification cycle and enhance reliability.
Ok to go with the current approach first.

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/LnurlPayVerify.kt
```diff
@@ -0,0 +1,105 @@
...
+            val getPaymentReq = GetPaymentRequest.PaymentHash(request.paymentHash)
+            val payment = liquidSDK.getPayment(getPaymentReq) ?: run {
+                throw InvalidLnurlPayException("Not found")
+            }
+            val response = when (val details = payment.details) {
+                is PaymentDetails.Lightning -> {
+                    // In the case of a Lightning payment, if it's paid via Lightning or MRH,
+                    // we can release the preimage
+                    val settled = when (payment.status) {
+                        PaymentState.COMPLETE, PaymentState.PENDING -> details.claimTxId != null
```
roeierez, line 61: I guess this means in case of mrh we only consider complete payments (after confirmation)?

roeierez (approved):

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/LnurlPayVerify.kt
dangeross: This actually doesn't take MRH into consideration at all because there will be no `claim_tx_id`. Need to fix this

dangeross: I made a fix to use the payment `tx_id` to determine if the `claim_tx_id` or `mrh_tx_id` is set. If it is set then it can be settled in Pending or Complete. If the `tx_id` is not set then it's Pending because we have the `lockup_tx_id`.

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/LnurlPayVerify.kt
```diff
@@ -0,0 +1,106 @@
...
+            val payment = liquidSDK.getPayment(getPaymentReq) ?: run {
+                throw InvalidLnurlPayException("Not found")
+            }
+            val response = when (val details = payment.details) {
+                is PaymentDetails.Lightning -> {
+                    // In the case of a Lightning payment, if it's paid via Lightning or MRH,
+                    // we can release the preimage
+                    val settled = when (payment.status) {
+                        // The payment txId is set when there is a claimTxId or mrhTxId
+                        PaymentState.COMPLETE, PaymentState.PENDING -> payment.txId != null
```
roeierez, line 62: For MRH shouldn't we wait for confirmation?
Also I wonder if releasing the preimage on mrh has some security issues.

roeierez (changes requested):

danielgranhao (approved):
LGTM

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/LnurlPayVerify.kt
kilrau: Good point: it confirms a Lightning invoice was paid that technically was never paid as things were settled on Liquid. Your take? @michael1011

michael1011: > For MRH shouldn't we wait for confirmation?

Depends on if you trust the sender of the tx to not double spend you

> Also I wonder if releasing the preimage on mrh has some security issues.

No security issues that I am aware of. The HTLCs are not being used when the payment is done via the MRH

PR #912, opened 2025-05-12
Persist wallet cache in db

danielgranhao:
This PR moves wallet caching from the fs to the existing sqlite db. The goal is to prevent corruption issues and allow the notification extension to share the cache with the main instance. Additionally, the Wasm project now doesn't need to care about wallet cache persistence.


roeierez:
@danielgranhao that looks really good!
Question: With the sqlite persister do we have a case where we clear the wallet as before? It seems now that wrong updates are just skipped. could you elaborate how these skipped updates are going to affect the background jobs? I am thinking if that error doesn't stop the job work we should be good as in that case the wallet should be in sync (conflict means the app is updating at that time). Interesting in your thoughts regarding these cases.

danielgranhao:
> With the sqlite persister do we have a case where we clear the wallet as before? It seems now that wrong updates are just skipped. could you elaborate how these skipped updates are going to affect the background jobs? I am thinking if that error doesn't stop the job work we should be good as in that case the wallet should be in sync (conflict means the app is updating at that time). Interested in your thoughts regarding these cases.

@roeierez Yes, my goal was for failed updates (due to conflicts) to be skipped without an error, so background jobs won't be affected. When persisting fails, the lwk wollet still has the correct state in memory, so it shouldn't be affected. 

I'm not sure yet if we need the clearing function. I think this implementation isn't able to prevent all possible issues. For example:
- Instance A is running and has just persisted update 10
- Instance B starts, reads update 10
- Instance A persists update 11
- Instance B overwrites update 11
- Instance A persists update 12

This wouldn't be prevented and would result in a potentially inconsistent update sequence that can cause a failure on the next start of the SDK (when the updates are read). The clearing logic would catch these potential issues and resolve them by syncing from scratch.

I plan to test this now to figure out if these issues can happen and, if so, with which frequency. But even if I can't trigger any inconsistencies, maybe it's a good idea to have the clearing safeguard anyway?

roeierez:
@danielgranhao  The lwk wallet right now locks the whole wallet for the time of sync so two batches of updates from two clients can't interfere with each other.
Now that we have the same persistent storage the case you mentioned may happen and I think the only way to prevent it is to update the batch in a transaction (currently we have a transaction for each single record).
I don't know if it is possible currently with lwk wallet but worth looking at this.

danielgranhao:
@roeierez I've adjusted the PR. As discussed out-of-band, we dropped update merging as it led to a significantly more complex conflict detection logic. More updates results in the wallet load to be less efficient, but this is unlikely to be significant. If it turns out to be an issue, we can implement improvements later.

Also, now, when we detect a conflict on update push, we simply reload the wallet instead of giving up on persisting updates.

roeierez (approved):
LGTM!  Great to have all platforms using the same persister.

hydra-yse (approved):
utACK

dangeross (commented):
Looks good. One q

dangeross (approved):
LGTM

PR #915, opened 2025-05-15
Side-Swap: SDK Integration

hydra-yse:
Built on top of #897.

## Description
This PR:
- Adds the methods `prepare_asset_swap`/`execute_asset_swap` and the corresponding interfaces to allow the user to swap funds via the SideSwap service.
- Adds a cli command to allow for testing of the swapping service.

## To-dos
- [ ] Fix `RequestId` issue when sending messages (currently server returns error) 
- [ ] Limit the scope of the tradeable assets (we just want USDt for now)
- [ ] Regenerate bindings

### DIFF
--- a/cli/src/commands.rs
+++ b/cli/src/commands.rs
@@ -258,6 +258,11 @@ pub(crate) enum Command {
     ListFiat {},
     /// Fetch available fiat rates
     FetchFiatRates {},
+    /// Swap L-BTC for any SideSwap asset
+    SwapAsset {
+        asset_id: String,
+        send_amount_sat: u64,
+    },
 }
 
 #[derive(Helper, Completer, Hinter, Validator)]
@@ -811,6 +816,29 @@ pub(crate) async fn handle_command(
         }
         Command::ListFiat {} => {
             let res = sdk.list_fiat_currencies().await?;
+            command_result!(res)
+        }
+        Command::SwapAsset {
+            asset_id,
+            send_amount_sat,
+        } => {
+            let prepare_response = sdk
+                .prepare_asset_swap(&PrepareAssetSwapRequest {
+                    asset_id,
+                    send_amount_sat,
+                })
+                .await?;
+
+            let asset_swap = &prepare_response.asset_swap;
+            let confirmation_msg = format!(
+                "Fees: approx {} sat. Recv amount: {}. Is this acceptable? (y/N) ",
+                asset_swap.fees_sat, asset_swap.recv_amount
+            );
+            wait_confirmation!(confirmation_msg, "Asset swap halted");
+
+            let res = sdk
+                .execute_asset_swap(&ExecuteAssetSwapRequest { prepare_response })
+                .await?;
             command_result!(res)
         }
     })
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -746,6 +746,14 @@ dictionary PrepareAssetSwapResponse {
     AssetSwap asset_swap;
 };
 
+dictionary ExecuteAssetSwapRequest {
+    PrepareAssetSwapResponse prepare_response;
+};
+
+dictionary ExecuteAssetSwapResponse {
+    Payment payment;
+};
+
 namespace breez_sdk_liquid {
     [Throws=SdkError]
     BindingLiquidSdk connect(ConnectRequest req);
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -737,6 +737,15 @@ dictionary AssetSwap {
     u64 fees_sat;
 };
 
+dictionary PrepareAssetSwapRequest {
+    string asset_id;
+    u64 send_amount_sat;
+};
+
+dictionary PrepareAssetSwapResponse {
+    AssetSwap asset_swap;
+};
+
 namespace breez_sdk_liquid {
     [Throws=SdkError]
     BindingLiquidSdk connect(ConnectRequest req);
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -24,7 +24,6 @@ use std::str::FromStr;
 use strum_macros::{Display, EnumString};
 
 use crate::receive_swap::DEFAULT_ZERO_CONF_MAX_SAT;
-use crate::utils;
 use crate::{
     bitcoin,
     chain::{bitcoin::BitcoinChainService, liquid::LiquidChainService},
@@ -34,6 +33,10 @@ use crate::{
 use crate::{
     chain::bitcoin::esplora::EsploraBitcoinChainService,
     chain::liquid::esplora::EsploraLiquidChainService, prelude::DEFAULT_EXTERNAL_INPUT_PARSERS,
+};
+use crate::{
+    side_swap::api::{SIDESWAP_MAINNET_URL, SIDESWAP_REGTEST_URL, SIDESWAP_TESTNET_URL},
+    utils,
 };
 
 // Uses f64 for the maximum precision when converting between units
@@ -349,6 +352,14 @@ impl Config {
             &electrum_url,
             lwk_wollet::ElectrumOptions { timeout: Some(3) },
         )
+    }
+
+    pub(crate) fn sideswap_url(&self) -> &'static str {
+        match self.network {
+            LiquidNetwork::Mainnet => SIDESWAP_MAINNET_URL,
+            LiquidNetwork::Testnet => SIDESWAP_TESTNET_URL,
+            LiquidNetwork::Regtest => SIDESWAP_REGTEST_URL,
+        }
     }
 }
 
@@ -2445,6 +2456,19 @@ impl TryFrom<&sideswap_api::SubscribePriceStreamResponse> for AssetSwap {
                 .context("Expected price when creating side swap")?,
         })
     }
+}
+
+/// An argument when calling [crate::sdk::LiquidSdk::prepare_asset_swap_request].
+#[derive(Debug, Clone, Serialize)]
+pub struct PrepareAssetSwapRequest {
+    pub asset_id: String,
+    pub send_amount_sat: u64,
+}
+
+/// Returned when calling [crate::sdk::LiquidSdk::prepare_asset_swap_request].
+#[derive(Debug, Clone, Serialize)]
+pub struct PrepareAssetSwapResponse {
+    pub asset_swap: AssetSwap,
 }
 
 #[derive(Clone, Debug)]
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -2471,6 +2471,18 @@ pub struct PrepareAssetSwapResponse {
     pub asset_swap: AssetSwap,
 }
 
+/// An argument when calling [crate::sdk::LiquidSdk::execute_asset_swap_request].
+#[derive(Debug, Clone, Serialize)]
+pub struct ExecuteAssetSwapRequest {
+    pub prepare_response: PrepareAssetSwapResponse,
+}
+
+/// Returned when calling [crate::sdk::LiquidSdk::execute_asset_swap_request].
+#[derive(Debug, Clone, Serialize)]
+pub struct ExecuteAssetSwapResponse {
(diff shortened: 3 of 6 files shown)


lib/core/src/sdk.rs
```diff
@@ -4031,6 +4050,115 @@ impl LiquidSdk {
...
     }
+
+    pub async fn prepare_asset_swap(
+        &self,
+        req: &PrepareAssetSwapRequest,
+    ) -> Result<PrepareAssetSwapResponse, PaymentError> {
+        self.sideswap_service.clone().start().await?;
+
+        ensure_sdk!(
+            self.get_info().await?.wallet_info.balance_sat > req.send_amount_sat,
```
dangeross, line 4061: Do we need to take into consideration the fee also?

lib/core/src/sdk.rs
```diff
@@ -4031,6 +4050,115 @@ impl LiquidSdk {
...
+        let tx_data = PaymentTxData {
+            tx_id: tx_id.clone(),
+            timestamp: Some(utils::now()),
+            amount: req_swap.recv_amount,
+            fees_sat: ongoing_swap.fees_sat,
+            payment_type: PaymentType::Send,
+            is_confirmed: false,
+            unblinding_data: None,
+            asset_id: req_swap.asset_id.clone(),
+        };
```
dangeross, line 4127: Maybe I misunderstand, so if we're sending shouldn't the amount be the send_amount_sat and L-BTC asset? It's not really clear to me because I don't know what we'll see after the wallet scan

lib/core/src/sdk.rs
```diff
@@ -4031,6 +4050,115 @@ impl LiquidSdk {
...
+        )?;
+        self.emit_payment_updated(Some(tx_id)).await?; // Emit Pending event
+
+        let asset_info = self
+            .persister
+            .get_asset_metadata(&req_swap.asset_id)?
+            .map(|ref am| AssetInfo {
+                name: am.name.clone(),
+                ticker: am.ticker.clone(),
+                amount: am.amount_from_sat(req_swap.send_amount_sat),
```
dangeross, line 4146: Isn't `send_amount_sat` the L-BTC we're sending? So this should be `req.swap.recv_amount` the asset amount in satoshi?

dangeross (commented):
From the interface point of view, it's not clear to me the direction of the swap. I think the direction needs to be more explicit in the naming. I would really prefer if it could be integrated into `prepare_send_payment` / `send_payment` but I guess it's difficult.

> Limit the scope of the tradeable assets (we just want USDt for now)
Maybe for this you can query if the asset id is in the asset_metadata table, if not then reject the request. By default only L-BTC and USDT are in there

lib/core/src/sdk.rs
```diff
@@ -4568,6 +4587,161 @@ impl LiquidSdk {
...
+    ///     * `payment` - the onchain [Payment] details of the swap, including the Liquid txid and
+    ///     amount
+    ///
+    /// # Errors
+    ///
+    /// * [PaymentError::InvalidOrExpiredFees] - returned if the swap fees have changed above the
+    ///   previously accepted amount
+    /// * [PaymentError::InsufficientFunds] - returned if you do not have enough funds to execute
+    ///   the payment
+    pub async fn execute_asset_swap(
```
danielgranhao, line 4653: Like I already mentioned in the other PR, I think here the name could better fit the existing naming scheme.

And the prepare one: `prepare_swap_asset`

```suggestion
    pub async fn swap_asset(
```

danielgranhao (commented):
Looking good

lib/core/src/sdk.rs
```diff
@@ -4568,6 +4587,161 @@ impl LiquidSdk {
...
+    /// # Returns
+    ///
+    /// A [PrepareAssetSwapResponse] containing:
+    ///     * `asset_swap` - the [AssetSwap] details of the ongoing swap, including the exchange
+    ///     rate, fees and the final amount you will receive
+    pub async fn prepare_asset_swap(
+        &self,
+        req: &PrepareAssetSwapRequest,
+    ) -> Result<PrepareAssetSwapResponse, PaymentError> {
+        self.sideswap_service.clone().start().await?;
```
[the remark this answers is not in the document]
roeierez: @hydra-yse  if two requests are done in parallel to the prepare_swap_asset then the first to complete will shutdown the service before the second one is completed.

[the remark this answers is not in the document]
hydra-yse: > Well it is being shut down also on the regular case. I am saying the start/stop should be used in a way that it is very clear and predictable when the service is running and when not.

I agree with this. It's better to just subscribe/unsubscribe then.

lib/core/src/sdk.rs
```diff
@@ -4568,6 +4587,161 @@ impl LiquidSdk {
...
+        self.sideswap_service.stop().await;
+
+        // We insert a pseudo-tx in case LWK fails to pick up the new mempool tx for a while
+        // This makes the tx known to the SDK (get_info, list_payments) instantly
+        let tx_data = PaymentTxData {
+            tx_id: tx_id.clone(),
+            timestamp: Some(utils::now()),
+            amount: req_swap.payer_amount_sat,
+            fees_sat: ongoing_swap.fees_sat,
+            payment_type: PaymentType::Send,
```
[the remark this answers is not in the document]
hydra-yse: I think, for the sake of simplicity, that we should just mark this as `Receive` of that specific asset amount. Wdyt @roeierez ?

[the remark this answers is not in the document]
roeierez: It seems a bit more complicated than that as we also send lbtc right?
Initially I thought perhaps we need here two payments. send for lbtc and receive for asset but then I realized it is the same transaction id that is set as unique on our PaymentTxData table.
One option is to add change that and set the (tx_id, asset_id) as unique in the PaymentTxData table instead. 
Let's discuss it.

lib/core/src/sdk.rs
```diff
@@ -1324,33 +1326,71 @@ impl LiquidSdk {
...
+                                (asset_id, receiver_amount_sat, fees_sat, asset_fees)
+                            }
+                            // If not, try and prepare the payment via SideSwap
+                            false => {
+                                let asset_id = AssetId::from_str(&asset_id)?;
+                                let sideswap_service = SideSwapService::from_sdk(self);
+
+                                let swap = sideswap_service
+                                    .get_asset_swap(asset_id, receiver_amount_sat)
+                                    .await?;
```
roeierez, line 1375: Curious how much time roughly starting and querying the swap service takes?

danielgranhao (commented):
Looks good! Just a small comment about making sure we stop the service on all paths

cli/src/commands.rs
```diff
@@ -64,6 +64,12 @@ pub(crate) enum Command {
         /// Delay for the send, in seconds
         #[arg(long)]
         delay: Option<u64>,
+
+        /// Can only be used when a BIP21 address with an amount is provided
```
dangeross, line 68: ~A normal asset send can be done without a BIP21 address, why not this also?~ Seems this is also works with an asset_id and amount provided

lib/core/src/sdk.rs
```diff
@@ -1924,6 +1999,76 @@ impl LiquidSdk {
...
+            .get_asset_swap(AssetId::from_str(&asset_id)?, receiver_amount_sat)
+            .await?;
+
+        ensure_sdk!(
+            swap.fees_sat <= fees_sat,
+            PaymentError::InvalidOrExpiredFees
+        );
+
+        ensure_sdk!(
+            self.get_info().await?.wallet_info.balance_sat >= swap.payer_amount_sat,
```
dangeross, line 2025: Should it be swap.payer_amount_sat + swap.fees_sat?

lib/core/src/model.rs
```diff
@@ -787,6 +787,9 @@ pub enum PayAmount {
         asset_id: String,
         receiver_amount: f64,
         estimate_asset_fees: Option<bool>,
+        /// Specifies whether or not to use the wallet's funds to execute the payment. Defaults to true.
+        /// If false, it will try swapping the asset via the [Side Swap Service](crate::side_swap::api::SideSwapService)
+        use_wallet_funds: Option<bool>,
```
dangeross, line 792: Could we change this to `use_assets` or `use_wallet_assets`? To me `use_wallet_funds` doesn't make 100% sense as L-BTC is also wallet funds

lib/core/src/sdk.rs
```diff
@@ -1924,6 +1999,76 @@ impl LiquidSdk {
...
+                ..Default::default()
+            }),
+            false,
+        )?;
+        self.emit_payment_updated(Some(tx_id.clone())).await?; // Emit Pending event
+
+        let payment = self
+            .persister
+            .get_payment(&tx_id)?
+            .context("Expected payment to be present in the database")?;
```
dangeross, line 2068: For consistency
```suggestion
            .context("Payment not found")?;
```

lib/core/src/sdk.rs
```diff
@@ -1924,6 +1999,76 @@ impl LiquidSdk {
...
+                    asset_id: utils::lbtc_asset_id(self.config.network).to_string(),
+                    amount: swap.payer_amount_sat,
+                    payment_type: PaymentType::Send,
+                },
+                PaymentTxBalance {
+                    asset_id: asset_id.to_string(),
+                    amount: swap.receiver_amount_sat,
+                    payment_type: PaymentType::Receive,
+                },
+            ],
```
dangeross, line 2054: I'm a bit lost here. Why do we assume we are also receiving the payment? The interface suggests it can be sent to any address...

lib/core/src/sdk.rs
```diff
@@ -1924,6 +1999,76 @@ impl LiquidSdk {
...
+            Some(PaymentTxDetails {
+                tx_id: tx_id.clone(),
+                destination: address_data.address,
+                description: Some("SideSwap transfer".to_string()),
```
dangeross, line 2058: The description should come from the `address_data.message`

lib/core/src/sdk.rs
hydra-yse: The payer amount includes receiver amount + fees. I'll document that better

lib/core/src/sdk.rs
```diff
@@ -1324,33 +1328,68 @@ impl LiquidSdk {
...
+                            }
+                            false => {
+                                let start = Instant::now();
+                                let asset_id = AssetId::from_str(&asset_id)?;
+                                let sideswap_service = SideSwapService::from_sdk(self);
+
+                                let swap = sideswap_service
+                                    .get_asset_swap(asset_id, receiver_amount_sat)
+                                    .map_err(|err| {
+                                        sideswap_service.stop();
```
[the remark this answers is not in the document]
hydra-yse: I had thought of that, but unfortunately the service lives under an `Arc`, so `Drop` can only be called once all references go out of scope. In this case, one Arc always lives in the ws stream loop, so it would never occur unless called manually. Perhaps we can think about better approaches, but I couldn't find anything straightforward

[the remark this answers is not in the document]
hydra-yse: 6b223d5ba0d4bffe10f66cfae6a1d210a758db5b should do it :+1:

dangeross (approved):
LGTM. Seems like the Flutter codegen needs rerunning to update the comments

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -527,7 +527,7 @@ dictionary OnchainPaymentLimitsResponse {
 [Enum]
 interface PayAmount {
     Bitcoin(u64 receiver_amount_sat);
-    Asset(string asset_id, f64 receiver_amount, boolean? estimate_asset_fees);
+    Asset(string asset_id, f64 receiver_amount, boolean? estimate_asset_fees, boolean? use_wallet_assets);
```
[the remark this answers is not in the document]
hydra-yse: I didn't want to expose the term "swap" to the user, like `always_swap` or `use_swap`. First thing that comes to mind would be something like `avoid_wallet_assets` to default it to false. Does that sound better?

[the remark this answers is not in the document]
hydra-yse: Sounds good :+1:

PR #921, opened 2025-05-16
Allow get_payment to get payment no matter the state

### LINKED ISSUE #919: LNURL-verify returns "Not found" if the payment isn't yet initiated

dangeross:
This PR updates the `get_payment` query to always return a payment, even when in the Created or Failed states. This then allows LNURL-verify to correctly query the payment and not return an error before the payment is initiated.

**Builds:**
- Android: https://github.com/breez/misty-breez/actions/runs/15071873025
- iOS 6429.1: https://github.com/breez/misty-breez/actions/runs/15071870599 

Fixes #919

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -485,7 +485,21 @@ impl Persister {
         offset: Option<u32>,
         limit: Option<u32>,
         sort_ascending: Option<bool>,
+        include_all_states: Option<bool>,
     ) -> String {
+        let (where_receive_swap_clause, where_chain_swap_clause) = if include_all_states
+            .unwrap_or_default()
+        {
+            ("true", "true")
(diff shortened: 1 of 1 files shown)


roeierez:
@dangeross Just verifying that we really need to do this change. Can't the lnurl service detect that the payment doesn't exist and return not verified? Or it needs the actual payment object anyway?

dangeross:
> @dangeross Just verifying that we really need to do this change. Can't the lnurl service detect that the payment doesn't exist and return not verified? Or it needs the actual payment object anyway?

I spoke with John from zaprite, we are being inconsistent with the spec. If the invoice is created we should be returning settled=false, we are currently returning "not found" indicating the invoice doesn't exist.

roeierez (approved):
LGTM

PR #923, opened 2025-05-19
Rework draining to enable BOLT12 drain

### LINKED ISSUE #922: Cannot drain using BOLT12
Misty issue: https://github.com/breez/misty-breez/issues/544

dangeross:
Enables BOLT12 drain by inverse calculating the receiver amount the same way as LNURL-pay. We also add the `PayAmount` to the prepare responses to allow them to be used in when sending to determine if the request was a drain.

**Testing notes:**
- Re-test draining via LNURL/lightning address to SDK/non-SDK devices
- Test draining via BOLT12 Offer/BIP535 address to SDK/non-SDK devices

Fixes #922 
Fixes https://github.com/breez/misty-breez/issues/544

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -1219,6 +1219,7 @@ impl LiquidSdk {
     /// # Returns
     /// Returns a [PrepareSendResponse] containing:
     ///     * `destination` - the parsed destination, of type [SendDestination]
+    ///     * `amount` - the optional [PayAmount] to be sent in either Bitcoin or another asset
     ///     * `fees_sat` - the optional estimated fee in satoshi. Is set when there is Bitcoin
     ///        available to pay fees. When not set, there are asset fees available to pay fees.
     ///     * `estimated_asset_fees` - the optional estimated fee in the asset. Is set when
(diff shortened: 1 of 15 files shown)


lib/core/src/sdk.rs
```diff
@@ -1399,76 +1400,105 @@ impl LiquidSdk {
...
-                            // The BOLT11 invoice has no MRH, so we calculate the fees using a swap
-                            let boltz_fees_total = lbtc_pair.fees.total(invoice_amount_sat);
-                            let user_lockup_amount_sat = invoice_amount_sat + boltz_fees_total;
-                            let lockup_fees_sat = self
-                                .estimate_lockup_tx_or_drain_tx_fee(user_lockup_amount_sat)
-                                .await?;
-                            let fees_sat = boltz_fees_total + lockup_fees_sat;
-                            (
+                (receiver_amount_sat, fees_sat) = match (mrh_address.clone(), req.amount.clone()) {
+                    (Some(lbtc_address), Some(PayAmount::Drain)) => {
```
roeierez, line 1404: I am not sure we should support drain with bolt11. It seems just something the implementation let us do atm but needs to be abstracted. Is there a specific case we are targeting here or just because we can?

roeierez (changes requested):
Looks good. One comment regarding bolt11 drain.

lib/core/src/sdk.rs
dangeross: It's to handle the LNURL-pay case, where we receive an invoice but we intend to drain to the MRH address

hydra-yse (approved):
utACK

roeierez (approved):

PR #927, opened 2025-05-20
Update Kotlin Multiplatform bindgen to uniffi 0.28

dangeross:


dangeross:
Looks like we're hitting this issue again:
```
Error: Call requires API level 33 (current min is 24): java.lang.ref.Cleaner#create [NewApi]
    val cleaner = java.lang.ref.Cleaner.create()
                                        ~~~~~~
```

erdemyerebasmaz:
> Looks like we're hitting this issue again:
> 
> ```
> Error: Call requires API level 33 (current min is 24): java.lang.ref.Cleaner#create [NewApi]
>     val cleaner = java.lang.ref.Cleaner.create()
>                                         ~~~~~~
> ```

Yes and the option to disable Java cleaner is only available on 0.29+.
[The UniFFI user guide - Kotlin - Configuration](https://mozilla.github.io/uniffi-rs/0.29/kotlin/configuration.html)

dangeross:
> Yes and the option to disable Java cleaner is only available on 0.29+.

I think there are two options:
- Upgrade only Kotlin to uniffi 0.29
- Use the Kotlin Multiplatform bindgen to generate the Kotlin bindings also, as it has the [option to disable the java cleaner](https://gobley.dev/docs/bindgen#bindgen-configuration)

erdemyerebasmaz:
> > Yes and the option to disable Java cleaner is only available on 0.29+.
> 
> I think there are two options:
> 
> * Upgrade only Kotlin to uniffi 0.29
> * Use the Kotlin Multiplatform bindgen to generate the Kotlin bindings also, as it has the [option to disable the java cleaner](https://gobley.dev/docs/bindgen#bindgen-configuration)

I'm leaning more towards the first option:
- Upgrade only Kotlin to uniffi 0.29

instead of managing another bindgen. Won't we eventually upgrade to 0.29 for other bindings as well?

roeierez (approved):
LGTM

packages/react-native/example/android/build.gradle
```diff
@@ -4,9 +4,9 @@ buildscript {
     ext {
         buildToolsVersion = "33.0.0"
         minSdkVersion = 24
-        compileSdkVersion = 34
-        targetSdkVersion = 34
-        kotlin_version = "1.8.0"
+        compileSdkVersion = 33
+        targetSdkVersion = 33
```
erdemyerebasmaz, line 8: Why were these versions got downgraded? Would it be better to keep them at 34 because of:
> Starting August 31 2024:
> 
> New apps and app updates must target Android 14 (API level 34) or higher to be submitted to Google Play; except for Wear OS and Android TV apps, which must target Android 13 (API level 33) or higher.

and their new rule of:
image

erdemyerebasmaz (changes requested):
Looks good! 

Have one comment regarding target API level requirements.

packages/react-native/example/android/build.gradle
dangeross: The example app was throwing errors specific to API 34. Apparently the version of React Native we use in the example doesn't support this. We should update it in another PR

erdemyerebasmaz (approved):
LGTM

packages/react-native/android/build.gradle
```diff
@@ -74,5 +76,6 @@ dependencies {
     // Therefore we ignore the version of JNA that comes with the Breez Liquid SDK from Jitpack
     // and manually add one that does include the necessary Android platform binaries.
     implementation("com.github.breez:breez-sdk-liquid:${getVersionFromNpmPackage()}") { exclude group:"net.java.dev.jna" } // remove for using locally built bindings during development
+    //implementation "org.jetbrains.kotlinx:atomicfu:0.23.1"
```
danielgranhao, line 79: Is this commented line intentional?

danielgranhao (commented):
LGTM. Just a question about a commented line

packages/react-native/android/build.gradle
dangeross: Done [35dadc5](https://github.com/breez/breez-sdk-liquid/pull/927/commits/35dadc57ad78b393c8a3f4e70eeefdc5a86bbf29)

danielgranhao (approved):

PR #928, opened 2025-05-21
Expose bitcoin destination address

### LINKED ISSUE #913: Allow matching created chain swap to resulting payment
Currently, there is no way to match a created chain swap with the resulting payment. 

The `ReceivePaymentResponse` includes just the BIP21 uri. Later, the created payment doesn't have any reference to the BIP21 or its embedded bitcoin address. 

Being able to match these is necessary if an integrator wants to match events to a specific created receive.

Alternatives for addressing this:

- Adding an optional swap id field to `ReceivePaymentResponse`
- Adding the BIP21 uri to `PaymentDetails::Bitcoin`

The latter seems preferable to me as it matches `PaymentDetails::Lightning` having the invoice.

danielgranhao:
Resolves #913

This exposes `bitcoin_address` in `PaymentDetails::Bitcoin`. The goal is to allow integrators to match created receive chain swaps with the resulting payment.

I opted for exposing just the address (not the bip21) as it is simpler and keeps the field consistent across payment directions.

### DIFF
--- a/lib/bindings/src/breez_sdk_liquid.udl
+++ b/lib/bindings/src/breez_sdk_liquid.udl
@@ -633,7 +633,7 @@ dictionary AssetInfo {
 interface PaymentDetails {
     Lightning(string swap_id, string description, u32 liquid_expiration_blockheight, string? preimage, string? invoice, string? bolt12_offer, string? payment_hash, string? destination_pubkey, LnUrlInfo? lnurl_info, string? bip353_address, string? claim_tx_id, string? refund_tx_id, u64? refund_tx_amount_sat);
     Liquid(string asset_id, string destination, string description, AssetInfo? asset_info, LnUrlInfo? lnurl_info, string? bip353_address);
-    Bitcoin(string swap_id, string destination_address, string description, boolean auto_accepted_fees, u32? bitcoin_expiration_blockheight, u32? liquid_expiration_blockheight, string? lockup_tx_id, string? claim_tx_id, string? refund_tx_id, u64? refund_tx_amount_sat);
+    Bitcoin(string swap_id, string bitcoin_address, string description, boolean auto_accepted_fees, u32? bitcoin_expiration_blockheight, u32? liquid_expiration_blockheight, string? lockup_tx_id, string? claim_tx_id, string? refund_tx_id, u64? refund_tx_amount_sat);
 };
 
 dictionary Payment {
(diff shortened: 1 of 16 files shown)


lib/core/src/persist/mod.rs
```diff
@@ -881,6 +883,12 @@ impl Persister {
 
                 PaymentDetails::Bitcoin {
                     swap_id,
+                    destination_address: match maybe_chain_swap_direction {
+                        Some(Direction::Incoming) => maybe_chain_swap_lockup_address,
+                        Some(Direction::Outgoing) => maybe_chain_swap_claim_address,
+                        None => None,
+                    }
+                    .unwrap_or("Destination unknown".to_string()),
```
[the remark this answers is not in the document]
danielgranhao: Applied

dangeross (commented):
I think it might make sense to set the Payment `destination` with this fix. It would then make Bitcoin payments consistent with Lightning payments (destination is invoice/offer, not Liquid address)

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -633,7 +633,7 @@ dictionary AssetInfo {
 interface PaymentDetails {
     Lightning(string swap_id, string description, u32 liquid_expiration_blockheight, string? preimage, string? invoice, string? bolt12_offer, string? payment_hash, string? destination_pubkey, LnUrlInfo? lnurl_info, string? bip353_address, string? claim_tx_id, string? refund_tx_id, u64? refund_tx_amount_sat);
     Liquid(string asset_id, string destination, string description, AssetInfo? asset_info, LnUrlInfo? lnurl_info, string? bip353_address);
-    Bitcoin(string swap_id, string description, boolean auto_accepted_fees, u32? bitcoin_expiration_blockheight, u32? liquid_expiration_blockheight, string? lockup_tx_id, string? claim_tx_id, string? refund_tx_id, u64? refund_tx_amount_sat);
+    Bitcoin(string swap_id, string destination_address, string description, boolean auto_accepted_fees, u32? bitcoin_expiration_blockheight, u32? liquid_expiration_blockheight, string? lockup_tx_id, string? claim_tx_id, string? refund_tx_id, u64? refund_tx_amount_sat);
```
roeierez, line 636: detination_address seems fine In case of outgoing but in case of incoming it might be misleading a bit.
What do you think about renaming to just "bitcoin_address"

roeierez (changes requested):
Looks good.

erdemyerebasmaz:
> I think it might make sense to set the Payment `destination` with this fix. It would then make Bitcoin payments consistent with Lightning payments (destination is invoice/offer, not Liquid address)

I agree with Ross here.

lib/bindings/src/breez_sdk_liquid.udl
danielgranhao: Renamed to `bitcoin_address` in [cee590e](https://github.com/breez/breez-sdk-liquid/pull/928/commits/cee590e110acc45c78e0af04be917d1a4a05c848)

danielgranhao:
@dangeross I've adjusted the payment's `destination` as you proposed.

roeierez (approved):

dangeross (approved):
LGTM

PR #935, opened 2025-05-26
Fix status stream task shutdown

danielgranhao:
This fixes an issue where if the status stream cannot be connected to, calling the SDK's `disconnect()` won't stop the stream task.


roeierez (approved):
LGTM

hydra-yse (approved):

PR #936, opened 2025-05-27
Add nodejs min version note to readme

danielgranhao:
Our implementation relies on the websocket client implementation introduced in node v22: https://nodejs.org/en/blog/announcements/v22-release-announce#websocket


roeierez (approved):
LGTM

PR #938, opened 2025-05-27
CLI: Add phrase file arg

danielgranhao:
This adds a convenient argument to the CLI that allows specifying which phrase file name is used. Not providing one keeps the same default.

### DIFF
--- a/cli/src/persist.rs
+++ b/cli/src/persist.rs
@@ -15,8 +15,8 @@ pub(crate) struct CliPersistence {
 }
 
 impl CliPersistence {
-    pub(crate) fn get_or_create_mnemonic(&self) -> Result<Mnemonic> {
-        let filename = Path::new(&self.data_dir).join(PHRASE_FILE_NAME);
+    pub(crate) fn get_or_create_mnemonic(&self, phrase_file_name: Option<&str>) -> Result<Mnemonic> {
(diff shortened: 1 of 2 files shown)


cli/src/persist.rs
```diff
@@ -15,8 +15,8 @@ pub(crate) struct CliPersistence {
 }
 
 impl CliPersistence {
-    pub(crate) fn get_or_create_mnemonic(&self) -> Result<Mnemonic> {
-        let filename = Path::new(&self.data_dir).join(PHRASE_FILE_NAME);
+    pub(crate) fn get_or_create_mnemonic(&self, phrase_file_name: Option<&str>) -> Result<Mnemonic> {
+        let filename = Path::new(&self.data_dir).join(phrase_file_name.unwrap_or(PHRASE_FILE_NAME));
```
roeierez, line 19: If the user specifies a file shouldn't we expect a full path? It seems that the user provided file should be a path relative the the data dir.

danielgranhao: Applied in [8ea49ab](https://github.com/breez/breez-sdk-liquid/pull/938/commits/8ea49ab017d4c9761cbfb44b5380cce0943014d0)

roeierez: I actually meant if the user provide a file path then we should take it as is and not assume it is under the data_dir.
Does it make sense to you?

PR #946, opened 2025-06-06
Use address as destination on incoming liquid payments

### LINKED ISSUE #942: Destination in incoming liquid payments is script pubkey but should be address

danielgranhao:
Resolves #942

### DIFF
--- a/lib/core/src/persist/mod.rs
+++ b/lib/core/src/persist/mod.rs
@@ -228,11 +228,17 @@ impl Persister {
                 return Ok(());
             }
         };
-        let maybe_script_pubkey = tx
+        let maybe_address = tx
             .outputs
             .iter()
             .find(|output| output.is_some())
-            .and_then(|output| output.clone().map(|o| o.script_pubkey.to_hex()));
+            .and_then(|output| {
+                output.clone().and_then(|o| {
+                    o.address.blinding_pubkey.map(|blinding_pubkey| {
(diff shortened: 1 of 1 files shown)


lib/core/src/persist/mod.rs
```diff
@@ -228,11 +228,21 @@ impl Persister {
...
-            .and_then(|output| output.clone().map(|o| o.script_pubkey.to_hex()));
+            .and_then(|output| {
+                output.clone().and_then(|o| {
+                    o.address
+                        .blinding_pubkey
+                        .ok_or_else(|| anyhow!("Missing blinding pubkey for address"))
+                        .map(|blinding_pubkey| {
+                            o.address.to_confidential(blinding_pubkey).to_string()
+                        })
+                        .ok()
```
dangeross, line 243: Could this be simplified?
```suggestion
                    o.address.blinding_pubkey.map(|blinding_pubkey| {
                        o.address.to_confidential(blinding_pubkey).to_string()
                    })
```

dangeross (commented):
Nice, one nit

lib/core/src/persist/mod.rs
danielgranhao: Fixed! Thanks

dangeross (approved):
LGTM

roeierez (approved):

PR #951, opened 2025-06-13
SDK: Decouple balance from `PaymentTxData`

hydra-yse:
This PR creates a separate table for payment balances so we can then display multiple payments depending on the output asset type.

### DIFF
--- a/packages/react-native/ios/RNBreezSDKLiquid.swift
+++ b/packages/react-native/ios/RNBreezSDKLiquid.swift
@@ -5,9 +5,9 @@ import Foundation
 class RNBreezSDKLiquid: RCTEventEmitter {
     static let TAG: String = "BreezSDKLiquid"
 
-    public static var emitter: RCTEventEmitter!
-    public static var hasListeners: Bool = false
-    public static var supportedEvents: [String] = ["breezSdkLiquidLog"]
+    static var emitter: RCTEventEmitter!
+    static var hasListeners: Bool = false
+    static var supportedEvents: [String] = ["breezSdkLiquidLog"]
 
     private var bindingLiquidSdk: BindingLiquidSdk!
 


lib/core/src/persist/migrations.rs
```diff
@@ -344,5 +344,26 @@ pub(crate) fn current_migrations(network: LiquidNetwork) -> Vec<&'static str> {
             data BLOB NOT NULL
         ) STRICT;
         ",
+        "
+        CREATE TABLE IF NOT EXISTS payment_balance (
+            tx_id TEXT NOT NULL REFERENCES payment_tx_data(tx_id),
+            asset_id TEXT NOT NULL REFERENCES asset_metadata(asset_id),
```
dangeross, line 350: Can you check `replace_asset_metadata()`, it removes non-default asset metadata and adds the new ones from the config. Can you make sure it still works? 

I suspect you need to handle referential errors now if there is a related row here. Same with inserting rows that already exist, just update the non asset_id columns

dangeross (commented):
Can you test sending USDT, paying the fee in L-BTC? I suspect we get two payment rows, one for the L-BTC fee and one for USDT. If that's the case, is that the intended behavior?

hydra-yse:
> Can you test sending USDT, paying the fee in L-BTC? I suspect we get two payment rows, one for the L-BTC fee and one for USDT. If that's the case, is that the intended behavior?

You're right, that's what happened. I'll try adding a condition to check whether or not it's a fee output and to exclude that from the `PaymentBalance` list. Thanks :+1:

lib/core/src/persist/migrations.rs
hydra-yse: cc @roeierez What should the expected behavior be here? I think we can set `ON DELETE CASCADE` as it's a setting that affects at the app level (so the user should make sure to wipe out the database cache when their asset metadata fields change). 

Alternatively we can just remove the asset_id reference here, and just manually exclude the payment from the list if its asset id is not present in the asset_metadata table.

hydra-yse: I see. I thought the idea behind the asset metadata table was to have more control over which assets can be used/displayed. But if it's fine showing it then I'm not against removing the reference from the table :+1:

lib/core/src/persist/mod.rs
```diff
@@ -306,35 +328,30 @@ impl Persister {
...
-                ptx.amount,
                 ptx.fees_sat,
-                ptx.payment_type,
                 ptx.is_confirmed,
                 ptx.unblinding_data,
             ),
         )?;
 
+        for balance in balances {
+            Self::insert_or_update_payment_balance(&tx, &ptx.tx_id, balance)?;
```
roeierez, line 352: should we delete the balances before updating in case the asset id no longer exists?

roeierez (changes requested):

lib/core/src/persist/migrations.rs
```diff
@@ -344,5 +344,26 @@ pub(crate) fn current_migrations(network: LiquidNetwork) -> Vec<&'static str> {
             data BLOB NOT NULL
         ) STRICT;
         ",
+        "
+        CREATE TABLE IF NOT EXISTS payment_balance (
+            tx_id TEXT NOT NULL REFERENCES payment_tx_data(tx_id),
```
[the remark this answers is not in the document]
roeierez: Sounds good to me

danielgranhao (commented):
Didn't test yet, but looks good. Just a question

lib/core/src/persist/mod.rs
```diff
@@ -306,35 +328,36 @@ impl Persister {
...
                 ptx.is_confirmed,
                 ptx.unblinding_data,
             ),
         )?;
 
+        if !balances.is_empty() {
+            tx.execute(
+                "DELETE FROM payment_balance WHERE tx_id = ?",
+                params![&ptx.tx_id],
+            )?;
```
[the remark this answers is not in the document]
hydra-yse: From Roei's comment [here](https://github.com/breez/breez-sdk-liquid/pull/951#discussion_r2150247786), if we update a mempool tx (via RBF for example) we need to ensure the previous balance is deleted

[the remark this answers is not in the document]
danielgranhao: Not sure if the RBF example applies. If the tx is the same, won't the balances be an exact match as well?

lib/core/src/persist/mod.rs
```diff
@@ -200,34 +199,43 @@ impl Persister {
...
+        tx_balances
+            .entry(lbtc_asset_id)
+            .and_modify(|b| *b += tx.fee as i64);
+
+        let num_assets = tx_balances.len();
+        let payment_balances: Vec<PaymentTxBalance> = tx_balances
+            .into_iter()
+            .filter_map(|(asset_id, balance)| {
+                // Ignore the fee balance only if there is more than one asset
+                if balance == 0 && num_assets >= 2 {
```
roeierez, line 213: Can you elaborate here? Why there is an exception for where the balance is zero?

hydra-yse: It's to account for the fees being present in the tx balance (so for example, payments in USDt still include a negative L-BTC fee balance, which we don't want to display as a separate payment of course, so we skip [see point 2 below]).

Looking at the code and thinking about the logic again, there's one last thing to change.   To recap:

### Regular Payments (single asset)
Note: The fees are always in L-BTC. Outbound => Paying others, Inbound => Paying ourselves

1. Sending L-BTC (outbound) => 1 balance `-(lbtc + fees)`. We want 1 payment (regular LBTC transfer with fees).
2. Sending L-BTC (inbound) => 1 balance (`0`). When we pay ourselves, we still want to show the payment, despite having a net balance of 0. Therefore I added the `num_assets >= 2` check.
3. Generic asset (outbound) => 2 balances `-asset_value, -fees`. We want 1 payment (regular asset transfer with fees)
4. Generic asset (inbound) => 2 balances `asset_value (=0), -fees`. We want 1 payment (0-amount asset transfer, with fees)

### SideSwap Payments (mixed assets)
1. LBTC (outbound) to other asset (inbound) => 2 balances `+asset_value, -(lbtc + fees)`. We want 2 payments (one asset receive, one lbtc send)

The check for fees encapsulates this directly rather than making distinctions for each case. But we can also have more explicit logic to show what's going on.

hydra-yse: Check out 8668cf5e8c1b09363a9b74c0783c6ffbf80c87f5 for my solution, hopefully it made things clearer. It should respect all of the scenarios listed above

lib/core/src/persist/mod.rs
```diff
@@ -199,35 +198,57 @@ impl Persister {
...
+                // We always compensate for fees for the outbound L-BTC balance
+                if asset_id == lbtc_asset_id
+                    && balance < 0
+                    && (num_assets == 1 || has_non_zero_foreign_asset_balances)
+                {
+                    balance += tx.fee as i64;
+
+                    // If the resulting balance is null, and we have more than one non-zero foreign asset,
+                    // we need to avoid inserting this payment balance
+                    if balance == 0 && has_non_zero_foreign_asset_balances {
```
roeierez, line 225: Why do we need the has_non_zero_foreign_asset_balances ?
In case we don't have balance for asset that is not lbtc would it make a different?

lib/core/src/persist/mod.rs
```diff
@@ -199,35 +198,57 @@ impl Persister {
...
+        });
+
+        let payment_balances: Vec<PaymentTxBalance> = tx_balances
+            .into_iter()
+            .filter_map(|(asset_id, mut balance)| {
+                // https://github.com/breez/breez-sdk-liquid/pull/951#discussion_r2203640193
+                // We always compensate for fees for the outbound L-BTC balance
+                if asset_id == lbtc_asset_id
+                    && balance < 0
+                    && (num_assets == 1 || has_non_zero_foreign_asset_balances)
```
roeierez, line 219: Why do we need to check the balance, num_assets and has_non_zero_foreign_asset_balances ?
I might be missing something here but I think naively if the lbtc balance is the same as the fees (that are always in lbtc) then we should ignore the payment. Am I wrong here?

roeierez (changes requested):

lib/core/src/persist/mod.rs
hydra-yse: That's close, but doesn't cover for the cases where we're sending funds to ourselves, both in the L-BTC case and in the asset case. The net balance is zero, but we still want to display the payment, thus the extra checks

lib/core/src/persist/mod.rs
```diff
@@ -199,35 +198,46 @@ impl Persister {
...
-        let (asset_id, payment_type, amount) = match balance {
-            Some((asset_id, asset_amount)) => {
-                let payment_type = match asset_amount >= 0 {
+        let num_outputs = tx.outputs.iter().filter(|out| out.is_some()).count();
+
+        let payment_balances: Vec<PaymentTxBalance> = tx_balances
+            .into_iter()
+            .filter_map(|(asset_id, mut balance)| {
+                // Only account for fee changes in case of outbound L-BTC payments
+                if asset_id == lbtc_asset_id && balance < 0 {
```
roeierez, line 210: Why do we need the balance < 0 condition? Didn't we say we always adding the fees to the balance in case of lbtc?

hydra-yse: No, only in case of outbound send. That's how it also was previously: https://github.com/breez/breez-sdk-liquid/blob/6cd2561f233924624e218d08764df6668b781e38/lib/core/src/persist/mod.rs#L221-L223

lib/core/src/persist/mod.rs
dangeross: I couldn't find where the delete was happening

dangeross (commented):
Looking good, just one question

dangeross (approved):

danielgranhao (approved):
LGTM - (edit: still need to check why the E2E tests are failing)

PR #955, opened 2025-06-20
Store the BOLT12 payer note / LNURL-pay comment

### LINKED ISSUE #949: Bolt12: Persist Payer Note on payment data
While we do send the payer note, and can receive and observe it in the Bolt 12 invoice data, it’s currently not being persisted for either incoming or outgoing payments.

Related Misty Breez issue:
- [Bolt12: Payer Note is missing on payment data](https://github.com/breez/misty-breez/issues/567)

dangeross:
This PR primarily updates the SDK to store the payer note for a BOLT12 invoice. In doing so I've also made the SDK [LUD-12](https://github.com/lnurl/luds/blob/luds/12.md) compatible by allowing the SDK to store a comment on the swap table before the payment is made. Now that a LNURL comment / BOLT12 invoice payer note can be stored ahead of the actual payment, this information is persisted even if the payment is not actually paid via the swap, but via MRH transaction.

Also changes the input tokenization in the CLI to correctly tokenize quoted strings.

Fixes #949

**Testing notes:**
- Sending with BOLT12 + payer note to an **SDK** receiver we should see a payer note in the sender payment
- Sending with BOLT12 + payer note to a **non-SDK** receiver we should see a payer note in the sender payment
- Receiving with BOLT12 + payer note from an **SDK** sender we should see a payer note in the receiver payment
- Receiving with BOLT12 + payer note from a **non-SDK** sender we should see a payer note in the receiver payment
- ~Receiving with LNURL Lightning address + comment from an **SDK** sender we should see a payer note in the receiver payment~ (Skip - not testable on Misty)
- Receiving with LNURL Lightning address + comment from a **non-SDK** sender we should see a payer note in the receiver payment


lib/core/src/sdk.rs
```diff
@@ -4299,14 +4303,15 @@ impl LiquidSdk {
                     description,
                     lnurl_info: Some(LnUrlInfo {
                         ln_address: prepare_response.data.ln_address,
-                        lnurl_pay_comment: prepare_response.comment,
+                        lnurl_pay_comment: prepare_response.comment.clone(),
```
danielgranhao, line 4306: Nit: is this added clone needed?

danielgranhao (approved):
LGTM. Left a nit question

roeierez:
Intuitively I think there are two features leaning on the same field.
`payer_note` - which is a comment added by the payer (lnurl, bolt12)
`tag` - which is some metadata text added by users for their own payments.

I hope I got it correct and if this is so then I suggest to start with payer_note only.

danielgranhao (approved):

roeierez:
@dangeross I think this can be merged now?

PR #957, opened 2025-06-23
fix: Wrong amount in payment tx data persisted on claim

danielgranhao:
This fixes an issue where amountless swaps would temporarily show an amount of zero and fees equal to the amount received.

The payment tx data we insert on claim was using `receiver_amount_sat`, which for amountless swaps is 0. Now, we use `accepted_receiver_amount_sat` in case it's present.


roeierez (approved):
LGTM

dangeross (approved):

PR #960, opened 2025-06-30
Use sats leeway for amountless swaps and set default

danielgranhao:
This PR:

- Change the configurable leeway for amountless swaps to be denominated in sats instead of sats/vbyte
- Sets the default to 500 sats


roeierez (approved):
LGTM

PR #968, opened 2025-07-10
Fix `list_refundables` swap to utxos matching

### LINKED ISSUE #966: Refundable incoming chain swap with 0 sats
If a refundable incoming chain swap has its user lockup dropped from the mempool, `list_refundables()` shows it as refundable with an amount of 0 sats.

A refundable swap can have its user lockup pending if the amount is outside of the allowed range (thus becoming refundable immediately, even before the tx confirms).

danielgranhao:
This fixes 2 issues:
- If the chain service provided utxos in a different order from the requested scripts, which we seen happen, this fixes the matching between swaps and their amount
- Also resolves #966 as swaps for which we don't find utxos no longer get included in the refundables


erdemyerebasmaz (approved):
LGTM

PR #972, opened 2025-07-15
Optimize bitcoin tip fetch + gracefully handle errors

danielgranhao:
This PR:
- Optimizes fetching of the Bitcoin tip by only doing it if the SDK knows of any chain swap
- Gracefully handles Bitcoin tip fetching errors by allowing the sync of other swaps to function normally. If we needed the tip but couldn't fetch it, only chain swaps are not synced. Warnings are logged accordingly.

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -599,37 +599,50 @@ impl LiquidSdk {
                 false
             }
         };
-        // Get the Bitcoin tip and process a new block (only if chain swaps are used)
-        let (maybe_bitcoin_tip, is_new_bitcoin_block) = if self
-            .persister
-            .has_chain_swaps()
+
+        // Get the recoverable swaps list using the last known tips
+        let last_monitored_swaps = self
+            .get_monitored_swaps_list(
+                false,
+                true,
+                ChainTips {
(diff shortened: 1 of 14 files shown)


dangeross (approved):
LGTM

lib/core/src/sdk.rs
```diff
@@ -4001,13 +4058,43 @@ impl LiquidSdk {
...
+                    .any(|s| matches!(s, Swap::Chain(_)))
+                    .not()
+                {
+                    info!("No chain swaps being monitored, skipping bitcoin tip fetch");
+                    None
+                } else {
+                    self.bitcoin_chain_service.tip().await.ok()
+                }
+            },
+        };
```
roeierez, line 4086: Eventually the sync_inner needs two things: recoverable_swaps and chain_tips.
The above code block gets this info and we also getting this info in the `track_new_blocks` function.
Do you think we can create one function to fetch this info that can be reused in these two places?
Something like `get_sync_context` ? It will help get out the small details out of the main flow.

roeierez (commented):
Looks good! One comment about reusing if possible. Other than that good to go.

roeierez (approved):
LGTM

PR #975, opened 2025-07-23
wasm: Clear in-memory db on disconnect

danielgranhao:
This PR fixes a memory leak that affected Wasm packages. The memory allocated by the in-memory database was never cleared, so in a server use case, where many different wallets are loaded/unloaded in a single run, wallets not in use would continue to occupy space.

The `rusqlite` fork was updated so we could access the `MemVfsUtil` struct, which allows deleting a db.

In local tests, I've verified that after these changes, starting and stopping different SDK instances no longer leads to a forever-increasing memory footprint and now plateaus.

### DIFF
--- a/cli/Cargo.lock
+++ b/cli/Cargo.lock
@@ -3,18 +3,6 @@
 version = 4
 
 [[package]]
-name = "accessory"
-version = "2.1.0"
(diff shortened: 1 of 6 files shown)


hydra-yse (approved):
Great work! Thanks for switching from my branch to Breez org 😅

PR #983, opened 2025-08-01
Add config option to avoid using MRH to send payments

dangeross:
If the global config option is disabled, then only swaps will be used to send payments. This also means the sender will always receive the lightning preimage on successful payment

### DIFF
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -99,7 +99,7 @@ pub struct Config {
     pub asset_metadata: Option<Vec<AssetMetadata>>,
     /// The SideSwap API key used for making requests to the SideSwap payjoin service
     pub sideswap_api_key: Option<String>,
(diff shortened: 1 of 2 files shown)


lib/core/src/model.rs
```diff
@@ -99,6 +99,8 @@ pub struct Config {
     pub asset_metadata: Option<Vec<AssetMetadata>>,
     /// The SideSwap API key used for making requests to the SideSwap payjoin service
     pub sideswap_api_key: Option<String>,
+    /// Set this to false to disable the use of Magic Routing Hints (MRH) to send payments.
```
danielgranhao, line 102:
```suggestion
    /// Set this to false to disable the use of Magic Routing Hints (MRH) to send payments. Enabled by default.
```

danielgranhao (approved):
LGTM. Left just a suggestion about the doc comment.

PR #985, opened 2025-08-05
Fix tasks leaking

danielgranhao:
This PR fixes memory leaks caused by tasks not shutting down soon after when we send the shutdown signal. There were several places within background tasks where futures were being awaited without allowing early shutdown.

A node.js Wasm testbench is also added that allows checking memory usage as batches of SDK instances are connected and disconnected. Before this fix, starting/stopping 50k instances in batches of 100 led to external memory increasing from ~50MB to ~350MB. With the fix, the same test leads to an increase of ~0.6MB.

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -43,7 +43,7 @@ use sdk_common::utils::Arc;
 use side_swap::api::SideSwapService;
 use signer::SdkSigner;
 use swapper::boltz::proxy::BoltzProxyFetcher;
-use tokio::sync::{watch, RwLock};
+use tokio::sync::{watch, Mutex, RwLock};
 use tokio_stream::wrappers::BroadcastStream;
 use tokio_with_wasm::alias as tokio;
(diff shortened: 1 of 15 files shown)


lib/core/src/sdk.rs
```diff
@@ -536,51 +554,85 @@ impl LiquidSdk {
...
+
+        let shutdown_receiver = self.shutdown_receiver.clone();
+        info!("Starting track-realtime-sync-events task");
+        Some(tokio::spawn(async move {
+            run_with_shutdown(
+                shutdown_receiver,
+                "Received shutdown signal, exiting real-time sync loop",
+                track_realtime_sync_events_future,
+            )
+            .await
```
JssDWt, line 634: Does this not hang in the `recv` until the task is eventually aborted after the timeout?

JssDWt (approved):
Approved, given the tasks actually stop the moment shutdown is requested, and not after the timeout.

PR #986, opened 2025-08-06
Side-Swap: Add any-to-any payments

hydra-yse:


lib/core/src/sdk.rs
```diff
@@ -1313,73 +1314,69 @@ impl LiquidSdk {
...
-                                err: format!("Asset {asset_id} is not supported"),
-                            },
-                        )?;
-                        let receiver_amount_sat = asset_metadata.amount_to_sat(receiver_amount);
+                        let from_asset = from_asset.unwrap_or(to_asset.clone());
+                        let _payer_asset_metadata = self
+                            .persister
+                            .get_asset_metadata(&from_asset)?
+                            .ok_or(PaymentError::AssetError {
+                                err: format!("Asset {to_asset} is not supported"),
```
roeierez, line 1327: from_asset ?

lib/core/src/sdk.rs
```diff
@@ -1313,73 +1314,69 @@ impl LiquidSdk {
...
                             }
                             true => {
-                                let asset_id = AssetId::from_str(&asset_id)?;
-                                let sideswap_service = SideSwapService::from_sdk(self);
-
-                                let swap = sideswap_service
-                                    .get_asset_swap(asset_id, receiver_amount_sat)
+                                let to_asset = AssetId::from_str(&to_asset)?;
+                                let from_asset = AssetId::from_str(&from_asset)?;
+                                let swap = SideSwapService::from_sdk(self)
```
roeierez, line 1360: We skipped balance check?

roeierez (changes requested):

lib/core/src/sdk.rs
```diff
@@ -1313,73 +1314,74 @@ impl LiquidSdk {
...
-                            },
-                        )?;
-                        let receiver_amount_sat = asset_metadata.amount_to_sat(receiver_amount);
+                        let from_asset = from_asset.unwrap_or(to_asset.clone());
+                        let _payer_asset_metadata = self
+                            .persister
+                            .get_asset_metadata(&from_asset)?
+                            .ok_or(PaymentError::AssetError {
+                                err: format!("Asset {from_asset} is not supported"),
+                            })?;
```
danielgranhao, line 1328: Nit: I think the intent would be clearer if we used something like 

```
ensure_sdk!(
    self.persister.get_asset_metadata(&from_asset)?.is_some(), 
    PaymentError::AssetError {err: format!("Asset {from_asset} is not supported")}
);
```

lib/core/src/sdk.rs
```diff
@@ -1313,73 +1314,74 @@ impl LiquidSdk {
...
+                            .get_asset_metadata(&to_asset)?
+                            .ok_or(PaymentError::AssetError {
+                                err: format!("Asset {to_asset} is not supported"),
+                            })?;
+                        let receiver_amount_sat =
+                            receiver_asset_metadata.amount_to_sat(receiver_amount);
+
+                        let asset_fees = if estimate_asset_fees.unwrap_or(false) {
+                            self.payjoin_service
+                                .estimate_payjoin_tx_fee(&from_asset, receiver_amount_sat)
```
danielgranhao, line 1340: Shouldn't it be `to_asset` here?

```suggestion
                                .estimate_payjoin_tx_fee(&to_asset, receiver_amount_sat)
```

lib/core/src/sdk.rs
```diff
@@ -1629,53 +1630,71 @@ impl LiquidSdk {
...
-
-                    let is_liquid_payment = *asset_id == self.config.lbtc_asset_id();
-                    let pay_with_bitcoin = match req.prepare_response.amount {
-                        Some(PayAmount::Asset {
-                            pay_with_bitcoin, ..
-                        }) => pay_with_bitcoin.unwrap_or(false),
-                        _ => false,
-                    };
+                let asset_pay_fees = req.use_asset_fees.unwrap_or_default();
+                let mut response = match amount.as_ref().is_some_and(|a| a.is_sideswap_payment()) {
```
danielgranhao, line 1634: I think the entire `match` block that starts here could become a lot more readable if we extracted some logic into separate methods

danielgranhao (changes requested):
Concept LGTM. Left some comments

lib/core/src/sdk.rs
```diff
@@ -1629,53 +1630,71 @@ impl LiquidSdk {
...
-                    let fees_sat = fees_sat.ok_or(PaymentError::InsufficientFunds)?;
-
-                    let is_liquid_payment = *asset_id == self.config.lbtc_asset_id();
-                    let pay_with_bitcoin = match req.prepare_response.amount {
-                        Some(PayAmount::Asset {
-                            pay_with_bitcoin, ..
-                        }) => pay_with_bitcoin.unwrap_or(false),
-                        _ => false,
-                    };
+                let asset_pay_fees = req.use_asset_fees.unwrap_or_default();
```
roeierez, line 1633: Are we missing a validation in case user specifies both asset_pay_fees and sideswap payment?

roeierez (changes requested):
Looks good. Some small comments.

lib/core/src/sdk.rs
hydra-yse: You're right, we'll have to account for that as they're mutually exclusive

lib/core/src/model.rs
```diff
@@ -795,18 +805,32 @@ pub enum PayAmount {
...
+        // The asset id specifying which asset will be sent
+        to_asset: String,
         receiver_amount: f64,
         estimate_asset_fees: Option<bool>,
-        /// Specifies whether or not to always use the wallet's L-BTC to execute the payment.
-        /// If true, it will try swapping the asset via the [Side Swap Service](crate::side_swap::api::SideSwapService)
-        pay_with_bitcoin: Option<bool>,
+        // The asset id whose balance we want to send funds with.
+        // Defaults to the value provided for [PayAmount::Asset::to_asset]
+        from_asset: Option<String>,
```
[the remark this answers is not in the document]
hydra-yse: @roeierez wdyt?

lib/core/src/sdk.rs
hydra-yse: You're right. I had set it to `from` because the first argument actually refers to the fee asset, so it would make sense. But later on we build the tx with the `to_asset`, so this needs correction.

hydra-yse: Same as [comment](https://github.com/breez/breez-sdk-liquid/pull/986#discussion_r2262990890), they are now mutually exclusive.

lib/core/src/model.rs
```diff
@@ -795,18 +811,32 @@ pub enum PayAmount {
 
     /// The amount of an asset that will be received
     Asset {
-        asset_id: String,
+        // The asset id specifying which asset will be sent
```
danielgranhao, line 814: I just noticed this now: should this and the comments below be doc comments (`///`)?

hydra-yse: Nice catch

danielgranhao (approved):
LGTM - just noting the caveat that asset self-transfers are not showing the asset flowing out, just coming in. @hydra-yse and I agreed to open an issue and fix it in a separate PR.

hydra-yse:
> LGTM - just noting the caveat that asset self-transfers are not showing the asset flowing out, just coming in. @hydra-yse and I agreed to open an issue and fix it in a separate PR.

See #991

PR #994, opened 2025-08-10
Only recover expired chain swaps on new Bitcoin block

### LINKED ISSUE #977: Query bitcoin chain data only on new bitcoin blocks for expired swaps
Currently, we run recovery (`recover_from_onchain`) on chain swaps on detection of either a new Bitcoin or Liquid block. This means that if we are running due to a new liquid block, we'll still query the Bitcoin esplora instance for related script histories, balances, etc. This can likely be avoided given that there was no new Bitcoin block.

(not sure about mempool txs though)

danielgranhao:
Resolves #977

### DIFF
--- a/lib/core/src/sdk.rs
+++ b/lib/core/src/sdk.rs
@@ -3804,12 +3804,11 @@ impl LiquidSdk {
 
     /// Returns a list of swaps that need to be monitored for recovery.
     ///
-    /// If `partial_sync` is true, only receive swaps will be considered.
-    ///
     /// If no Bitcoin tip is provided, chain swaps will not be considered.
     pub(crate) async fn get_monitored_swaps_list(
         &self,
-        partial_sync: bool,
+        only_receive_swaps: bool,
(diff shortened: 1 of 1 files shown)


lib/core/src/sdk.rs
```diff
@@ -4350,19 +4355,19 @@ impl LiquidSdk {
...
+                let only_receive_swaps = !is_new_liquid_block && !is_new_bitcoin_block;
+                let include_expired_incoming_chain_swaps = is_new_bitcoin_block;
+
+                recoverable_swaps = self
+                    .get_monitored_swaps_list(
+                        only_receive_swaps,
+                        include_expired_incoming_chain_swaps,
+                        ChainTips {
+                            liquid_tip,
+                            bitcoin_tip: None,
```
dangeross, line 4367: Shouldn't we pass `bitcoin_tip` here? else we'll never get the chain swaps

roeierez (approved):
LGTM

PR #1000, opened 2025-08-13
Add esplora client connect timeout

dangeross:
This PR:
- Adds a connect timeout to the Bitcoin esplora client
- Adds logging around sync


roeierez (approved):
LGTM

dangeross:
> @dangeross can you confirm we also have proper timeout on requests?

It sets correctly the reqwest client request timeout. We should also be aware that reqwest for WASM doesn't support these timeouts (connect & request).

hydra-yse (approved):

PR #1003, opened 2025-08-14
SDK: Add onchain scan logging and reduce requests max_retries 

hydra-yse:


lib/core/src/chain/bitcoin/esplora.rs
```diff
@@ -51,7 +51,7 @@ impl EsploraBitcoinChainService {
         let client = Builder::new(esplora_url)
             .connect_timeout(10)
             .timeout(3)
-            .max_retries(10)
+            .max_retries(3)
```
roeierez, line 54: I would prefer 2 or 1. I don't see why we need 3 now after the change of policy. Also request timeout takes 3 seconds and 3 retries may reach 9 seconds worst case which is a lot.

roeierez (commented):
One small comment, otherwise LGTM

roeierez (approved):
LGTM

PR #1004, opened 2025-08-14
Notification Plugin: Disconnect SDK during shutdown

dangeross:
Force SDK disconnect during shutdown in case the foreground service isn't stopped


hydra-yse (approved):

erdemyerebasmaz (approved):

erdemyerebasmaz:
These do not effect upcoming Cargokit changes and we can move forward with it.

PR #1009, opened 2025-09-02
fix: update Vite config to include WASM assets and optimize dependencies

samuelmanzanera:
This PR aims to fix the Vite's example WebAssembly loading error `CompileError: WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 3c 21 64 6f @+0` 

The problem was how Vite was handling .wasm and did not include the file - causing the HTML page to load in the `init` function.


danielgranhao:
Thanks for the PR @samuelmanzanera!

I'm having trouble reproducing the issue. Could you share more about it? Which browser you tried, how you built the project, and any other information you think might be relevant would be helpful

danielgranhao (approved):

dangeross (approved):

PR #1011, opened 2025-09-05
Add plugins support to SDK

hydra-yse:


lib/core/src/plugin.rs
```diff
@@ -0,0 +1,9 @@
+use maybe_sync::{MaybeSend, MaybeSync};
+
+use crate::sdk::LiquidSdk;
+use sdk_common::utils::Arc;
+
+pub trait Plugin: MaybeSend + MaybeSync {
+    fn on_start(&self, sdk: Arc<LiquidSdk>);
```
roeierez, line 7: From the nwc PR I think I noticed the plugin uses storage. Shouldn't we add here a key value store trait and pass it to on_start?

lib/core/src/sdk.rs
```diff
@@ -120,6 +122,7 @@ impl LiquidSdkBuilder {
         config: Config,
         server_url: String,
         signer: Arc<Box<dyn Signer>>,
+        plugins: Option<Vec<Arc<dyn Plugin>>>,
```
roeierez, line 125: Since this is optional I think it is better to add a function to the builder `plugins` and pass the list there and then we can remove it from the constructor.

roeierez (changes requested):
Looks good! I like the simplicity.

lib/core/src/sdk.rs
hydra-yse: Sure, I've added it to the constructor just because we pass it directly to the connect method

lib/core/src/sdk.rs
```diff
@@ -539,6 +559,11 @@ impl LiquidSdk {
                 handle,
             });
         }
+        for plugin in &self.plugins {
+            plugin
+                .on_start(self.clone(), PluginStorage::new(self.persister.clone()))
```
roeierez, line 564: This passes the same storage to all plugins allowing them to override each other's keys. 
If the PluginStorage is also passed with some identifier (plugin id?) we can add this prefix to every key in the implementation.

lib/bindings/src/breez_sdk_liquid.udl
```diff
@@ -746,12 +746,29 @@ dictionary AssetMetadata {
...
+    void set_item(string key, string value);
+
+    [Throws=SdkError]
+    string? get_item(string key);
+
+    [Throws=SdkError]
+    void remove_item(string key);
+};
+
+callback interface BindingPlugin {
```
roeierez, line 760: NIT: SdkPlugin?

roeierez (approved):
LGTM

lib/core/src/plugin.rs
```diff
@@ -0,0 +1,56 @@
...
+            persister,
+            plugin_id,
+        })
+    }
+
+    pub(crate) fn scoped_key(&self, key: String) -> String {
+        format!("{}-{}", self.plugin_id, key)
+    }
+
+    pub fn set_item(&self, key: String, value: String) -> SdkResult<()> {
```
danielgranhao, line 29: Would it make sense to have a PluginStorage-specific result here instead of `SdkResult`? Same for the other public methods

danielgranhao (commented):
Looks great! Just a couple of nits. Also, need to check the E2E tests. Seems like they need to be adjusted:

```
error[E0061]: this function takes 2 arguments but 1 argument was supplied
   --> core/tests/regtest/mod.rs:100:19
    |
100 |           let sdk = LiquidSdk::connect(ConnectRequest {
    |  ___________________^^^^^^^^^^^^^^^^^^-
101 | |             config,
102 | |             mnemonic: Some(mnemonic.to_string()),
103 | |             passphrase: None,
104 | |             seed: None,
105 | |         })
    | |__________- argument #2 of type `std::option::Option<Vec<Arc<(dyn Plugin + 'static)>>>` is missing
```

lib/core/src/plugin.rs
hydra-yse: Tbh the only error variant that can occur is if there's a problem connecting and writing to the database. Not sure whether it's worth creating a new error type for, but I'm not strongly against it either. Maybe we could also just return a String error "could not write to database".

dangeross (commented):
The React Native generated code doesn't look right to me. It's generating the PluginStorage functions in the main module. I think for now it's best to exclude these functions from being generated. 
- https://github.com/breez/breez-sdk-liquid/blob/da3533cde7b8dac202689416464af9be4620ecb7/lib/bindings/langs/react-native/src/gen_kotlin/mod.rs#L12-L20
- https://github.com/breez/breez-sdk-liquid/blob/da3533cde7b8dac202689416464af9be4620ecb7/lib/bindings/langs/react-native/src/gen_swift/mod.rs#L11-L19

The manually written `connect` function would also need updating.
- https://github.com/breez/breez-sdk-liquid/blob/da3533cde7b8dac202689416464af9be4620ecb7/lib/bindings/langs/react-native/src/gen_kotlin/templates/module.kt#L80-L98
- https://github.com/breez/breez-sdk-liquid/blob/da3533cde7b8dac202689416464af9be4620ecb7/lib/bindings/langs/react-native/src/gen_swift/templates/module.swift#L92-L107

hydra-yse:
Thanks for letting me know Ross, I'll update accordingly

PR #1019, opened 2025-10-03
SDK: Add `SyncFailed` event

### LINKED ISSUE #1018: Sync: Notify sync health status
## Description
Currently, when `sync()` fails, clients have no way to be notified about sync service health status. While apps can detect when the SDK disconnects, they cannot tell when the SDK is connected but the sync service is experiencing instability, forcing them to manually track sync failures and make assumptions.

## Expected behavior
The SDK should expose sync health status, allowing clients to properly notify their users about sync issues without implementing custom health check logic.

hydra-yse:
Closes #1018.
Adds a `SyncFailed` enum variant to `SdkEvent` which is triggered only when the background task `track_new_blocks` fails. The manual `sync` trigger is left as-is as the error is returned per-call.

### DIFF
--- a/lib/core/src/model.rs
+++ b/lib/core/src/model.rs
@@ -502,7 +502,7 @@ pub enum SdkEvent {
     },
     /// Synced with mempool and onchain data
     Synced,
-    /// Failed to sync with mempool
+    /// Failed to sync with mempool and onchain data
     SyncFailed {
         error: String,
     },


dangeross (approved):
LGTM

lib/core/src/model.rs
```diff
@@ -501,6 +502,10 @@ pub enum SdkEvent {
     },
     /// Synced with mempool and onchain data
     Synced,
+    /// Failed to sync with mempool
```
danielgranhao, line 505: Nit: just to keep it matched with the `Synced` event
```suggestion
    /// Failed to sync with mempool and onchain data
```

danielgranhao (approved):

PR #1036, opened 2025-11-07
NWC: Add final features

hydra-yse:
This PR:
- Adds support for the `make_invoice` command
- Adds an optional expiry per NWC connection
- Adds an optional periodic balance per NWC connection

## TO-DOs
- [x] Properly expose, without runtime errors, the `BreezNwcService` struct through FRB

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -406,7 +406,7 @@ jobs:
       - name: Build Android bindings
         working-directory: lib/bindings
         run: |
-          cargo install gobley-uniffi-bindgen --git https://github.com/breez/gobley --rev f920c275e4f60e22080a6cd280d9562d64eb6ac9
+          cargo install gobley-uniffi-bindgen --git https://github.com/breez/gobley --rev 72753910f54a157de2f84b78dc6f78415e28d4f2
           gobley-uniffi-bindgen -c ./uniffi.toml -o langs/android/lib/src src/breez_sdk_liquid.udl
 
       - name: Run Android build
--- a/packages/react-native/yarn.lock
+++ b/packages/react-native/yarn.lock
@@ -9347,9 +9347,9 @@ __metadata:
   linkType: hard
 
 "node-forge@npm:^1":
-  version: 1.3.1
-  resolution: "node-forge@npm:1.3.1"
-  checksum: 08fb072d3d670599c89a1704b3e9c649ff1b998256737f0e06fbd1a5bf41cae4457ccaee32d95052d80bbafd9ffe01284e078c8071f0267dc9744e51c5ed42a9
+  version: 1.3.2
+  resolution: "node-forge@npm:1.3.2"
+  checksum: b6f905b0fcc39a2d59598e12ca2c071bfd760e56a9163aab8da7f8d6622547f8db60cfb2aefc39277d5a13af32e58573674b38107a2d4df7c243e12794839546
   languageName: node
   linkType: hard
 


roeierez (changes requested):
Looks good

lib/plugins/nwc/src/utils.rs
```diff
@@ -0,0 +1,8 @@
+use std::time::{SystemTime, UNIX_EPOCH};
```
dangeross: web_time?

lib/plugins/nwc/src/lib.rs
```diff
@@ -225,9 +259,52 @@ impl SdkNwcService {
             }
         };
 
+        if client.connection.receive_only && !matches!(req.params, RequestParams::MakeInvoice(_)) {
```
roeierez: Do we block also other commands such as ListTransactions on receive only?
Should we also return error to the caller here?

lib/plugins/nwc/src/lib.rs
```diff
@@ -330,11 +433,55 @@ impl SdkNwcService {
...
+
+    async fn min_refresh_interval(maybe_interval: &mut Option<Interval>) -> Option<()> {
+        match maybe_interval {
+            Some(interval) => {
+                interval.tick().await;
+                Some(())
+            }
+            None => None,
+        }
+    }
```
danielgranhao: We should test this carefully on Wasm node.js. I don't recall the details but according to #880 it may cause panics.

lib/plugins/nwc/src/persist.rs
```diff
@@ -26,30 +31,113 @@ impl Persister {
...
+        if connections.contains_key(&name) {
+            return Err(NwcError::generic(format!(
+                "Could not insert connection: `{name}` already exists"
+            )));
+        }
+        connections.insert(name, connection);
+        self.storage
+            .set_item(KEY_NWC_CONNECTIONS, serde_json::to_string(&connections)?)?;
+        Ok(())
+    }
```
danielgranhao: Could we get races here that cause one of the connections to end up not persisted?

danielgranhao: Same with `edit_nwc_connection` and `remove_nwc_connection`. Maybe this has been discussed in the past but I'm missing why it's safe.

lib/plugins/nwc/src/persist.rs
```diff
@@ -26,30 +43,142 @@ impl Persister {
...
+                break;
+            }
+            tokio::time::sleep(Duration::from_millis(500)).await;
+        }
+        if is_locked {
+            return Err(NwcError::generic("Could not acquire database lock"));
+        }
+        self.storage.set_item(KEY_NWC_LOCK, 1.to_string())?;
+        Ok(Lock { p: self })
+    }
```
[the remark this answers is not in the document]
hydra-yse: Yes, I was thinking as we did in the context of multiple SDK instances running (and therefore multiple NwcServices). In that case we cannot use a Mutex as it's not shared across plugin instances. But if that's overkill then yes, we can switch to a simple instance lock. Also yes, we should be careful about infinite locks, they can be set to 0 at every startup in case of a crash.

hydra-yse:
@dangeross I've fixed the CI errors and tested on WASM, everything is working as expected. I've also added ConnectionExpired/Refreshed events for better UX

lib/plugins/nwc/src/persist.rs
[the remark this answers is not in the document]
roeierez: What are the consequences of this race? expiry/refresh is something we can get over by periodic query.

lib/plugins/nwc/src/persist.rs
```diff
@@ -26,30 +43,180 @@ impl Persister {
...
+        }
+        let connection = connection.clone();
+        self.storage
+            .set_item(KEY_NWC_CONNECTIONS, serde_json::to_string(&connections)?)?;
+        Ok(connection)
+    }
+
+    pub(crate) async fn get_min_interval(&self) -> Option<u64> {
+        let get_min_connection_interval = async || -> NwcResult<Option<u64>> {
+            let _ = self.lock().await?;
```
dangeross: I think the race concern is then read/writing, so I don't think need to hold a lock when reading

lib/plugins/nwc/src/persist.rs
```diff
@@ -26,30 +43,180 @@ impl Persister {
...
+            .get_item(KEY_NWC_CONNECTIONS)?
             .unwrap_or("{}".to_string());
-        let uris = serde_json::from_str(&raw_uris)?;
-        Ok(uris)
+        let connections = serde_json::from_str(&connections)?;
+        Ok(connections)
+    }
+
+    pub(crate) async fn list_nwc_connections(&self) -> NwcResult<HashMap<String, NwcConnection>> {
+        let _ = self.lock().await?;
```
dangeross: Same here, I'm not sure its needed. Only when we plan to add/edit/remove connections

lib/plugins/nwc/src/persist.rs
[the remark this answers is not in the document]
roeierez: @dangeross  I discussed with @hydra-yse to change the implementation of set_item to optionally receive the old value as parameter and then only update if the old value equals to the current value ("get and set") which achieves atomicity without db locking.

lib/core/src/plugin/storage.rs
```diff
@@ -93,8 +96,28 @@ impl PluginStorage {
...
+        old_value: Option<String>,
+    ) -> Result<(), PluginStorageError> {
         let scoped_key = self.scoped_key(key);
+        if let Some(old_value) = old_value {
+            if let Some(current_value) = self.get_item(key)? {
+                if old_value != current_value {
+                    return Err(PluginStorageError::DataTooOld);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
hydra-yse: Solved in 2b048b9

roeierez (approved):
One small issue to retry otherwise LGTM

dangeross (commented):
Looks good. ~Is the TODO in the PR still valid?~

lib/core/src/plugin/storage.rs
```diff
@@ -93,11 +104,35 @@ impl PluginStorage {
...
+    ) -> Result<(), PluginStorageError> {
         let scoped_key = self.scoped_key(key);
-        self.get_persister()?
-            .update_cached_item(&scoped_key, self.encrypt(value)?)
-            .map_err(Into::into)
+        let persister = self.get_persister()?;
+        let mut con = persister.get_connection()?;
+        let tx = con.transaction_with_behavior(rusqlite::TransactionBehavior::Immediate)?;
+        if let Some(old_value) = old_value {
+            if let Some(current_value) = Persister::get_cached_item_inner(&tx, &scoped_key)? {
```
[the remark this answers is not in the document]
roeierez: I think the idea is that if we pass None we don't care about the old value, just simply update.

lib/plugins/nwc/src/event.rs
```diff
@@ -17,11 +17,18 @@ pub enum NwcEventDetails {
...
+    },
 }
 
 /// The event emitted when an NWC operation has been handled
 #[derive(Clone, Debug, PartialEq)]
 pub struct NwcEvent {
+    pub connection_name: Option<String>,
     pub event_id: Option<String>,
     pub details: NwcEventDetails,
 }
```
dangeross: I see sometimes the connection name is set in NwcEvent, sometimes set in the NwcEventDetails. It feels a bit inconsistent. I think now you've added `connection_name` in NwcEvent it can be removed from ConnectionExpired and ConnectionRefreshed events (Alternatively only add it to the events it is set). Wdyt?

hydra-yse: Makes sense, thanks Ross

lib/plugins/nwc/src/model.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// The creation time of the connection
+    pub created_at: u32,
+    /// Specifies whether this is a receive-only connection. Defaults to false.
+    pub receive_only: bool,
+    /// The expiry time of the connection
+    /// ## Dev Note:
+    /// If the expiry time is less than [crate::MIN_REFRESH_INTERVAL_SEC] seconds,
+    /// then it will take at most [crate::MIN_REFRESH_INTERVAL_SEC] seconds in order for the
+    /// connection to be deleted.
+    pub expiry_time_sec: Option<u32>,
```
erdemyerebasmaz: Shouldn't we expose `expiration_timestamp` instead? How does this translate to the end-user? Do they have to calculate expiration time by adding up `created_at` & `expiry_time_sec`?

lib/wasm/src/lib.rs
```diff
@@ -403,4 +393,11 @@ impl BindingLiquidSdk {
         self.sdk.disconnect().await?;
         Ok(())
     }
+
+    #[wasm_bindgen(js_name = "useNwcService")]
+    pub async fn use_nwc_service(&self, config: NwcConfig) -> WasmResult<BindingNwcService> {
```
dangeross:
```suggestion
    #[wasm_bindgen(js_name = "useNwcPlugin")]
    pub async fn use_nwc_plugin(&self, config: NwcConfig) -> WasmResult<BindingNwcPlugin> {
```

packages/flutter_breez_liquid/rust/src/sdk.rs
```diff
@@ -268,4 +260,10 @@ impl BreezSdkLiquid {
     pub async fn disconnect(&self) -> Result<(), SdkError> {
         self.sdk.disconnect().await
     }
+
+    pub async fn use_nwc_service(&self, config: NwcConfig) -> Result<BreezNwcService, SdkError> {
```
dangeross:
```suggestion
    pub async fn use_nwc_plugin(&self, config: NwcConfig) -> Result<BreezNwcPlugin, SdkError> {
```

lib/core/src/sdk.rs
```diff
@@ -5109,6 +5074,47 @@ impl LiquidSdk {
...
+    }
+
+    pub async fn stop_plugin(&self, plugin_id: &str) -> SdkResult<()> {
+        let mut plugins = self.plugins.lock().await;
+        let Some(plugin) = plugins.remove(plugin_id) else {
+            return Err(SdkError::generic(format!("Plugin {plugin_id} not found")));
+        };
+        plugin.on_stop().await;
+        Ok(())
+    }
```
dangeross: I don't see this being used

dangeross (commented):
It looks very good. I'm mostly looking top level now, so consistent naming. I think it would be good to name the exposed NWC plugin and functions to access it more clearly. So `SdkNwcPlugin`, `BindingNwcPlugin`, `BreezNwcPlugin`, `use_nwc_plugin`. Just to really cement and make it clear how things fit together.

lib/core/src/sdk.rs
hydra-yse: Yeah, it's just as another `stop` method to avoid calling `on_stop` from the plugin directly, or if a stop isn't exposed at all in the bindings. Also adds symmetry to the start_plugin call. You think it should be removed?

dangeross: So it can be safely removed from the SDK? I just don't see it in the binding interfaces, or is it more something the plugin should be calling? (e.g. a disconnect() fn in the plugin should call it?)

PR #1042, opened 2025-11-24
NWC: Add offline notifications support

hydra-yse:
Adds offline notifications support for both the Kotlin and Swift notification plugins

### DIFF
--- a/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/NwcEvent.kt
+++ b/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/NwcEvent.kt
@@ -39,13 +39,13 @@ class NwcEventJob(
 
     override fun start(liquidSDK: BindingLiquidSdk, pluginConfigs: PluginConfigs) {
         val nwcService = PluginManager.nwc(liquidSDK, pluginConfigs, logger)
-        if (nwcService == null) return;
+        if (nwcService == null) return
         nwcService.addEventListener(this)
         try {
             val decoder = Json { ignoreUnknownKeys = true }
             var notification = decoder.decodeFromString(NwcEventNotification.serializer(), payload)
-            nwcService.handleEvent(notification.eventId)
             eventId = notification.eventId
+            nwcService.handleEvent(notification.eventId)
         } catch (e: Exception) {
             logger.log(TAG, "Failed to process NWC event: ${e.message}", "WARN")
             notifyChannel(
--- a/lib/bindings/langs/swift/Sources/BreezSDKLiquid/Task/NwcEvent.swift
+++ b/lib/bindings/langs/swift/Sources/BreezSDKLiquid/Task/NwcEvent.swift
@@ -29,8 +29,8 @@ class NwcEventTask: TaskProtocol, NwcEventListener {
         var request: NwcEventNotification? = nil
         do {
             request = try JSONDecoder().decode(NwcEventNotification.self, from: self.payload.data(using: .utf8)!)
-            try nwcService.handleEvent(eventId: request!.eventId)
             eventId = request!.eventId
+            try nwcService.handleEvent(eventId: request!.eventId)
         } catch let e {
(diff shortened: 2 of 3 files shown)


lib/bindings/langs/swift/Sources/BreezSDKLiquid/Task/NwcEvent.swift
```diff
@@ -0,0 +1,37 @@
...
+            request = try JSONDecoder().decode(NwcEventNotification.self, from: self.payload.data(using: .utf8)!)
+            try plugins.nwc!.handleEvent(eventId: request!.eventId)
+        } catch let e {
+            self.logger.log(tag: TAG, line: "failed to decode payload: \(e)", level: "ERROR")
+            throw e
+        }
+    }
+
+    func onEvent(e: SdkEvent) {}
+    func onShutdown() {}
```
dangeross: Same here, on iOS we definitely need to respond to the notification request with displaying notification content

lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/job/NwcEvent.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    override fun onEvent(event: NwcEvent) {
+        if (eventId == null || eventId != event.eventId) {
+            return
+        }
+        var eventName: String
+        when(event.details) {
+            is NwcEventDetails.GetBalance -> eventName = "get_balance"
+            is NwcEventDetails.ListTransactions -> eventName = "list_transactions"
+            is NwcEventDetails.PayInvoice -> eventName = "pay_invoice"
```
dangeross: Title case is better, it's being displayed to the end user
```suggestion
            is NwcEventDetails.GetBalance -> eventName = "Get Balance"
            is NwcEventDetails.ListTransactions -> eventName = "List Transactions"
            is NwcEventDetails.PayInvoice -> eventName = "Pay Invoice"
```

lib/bindings/langs/swift/Sources/BreezSDKLiquid/SDKNotificationService.swift
```diff
@@ -26,16 +27,20 @@ open class SDKNotificationService: UNNotificationServiceExtension {
...
         if let currentTask = self.getTaskFromNotification() {
             self.currentTask = currentTask
             
             DispatchQueue.main.async { [self] in
                 do {
                     logger.log(tag: TAG, line: "Breez Liquid SDK is not connected, connecting...", level: "INFO")
                     liquidSDK = try BreezSDKLiquidConnector.register(connectRequest: connectRequest, listener: currentTask)
+                    let pluginConfigs = getPluginConfigs()
+                    if let nwcConfig = pluginConfigs?.nwc {
+                        plugins.nwc = try NwcPlugin.register(sdk: liquidSDK!, config: nwcConfig, listener: currentTask)
```
dangeross: I feel like we need a cleaner way to cache the current plugins. It's like we're mixing up how we initialize and cleanup. Maybe we need a plugin manager that takes the pluginConfig on to register the plugins. If the plugins are already registered, return them instead. Have an unregister method to cleanup all plugins. 

Maybe you can repurpose SDKPlugins for this? Wdyt?

dangeross (commented):
Looks good, just one comment

hydra-yse:
@roeierez @dangeross please review when you can, only thing left to fix is making the concurrent messaging more robust. Atm if you spam events (e.g. multiple zaps in the span of few seconds) only half seem to actually get processed by the wallet. Will check exactly why.

lib/plugins/nwc/src/context.rs
```diff
@@ -127,4 +128,9 @@ impl RuntimeContext {
             warn!("Could not send info event to relay pool: {err:?}");
         }
     }
+
+    /// Returns true when value was inserted, or false otherwise
+    pub async fn try_insert_replied_event(&self, event_id: String) -> bool {
```
dangeross: Doesn't `try_` normally imply it returns a Result?

PR #1045, opened 2025-12-02
NWC: Add final features

hydra-yse:
This PR:
- Adds support for the `make_invoice` command
- Adds an optional expiry per NWC connection
- Adds an optional periodic balance per NWC connection

## TO-DOs
- [x] Properly expose, without runtime errors, the `BreezNwcService` struct through FRB

## Note
Re-opening as previous PR was stuck in stale state

### DIFF
--- a/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/Constants.kt
+++ b/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/Constants.kt
@@ -36,6 +36,7 @@ object Constants {
     const val MESSAGE_TYPE_LNURL_PAY_INVOICE = "lnurlpay_invoice"
     const val MESSAGE_TYPE_LNURL_PAY_VERIFY = "lnurlpay_verify"
     const val MESSAGE_TYPE_SWAP_UPDATED = "swap_updated"
+    const val MESSAGE_TYPE_NWC_EVENT = "nwc_event"
 
     // Resource Identifiers
     const val DISMISSIBLE_NOTIFICATION_CHANNEL_DESCRIPTION =
@@ -95,6 +96,8 @@ object Constants {
         "swap_confirmed_notification_failure_title"
     const val SWAP_CONFIRMED_NOTIFICATION_TITLE =
         "swap_confirmed_notification_title"
+    const val NWC_SUCCESS_NOTIFICATION_TITLE = "nwc_success_notification_title"
+    const val NWC_FAILURE_NOTIFICATION_TITLE = "nwc_failure_notification_title"
 
     // Resource Identifier Defaults
     const val DEFAULT_DISMISSIBLE_NOTIFICATION_CHANNEL_DESCRIPTION =
@@ -149,4 +152,6 @@ object Constants {
         "Tap to complete payment"
     const val DEFAULT_SWAP_CONFIRMED_NOTIFICATION_FAILURE_TITLE =
         "Payment Pending"
+    const val DEFAULT_NWC_SUCCESS_NOTIFICATION_TITLE = "NWC %s success"
+    const val DEFAULT_NWC_FAILURE_NOTIFICATION_TITLE = "NWC request failed"
 }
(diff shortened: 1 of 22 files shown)


packages/wasm/examples/node/src/cli.js
```diff
@@ -264,6 +267,28 @@ const initCommand = () => {
...
     program.command('unregister-webhook').description('Unregister the webhook URL').action(unregisterWebhook)
 
+    const nwc = program.command('nwc').description('Run NWC features')
+    nwc
+        .command("add-connection")
+        .description("Adds an NWC connection")
+        .addArgument(new Argument('<name>', 'The unique identifier of the connection'))
+        .addOption(new Option('--receive-only', 'Whether the connection is receive-only'))
+        .addOption(new Option('--expiry <number>', 'The expiry time of the connection, in seconds').argParser(parseInt))
+        .addOption(new Option('--period-time <number>', 'The amount of time a periodic budget lasts, in seconds').argParser(parseInt))
```
dangeross:
```suggestion
        .addOption(new Option('--expiry <number>', 'The expiry time of the connection, in minutes').argParser(parseInt))
        .addOption(new Option('--period-time <number>', 'The duration of a periodic budget before it renews, in minutes').argParser(parseInt))
```

dangeross (commented):
Super close 🏁

dangeross (commented):
I thought we were feature complete... 😅

lib/plugins/nwc/src/lib.rs
```diff
@@ -181,53 +218,114 @@ impl SdkNwcService {
...
 
         // Verify the event has not expired
         if event
             .tags
             .expiration()
             .is_some_and(|t| *t > Timestamp::now())
         {
-            warn!("Event {} has expired. Skipping.", event.id);
-            return;
+            return Err(NwcError::EventExpired);
```
roeierez: I think the expiration comparison is wrong.  Seems like it should be:
`.expiration().is_some_and(|t| *t < Timestamp::now())`

lib/plugins/nwc/src/model.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// The latest budget update time (last reset time)
+    pub updated_at: u32,
+}
+
+impl PeriodicBudget {
+    fn from_budget_inner(b: PeriodicBudgetInner, created_at: u32) -> Self {
+        Self {
+            used_budget_sat: b.used_budget_sat,
+            max_budget_sat: b.max_budget_sat,
+            renews_at: b.renewal_time_sec.map(|t| created_at + t),
```
roeierez: should we consider last_reset_time (when exists) rather than created_at when calculating renews_at?

roeierez (changes requested):

lib/plugins/nwc/src/persist.rs
```diff
@@ -18,38 +25,241 @@ impl Persister {
...
+            .get_item(KEY_NWC_CONNECTIONS)?
             .unwrap_or("{}".to_string());
-        let uris = serde_json::from_str(&raw_uris)?;
-        Ok(uris)
+        let connections = serde_json::from_str(&connections)?;
+        Ok(connections)
+    }
+
+    pub(crate) fn remove_nwc_connection(&self, name: String) -> NwcResult<()> {
+        self.set_connections_safe(|connections| {
```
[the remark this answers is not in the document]
hydra-yse: I don't think there's a way we can do this atomically unfortunately, how should we proceed?

dangeross (commented):
I think you need to adjust the use of the useDescriptionHash in the notification plugin also

dangeross (approved):

lib/plugins/nwc/src/handler.rs
```diff
@@ -32,8 +52,135 @@ impl SdkRelayMessageHandler {
...
+            Ok(preimage) => {
+                let fees_paid = response.payment.fees_sat * 1000; // Convert sats to msats
+                Ok(PayInvoiceResponse {
+                    preimage,
+                    fees_paid: Some(fees_paid),
+                })
+            }
+            Err(err) => Err(anyhow!("Could not retrieve payment preimage: {err}")),
+        };
+        let _ = self.sdk.remove_event_listener(listener_id).await;
```
hydra-yse: TODO: Need to make sure that we unsubscribe from the event listener independently from success/failure

hydra-yse: Fixed in ff46a04

Issue #892, opened 2025-04-21
Optimize CI jobs that rely on regtest


danielgranhao:
With the changes introduced in #846, starting the regtest environment can take 16+ minutes. We should optimize it to reduce job times. 

Things to improve:

- Explore caching
- Publish images instead of building them every time (ssl-proxy and waterfalls)

PR #1047, opened 2025-12-22
SDK: Bump rusqlite to v0.38.0

hydra-yse:
The Spxg/rusqlite fork has been deleted, therefore I have pointed the rusqlite dependency to ~the official package~ our own fork, version 0.38.0.
There's been a breaking change in the newest release where u64s are not supported by default anymore. To fix this, I've manually mapped every db read/write of a u64 to an i64, with failure. This results in a loss of precision of 2^63 values for the final missed bit, but it's an acceptable tradeoff as this edge-case is rarely encountered.

Also:
- [X] Updated CLI to fix clippy issues on rustc 1.92.0
- [X] ~Need to fix WASM compilation issue due to unrecognized cfg flags~ **Edit:** Turns out that the wasm-bindgen macros in sdk-macros need updating with wasm-bindgen versions >0.2.101 (and wasm-bindgen-test versions >0.3.51), so for now I've pinned all our rusqlite dependencies and sub-dependencies to that version via Breez fork. Temporary fix.
- [x] Bumped waterfalls service to latest production version (yse-production-v2 branch) and moved both waterfalls and rt-sync to docker.io image for faster CI (this tripled the regtest startup speed)

### DIFF
--- a/packages/flutter_breez_liquid/rust/Cargo.lock
+++ b/packages/flutter_breez_liquid/rust/Cargo.lock
@@ -696,6 +696,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sideswap_api",
+ "sqlite-wasm-rs",
  "strum 0.25.0",
  "strum_macros 0.25.3",
  "tempfile",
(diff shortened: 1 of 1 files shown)


regtest/docker-compose.yml
danielgranhao: This relates to issue #892, but I'm not sure if this change is worth adding in this PR. 

Using published images adds more margin for CI to use unexpected versions. If possible, I would prefer to use caching to improve this.

hydra-yse:
> Also, please check the failing CI tests. I don't think the failures in the non-wasm tests are expected.

I noticed they were happening due to timeouts being too stringent, increasing the timeouts should have made it more deterministic (see 8d8fcdc)

> I would prefer to use caching to improve this.

Sure, I'll look into it!

PR #1053, opened 2026-01-19
Remove testnet support from the SDK

### LINKED ISSUE #1051: Deprecate Testnet3 support
The SDK's testnet network uses Testnet3, which has been broken for some time. Boltz has already [deprecated it themselves](https://testnet.boltz.exchange/), and so should we.

danielgranhao:
Closes #1051 

Testnet (testnet3) is not usable. This change prevents the SDK from being used with testnet while keeping the LiquidNetwork::Testnet enum variant for potential future testnet4 support.

Changes:
- Add SdkError::NetworkNotSupported error variant
- Return error when attempting to connect with testnet configuration
- Return error from default_config when testnet is requested
- Remove Config::testnet() and Config::testnet_esplora() methods
- Update CLI default network from testnet to mainnet
- Update test utilities to use regtest instead of testnet
- Update WASM default_config to return error for testnet

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -28,19 +28,16 @@ To set a specific network, use one of
 
 ```bash
 cargo run -- --network mainnet
-cargo run -- --network testnet
+cargo run -- --network regtest
 ```
+
+> **Note:** Testnet is not currently supported. The `LiquidNetwork::Testnet` variant exists for potential future testnet4 support but attempting to connect will result in an error.
 
 To specify a custom data directory, use
 
 ```bash
 cargo run -- --data-dir temp-dir
 ```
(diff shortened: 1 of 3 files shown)


dangeross:
> @dangeross Do you think [a8e1158](https://github.com/breez/breez-sdk-liquid/pull/1053/commits/a8e11581cadfa3b0e994ad7384c9d446d5858dcd) makes sense? The previous patch-based approach seemingly can break on SDK API changes

Seems the CI passes and looks more maintainable 👍 Did you validate the patch is applied?

dangeross (approved):
LGTM

PR #1057, opened 2026-01-26
add logs surrounding lnurl notifications

JssDWt:
Add some logs surrounding invoice creation in lnurl notifications on iOS.


erdemyerebasmaz:
Merged as part of #1067

Issue #1062, opened 2026-01-28
Chain swap temporarily marked as refundable when fee acceptance fails


danielgranhao:
When accepting swap fees fails, a chain swap may be temporarily classified as refundable:
  - Triggers Refundable events
  - Listed in list_refundables()

  Shortly after, the recoverer corrects the state back to WaitingFeeAcceptance.

  Problem: No mechanism exists to communicate that a refundable is no longer refundable. Consumers may act on stale refundable state.

  Possible fixes:
  1. Add event/mechanism to retract refundables
  2. Prevent swaps from ever being temporarily marked refundable (better, but feasibility unclear)

PR #1073, opened 2026-02-14
 fix: Prevent zero-amount swaps from being marked refundable on updat…

Abhay349:
**Fixes**- #1062 
This PR adds an early return when `handle_amountless_update` fails. Instead of falling through to the irreversible failure logic, we now:
-Log the error.
-Return early, leaving the swap in its current state.
-Allow the background recoverer (specifically `ChainReceiveSwapHandler`) to pick up the swap on the next cycle and correctly derive the state, avoiding the confusing intermediate Refundable state.

### DIFF
--- a/lib/core/src/chain_swap.rs
+++ b/lib/core/src/chain_swap.rs
@@ -350,14 +350,12 @@ impl ChainSwapHandler {
                 let is_zero_amount = swap.payer_amount_sat == 0;
                 if matches!(swap_state, ChainSwapStates::TransactionLockupFailed) && is_zero_amount
                 {
-                    match self.handle_amountless_update(swap).await {
-                        Ok(_) => {
-                            // Either we accepted the quote, or we will be waiting for user fee acceptance
-                            return Ok(()); // Break from TxLockupFailed branch
-                        }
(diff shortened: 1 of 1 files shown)


Abhay349:
Hi @danielgranhao,
Please let me know if any changes or improvements are needed. I’m happy to update it accordingly based on yrr guidance.
Thanks :)

lib/core/src/chain_swap.rs
```diff
@@ -353,11 +353,12 @@ impl ChainSwapHandler {
                     match self.handle_amountless_update(swap).await {
                         Ok(_) => {
                             // Either we accepted the quote, or we will be waiting for user fee acceptance
-                            return Ok(()); // Break from TxLockupFailed branch
                         }
-                        // In case of error, we continue and mark it as refundable
+                        // In case of error, we log the error but don't mark as refundable,
+                        // letting the recovery logic handle the state.
                         Err(e) => error!("Failed to accept the quote for swap {}: {e:?}", &swap.id),
                     }
```
danielgranhao: We can simplify this into a `if let Err(e) = ...` block

Issue #997, opened 2025-08-11
Wasm E2E tests failing


danielgranhao:
Since commit 3cd21b9b0142d23081ac31c6a8f0ae703a36da7a the Wasm E2E tests have been failing.

Issue #847, opened 2025-04-01
WASM: events occasionally skipped


danielgranhao:
As shown by the tests proposed in https://github.com/breez/breez-sdk-liquid/pull/846, on Wasm, sometimes events are not emitted. Needs further investigation to know the cause.

Issue #829, opened 2025-03-28
Receive swap paid using MRH occasionally seen as Liquid payment


danielgranhao:
The end-to-end tests proposed in #816 show that occasionally a receive swap paid using MRH is shown as a Liquid payment (it should still be Lightning). This is likely temporary (needs further investigation).

PR #1094, opened 2026-04-10
fix: eliminate chain swap test flakiness (#996) + WASM test flakiness (#997)

### LINKED ISSUE #996: Chain swap E2E test flakiness
The E2E `bitcoin` tests have become flaky since we reduced the frequency of querying the bitcoin tip (#978).

maggo83:
## Summary

Closes #996, #997 and #847. References #828, #829, #978

### Root cause analysis (#996)

Reasons the chain swap E2E tests were flaky:
1. Chain tip gating (#978) + indexer processing times → race condition into deadlock (also root-cause of #847)
2. "Early" (not synced mempool readiness) mining → unconfirmed TX → test condition timeout
3. Unexpected reaction/output from esplora: same tx reported as mined and in mempool (see #1097) → degradation into test timeout
4. Even after SDK has advanced internal states, further LWK updates of wallet balances are delayed (see #828) → race condition into test failure

Details:
1. Base situation: tests use one function to mine one bitcoin and one liquid block to advance state. However different propagation times of Bitcoin and Liquid blocks through their indexers (esplora/electrs) combined with async background polling for chain states by the SDK can lead to a situation where:
   1.1. The Liquid block is indexed before the Bitcoin block
   1.2. The SDK's background polling (`track_new_blocks()`) updates swap states but does not receive/handle payment because the Bitcoin block is not yet detected (because it was not indexed when polling detected the liquid block). But the liquid block was indexed and detected and "consumed" by the gating introduced in #978
   1.3. Subsequently no new blocks get mined and the polling will never detect the latest Bitcoin block, because it is gated according to #978 → no further advances in system state (deadlock → timeout)
   1.4. Even if only the Liquid tx/block is relevant, it is not possible to guarantee the SDK's background poller runs after the block has been fully processed by all relevant indexers and might consume the Liquid tip w/o retrieving all the relevant information from it. (Most likely root-cause of #847 as well)

   **Effect in production:**
   - The deadlock gets auto-resolved after a new Liquid block (~1 min) is mined
   - Detection of new Bitcoin blocks can be delayed by one Liquid block
   - This trade-off appears acceptable given that swap timeouts >> 1 min, and was probably considered when creating #978
   - *Behavior is acceptable in production*

   **Effect in regtest:**
   - No automatic block generation → permanent deadlock → test hangs until timeout

2. Some test steps expect one of the test entities (SDK or Boltz) to broadcast a TX and then mine a block to confirm that TX w/o confirming that the TX is actually in the mempool already

   **Effect in production:** No negative effect as this TX will be included in one of the next blocks. Not a fault of SDK.
   **Effect in regtest:** No new blocks are produced → test waits with timeout on some reaction to the assumed newly mined transaction → timeout

3. `determine_incoming_lockup_and_claim_txs()` checks via the blockexplorer how many transactions are available for a certain script (swap claim). It assumes the first tx is the lockup tx, the second is the claim tx. If more than 2 tx are reported it assumes this to be strange and does nothing, triggering the SDK to go into a 120s grace period where it skips recovery. However esplora sometimes reports a freshly mined claim tx twice in one query: once still as in the mempool and once freshly confirmed, thus triggering the "this is strange" path which suppresses further updates of the swap state (unless some other conditions are met, e.g. new liquid block is mined), which exceeds the test's timeout.

   **Effect in production:** After the next liquid block (60s < 120s grace period) the state is healed and claim continues to finalized. Acceptable, but UX could be improved.
   **Effect in regtest:** No new blocks produced → timeout → failed test

4. Even when the swap is detected as successfully completed, SDK still has to wait on LWK to complete its queries to update the wallet balances completely (see #828). Tests assumed that immediately after "PaymentSucceeded" the wallet balances are up-to-date → stale wallet balances do not match expectation → failed test

   **Effect in production:** Depending on the propagation delay the user might not even notice. Otherwise heals once delay has passed. Not critical.
   **Effect in regtest:** Test fails due to unexpected wallet amounts after "PaymentSucceeded"

### Proposed solution (#996)

General guidelines/goals:
1. Do not change production code if there is no bug
2. Make implicit assumptions better visible by converting them to explicit checks
3. Make system state at each test step predictable/understood

This includes:
* Enable the test environment to have more granular control over the mining process and timing:
  - Add option to mine on either chain (Bitcoin, Liquid) or both
  - Add option to wait for indexer(s) to catch up before continuing
  - Extract expected TX id and utilities to wait until they are in the mempool
  - Add a utility to drain mempools at the end of a test section
  - Move (series of) common checks into utility functions such that individual checks are not forgotten in a specific test due to copy&paste

### Discarded options

| Option | Why discarded |
|--------|---------------|
| **Continuous simulated mining** — Run a background process that keeps producing blocks on both chains at a fixed interval (similar to production) | Would increase pass rate but make test behavior *less* predictable and harder to debug. Failures would become timing-dependent rather than logic-dependent. Contradicts guideline 3 (predictable system state at each step). |
| **Randomly mine extra blocks** after each test step, unless explicitly documented why they are needed | Same argument as above: it papers over root causes instead of making them visible. Contradicts guidelines 2 and 3. |
| **Modify `sync()` to drop the gating from #978** or to also call `on_x_block` handlers | `sync()` is not expected to trigger externally visible actions (as calling block handlers would). Would violate guideline 1 (no production code changes unless there is a bug). The gating is working as designed; the issue is purely a regtest artifact. |

### Specific solutions per root cause

1. **Race between indexers / background poller**
   1.1. When a bitcoin tx needs to be noticed: first mine one bitcoin block (and wait for indexers to catch up), then mine one liquid block. This way the liquid block will trigger gating only after the needed data is available.
   1.2. When a liquid tx needs to be noticed: either mine one liquid block waiting for indexers and then a second. However that can introduce further race conditions if after the first mined block a test entity (e.g. SDK or Boltz) is issuing the next tx which might already be in the second mined block. To avoid having to track two possible states, this is circumvented by first assuming the block triggers the expected reaction and waiting with a timeout a bit longer than the background poller's cycle time. Only if that times out, mine another liquid block as a last retry. This way the state of the test is always clear and subsequent steps have one relevant state they can assume.

2. **Mempool readiness** — Get expected TX id via available interfaces and explicitly wait for the TX to be in the mempool before mining.

3. **Esplora duplicate reporting** — Proposed long-term solutions:
   3.1. Verify with esplora team if this is expected behavior → raise issue?
   3.2. Make SDK more robust by explicitly checking if relevant tx ids are in the mempool or on chain (regardless of number of txs for that script)
   Current workaround in test: break "skipping_recovery" by mining one bitcoin block (to gain time for indexers to settle) and one liquid block (to give background poller another chance). Marked as "Workaround for #1097".

4. **LWK balance propagation delay** — No publicly available interface found to check if LWK update is complete. So fallback to "Mine one more liquid block to force an update of LWK wallet data". (Possible future approach skipped due to PlanB assignment deadline: poll wallet balance until it meets expected value instead of mining a new block)

### Root cause analysis (#997)

Many issues in WASM environment were fixed by measures introduced for #996. Remaining:

When SDK from WASM tries to upgrade connection to swapproxy to websocket, the request gets declined because the swapproxy is configured to reject upgrades for connections not coming from the same origin. WASM SDK always has a different origin than swapproxy.

**Effect in production:** WASM SDK clients connect from unpredictable origins → WebSocket connection fails.
**Effect in regtest:** WASM tests systematically fail if they rely on the websocket connection working.

### Proposed solution (#997)

Swapproxy: accept WebSocket connections from any origin. Add `OriginPatterns: ["*"]` to allow all origins. Security is maintained by the existing certificate-based API key authentication (`Bearer` token on HTTP(S), first-message `apikey` on WS), which is independent of the Origin header.

**Discarded option:** Explore and incorporate further upstream updates/changes in the swapproxy sub-repo. This was not chosen because the scope of this task was to first establish a stable test environment. The minimal change achieves that. Other changes can be more easily evaluated against this stable baseline afterward. (Also note the swapproxy PR is made against a fork due to the limited scope/timeline of the PlanB programme assignment.)

**Depends on: [breez/swapproxy#16](https://github.com/breez/swapproxy/pull/16)**

> **Note:** The `.gitmodules` URL temporarily points to `maggo83/swapproxy` so the submodule commit resolves. Revert to `breez/swapproxy` once breez/swapproxy#16 is merged.

### Added test (bitcoin amountless receive) + Findings/Fixes (Skip overwriting swaps after fees accepted)

**Not added test cases:**
* Unhappy paths (e.g. boltz going down/rogue -> reclaim via time-out):
  *  Although high value add also complexity to manipulate connection in current regtest setup hard to do. But in my opinion should be added sooner rather than later.
* Bolt12 actual send/receive paths:
   * Not yet fully supported by SDK itself. Adding CLN to regtest also considered too high effort for given amount of time (no prior experience with this so just a gut feel). Will need to be added once feature ready

**Actually added testcase: Bitcoin amountless receive**
* Interesting because boltz initially reports a failed swap but then SDK "fixes" this -> different behavior than simple receive
* Test revealed a possible weakness in the code: repeated triggers from boltz can overwrite an already aligned state of a swap and hence lead to swap state regression 
* As this is considered a bug/robustness issue, a fix is proposed: make repeated failedLockup idempotent. (See <COMMIT ID>)

### Additional Bugfix: wrong block height in receive_swap_handler

- `receive_swap_handler.on_bitcoin_block()` was incorrectly called with `*current_liquid_block` instead of `*current_bitcoin_block` in `sdk.rs`. Found during code review.

**Effect in production:** None as this input seems to be ignored anyway.
**Effect in testing:** None.

## Validation

Ran multiple iterations of the full test suite (native, browser, node — all test types per iteration) on branch rebased to current main: without regressions.

This depends on the fixes
* swapproxy allows all origins (otherwise WASM tests fail)
* make repeated failedLockup idempotent (otherwise bitcoin amountless receive test is flaky)

#847 was no longer encountered in the test runs, hence assumed closed as well.

## Known limitations

The amounts in the send/receive tests are not symmetric. I.e. per iteration of running the test suite wallet/lightning channel balances get shifted. When running the testsuite repeatedly this can lead to their depletion and then either changed behavior (e.g. boltz switching to direct transactions) or no more sends being possible.

**Effect:** After running > 100 test iterations in a row w/o resetting/restarting regtest or manually rebalancing wallets/channels, tests will start systematically failing afterwards (currently bolt11.rs fails first after 116 repeated executions).

**Proposed solution:** No change. Running test cases repeatedly was only needed to check for flakiness which is solved with this PR. If needed again later for some reason, one can easily circumvent this by stopping/starting regtest after some iterations. Also just due to the TX fees in the (very) end this issue will appear sooner or later. Manually rebalancing after each test seems not worth the effort.

#829 is not fully analysed or fixed although referenced in the tests. Initial analysis (not verified yet) points to two possible issues:
* possible race condition (to be added to internal mrh_tx database in handle_receive_swap.rs, the tx has to have a timestamp _strictly_  > the swap's timestamp (resolution seconds). In fast regtest the "=" case [same second mined as swap initiated] might happen)
* insert_or_update_receive_swap_inner has no COALESCE protection unlike try_handle_receive_swap_update

## Further proposals

* The race between background poller and indexers after a (liquid) block is found can delay reaction of the SDK to new chain-info by ~1 min. Maybe it can be considered to change the gating to be not only on directly newly received liquid block but also one cycle after, to give other propagation depended subsystems time to settle / update. This would increase calls by a factor of 2, but the overall saving from #978 (+this) would still be substantial while ensuring high responsiveness of SDK
* In a similar vein: several state transitions are triggered by WS inputs from boltz and fail w/o retry if some information fetched from SDK external sources is not as expected (which might be caused by small propagation delays). Maybe it should be considered to also here add one try after a small settle time to make the overall flows more robust.

## Commits

1. **fix(regtest): wait for block indexer after mining in chain swap test (#996)** [`2a5d333d`] — Replace non-deterministic `sleep(5)` waits with deterministic indexer-readiness checks. Add `Chain` enum, `mine_and_index_blocks()`, `mine_bitcoin_then_liquid()` utilities. Addresses root cause 1 (indexer race / gating deadlock).
2. **fix: pass bitcoin block height to receive_swap_handler.on_bitcoin_block** [`b2c4ed23`] — One-line fix in sdk.rs: `on_bitcoin_block(*current_liquid_block)` → `on_bitcoin_block(*current_bitcoin_block)`. Found during code review.
3. **fix(regtest): eliminate mempool races in chain swap test (#996)** [`b7385eec`] — Add mempool polling helpers (`wait_for_tx_in_mempool`, `poll_boltz_server_lockup_txid`). Wait for each broadcast to reach the mempool before mining its confirming block. Addresses root cause 2.
4. **fix(regtest): improve robustness of bitcoin native test** [`6bfbdedf`] — Handle poller/indexer desync with try-then-mine retry pattern for `PaymentWaitingConfirmation`. Add esplora flakiness workaround (extra block to break `skip_recovery`). Add `drain_mempools()` cleanup between test sections. Addresses root causes 3 and 4.
5. **fix(regtest): roll out stricter test checks to bitcoin SEND and REFUND tests** [`b59b7768`] — Add `poll_boltz_zero_conf_accepted` utility. Apply mempool-aware mining and explicit tx-id tracking to SEND and REFUND sections. Add `prepare_refund` assertions and RBF fee validation.
6. **chore: update swapproxy submodule to accept all WebSocket origins** [`013e1018`] — Points to maggo83/swapproxy@781171c. Depends on [breez/swapproxy#16](https://github.com/breez/swapproxy/pull/16).
7. **chore(regtest): document zero-conf coverage constraints** [`a303ca46`] — Add inline comments explaining why the zero-conf acceptance path cannot be fully covered in the current regtest setup.
8. **fix(regtest): wait for electrs to sync after mining** [`68e42a30`] — Ensure electrs has indexed the freshly mined block before the test proceeds. Complements commit 1 for cases where esplora is the active indexer.
9. **test: add amountless receive E2E test** [`eb28e28f`] — New test for the bitcoin chain swap amountless receive flow (BIP-21 / no pre-specified amount). Exercises the auto-accept fee path and exposes the repeated `lockupFailed` regression fixed in commit 10.
10. **fix: make repeated lockupFailed idempotent after zero-amount fee acceptance** [`53905546`] — Production fix: prevent a subsequent `lockupFailed` WebSocket message from overwriting swap state after fees have already been accepted, avoiding state regression.
11. **test(regtest): introduce wait-with-one-retry helper for robustness** [`0ac393a6`] — Add `wait_for_event_with_retry()`. Replaces ad-hoc mine-and-retry patterns across tests with a consistent, documented utility.
12. **test: harden bolt11 & bolt12 regtest tests** [`8cfb1670`] — Apply mempool-aware mining, indexer waits, and stricter payment/wallet assertions to bolt11.rs and bolt12.rs.
13. **test: harden liquid regtest tests** [`e195ffea`] — Apply the same robustness improvements to the liquid send/receive test flows.
14. **test: extract assert_payment and assert_wallet_settled helpers** [`a4b9e8c2`] — Extract repeated payment verification and settled-balance assertion patterns into shared utilities to prevent omissions across test sections.
15. **test(regtest): add assert_wallet_pending helper and apply across all flows** [`f612bd1d`] — Add `assert_wallet_pending()` for mid-flight balance checks; apply consistently across all test flows.
16. **refactor(regtest): trim verbose comments and tighten test structure** [`9ad900b2`] — Remove scaffolding comments left from development, tighten overall test file structure for readability.

### DIFF
--- a/lib/core/tests/regtest/bolt11.rs
+++ b/lib/core/tests/regtest/bolt11.rs
@@ -1,10 +1,7 @@
-use std::time::Duration;
-
 use breez_sdk_liquid::model::{
     PaymentDetails, PaymentState, PaymentType, PrepareReceiveRequest, PrepareSendRequest, SdkEvent,
 };
 use serial_test::serial;
-use tokio_with_wasm::alias as tokio;
 
 use crate::regtest::{
     utils::{self},
@@ -39,6 +36,19 @@ async fn bolt11_esplora() {
     bolt11(handle_alice, handle_bob).await;
 }
 
+#[sdk_macros::async_test_not_wasm]
+#[serial]
+#[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+async fn bolt11_mixed() {
+    let handle_alice = SdkNodeHandle::init_node(ChainBackend::Electrum)
+        .await
+        .unwrap();
+    let handle_bob = SdkNodeHandle::init_node(ChainBackend::Esplora)
+        .await
+        .unwrap();
+    bolt11(handle_alice, handle_bob).await;
+}
+
 async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) {
     handle_alice
         .wait_for_event(|e| matches!(e, SdkEvent::Synced { .. }), TIMEOUT)
@@ -48,6 +58,14 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
         .wait_for_event(|e| matches!(e, SdkEvent::Synced { .. }), TIMEOUT)
         .await
         .unwrap();
+
+    // Merge indexers from both handles: if either node uses an indexer, we must wait for it.
+    let indexers = utils::Indexers {
+        btc_esplora: handle_alice.indexers.btc_esplora || handle_bob.indexers.btc_esplora,
+        btc_electrs: handle_alice.indexers.btc_electrs || handle_bob.indexers.btc_electrs,
+        lbtc_esplora: handle_alice.indexers.lbtc_esplora || handle_bob.indexers.lbtc_esplora,
+        waterfalls: handle_alice.indexers.waterfalls || handle_bob.indexers.waterfalls,
+    };
 
     // -------------------RECEIVE SWAP-------------------
     let payer_amount_sat = 200_000;
@@ -77,7 +95,10 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
         receiver_amount_sat
     );
 
-    utils::mine_blocks(1).await.unwrap();
+    // Confirm the server lockup and wait for swap to complete
+    utils::mine_and_index_blocks(1, utils::Chain::Liquid, Some(&indexers))
+        .await
+        .unwrap();
 
     utils::wait_for_event_with_retry(
         &mut handle_alice,
@@ -88,9 +109,11 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
     .await
     .unwrap();
 
-    // TODO: this shouldn't be needed, but without it, sometimes get_balance_sat isn't updated in time
-    // https://github.com/breez/breez-sdk-liquid/issues/828
-    tokio::time::sleep(Duration::from_secs(1)).await;
+    // Workaround for #828: mine an extra Liquid block so that sync() sees a
+    // new tip and takes the full scan path, refreshing the LWK wallet state.
+    utils::mine_and_index_blocks(1, utils::Chain::Liquid, Some(&indexers))
+        .await
+        .unwrap();
     handle_alice.sdk.sync(false).await.unwrap();
 
     assert_eq!(handle_alice.get_pending_receive_sat().await.unwrap(), 0);
@@ -142,7 +165,10 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
         initial_balance - payer_amount_sat
     );
 
-    utils::mine_blocks(1).await.unwrap();
+    // Confirm the server lockup and wait for swap to complete
+    utils::mine_and_index_blocks(1, utils::Chain::Liquid, Some(&indexers))
+        .await
+        .unwrap();
 
     utils::wait_for_event_with_retry(
         &mut handle_alice,
@@ -153,9 +179,11 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
     .await
     .unwrap();
 
-    // TODO: this shouldn't be needed, but without it, sometimes get_balance_sat isn't updated in time
-    // https://github.com/breez/breez-sdk-liquid/issues/828
-    tokio::time::sleep(Duration::from_secs(10)).await;
+    // Workaround for #828: mine an extra Liquid block so that sync() sees a
+    // new tip and takes the full scan path, refreshing the LWK wallet state.
+    utils::mine_and_index_blocks(1, utils::Chain::Liquid, Some(&indexers))
+        .await
+        .unwrap();
     handle_alice.sdk.sync(false).await.unwrap();
 
     assert_eq!(handle_alice.get_pending_receive_sat().await.unwrap(), 0);
@@ -206,7 +234,10 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
         .await
         .unwrap();
 
-    utils::mine_blocks(1).await.unwrap();
+    // Confirm the liquid tx
+    utils::mine_and_index_blocks(1, utils::Chain::Liquid, Some(&indexers))
+        .await
+        .unwrap();
 
     // TODO: figure out why on Wasm this event is occasionally skipped
     // https://github.com/breez/breez-sdk-liquid/issues/847
@@ -226,9 +257,11 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
     .await
     .unwrap();
 
-    // TODO: this shouldn't be needed, but without it, sometimes get_balance_sat isn't updated in time
-    // https://github.com/breez/breez-sdk-liquid/issues/828
-    tokio::time::sleep(Duration::from_secs(10)).await;
+    // Workaround for #828: mine an extra Liquid block so that sync() sees a
+    // new tip and takes the full scan path, refreshing the LWK wallet state.
+    utils::mine_and_index_blocks(1, utils::Chain::Liquid, Some(&indexers))
+        .await
+        .unwrap();
     handle_alice.sdk.sync(false).await.unwrap();
     handle_bob.sdk.sync(false).await.unwrap();
 
--- a/lib/core/tests/regtest/bolt11.rs
+++ b/lib/core/tests/regtest/bolt11.rs
@@ -234,8 +234,6 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
         .await
         .unwrap();
 
-    // TODO: figure out why on Wasm this event is occasionally skipped
-    // https://github.com/breez/breez-sdk-liquid/issues/847
     let _ = utils::wait_for_event_with_retry(
         &mut handle_alice,
         &indexers,
--- a/lib/core/tests/regtest/bolt11.rs
+++ b/lib/core/tests/regtest/bolt11.rs
@@ -90,10 +90,9 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
         .await
         .unwrap();
 
-    assert_eq!(
-        handle_alice.get_pending_receive_sat().await.unwrap(),
-        receiver_amount_sat
-    );
+    handle_alice
+        .assert_wallet_pending(receiver_amount_sat, 0, 0)
+        .await;
 
     // Confirm the server lockup and wait for swap to complete
     utils::mine_and_index_blocks(1, utils::Chain::Liquid, Some(&indexers))
@@ -156,14 +155,9 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
         .await
         .unwrap();
 
-    assert_eq!(
-        handle_alice.get_pending_send_sat().await.unwrap(),
-        payer_amount_sat
-    );
-    assert_eq!(
-        handle_alice.get_balance_sat().await.unwrap(),
-        initial_balance - payer_amount_sat
-    );
+    handle_alice
+        .assert_wallet_pending(0, payer_amount_sat, initial_balance - payer_amount_sat)
+        .await;
 
     // Confirm the server lockup and wait for swap to complete
     utils::mine_and_index_blocks(1, utils::Chain::Liquid, Some(&indexers))
--- a/lib/core/tests/regtest/bolt11.rs
+++ b/lib/core/tests/regtest/bolt11.rs
@@ -116,20 +116,20 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle) 
         .unwrap();
     handle_alice.sdk.sync(false).await.unwrap();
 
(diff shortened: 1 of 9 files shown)


lib/core/tests/regtest/utils.rs
```diff
@@ -140,25 +219,347 @@ pub async fn pay_invoice_lnd(invoice: &str) -> Result<(), Box<dyn Error>> {
...
+pub async fn poll_boltz_zero_conf_accepted(
+    swap_id: &str,
+    timeout: Duration,
+) -> Result<bool, Box<dyn Error>> {
+    // Duplicate the id so Express always parses `ids` as an array
+    // (a single `?ids=X` is parsed as a string, not a one-element array).
+    let url = format!(
+        "{}/v2/swap/status?ids={}&ids={}",
+        SWAPPROXY_URL, swap_id, swap_id
+    );
```
hydra-yse, line 540: Nit:
We can use reqwest's [query](https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.query) method on the get request (returns a `RequestBuilder`) to make it more idiomatic.
If we need the `query` feature we can enable it as necessary

lib/core/tests/regtest/bolt11.rs
```diff
@@ -49,6 +59,14 @@ async fn bolt11(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle)
...
         .unwrap();
 
+    // Merge indexers from both handles: if either node uses an indexer, we must wait for it.
+    let indexers = utils::Indexers {
+        btc_esplora: handle_alice.indexers.btc_esplora || handle_bob.indexers.btc_esplora,
+        btc_electrs: handle_alice.indexers.btc_electrs || handle_bob.indexers.btc_electrs,
+        lbtc_esplora: handle_alice.indexers.lbtc_esplora || handle_bob.indexers.lbtc_esplora,
+        waterfalls: handle_alice.indexers.waterfalls || handle_bob.indexers.waterfalls,
+    };
+
```
hydra-yse: Nit: `Indexers::from_handles(&SdkNodeHandle, &SdkNodeHandle)` to avoid repetition?

lib/core/tests/regtest/bolt12.rs
```diff
@@ -110,33 +137,57 @@ async fn bolt12(mut handle_alice: SdkNodeHandle, mut handle_bob: SdkNodeHandle)
...
-    assert_eq!(handle_bob.get_pending_receive_sat().await.unwrap(), 0);
-    assert_eq!(handle_bob.get_pending_send_sat().await.unwrap(), 0);
-    assert_eq!(
-        handle_bob.get_balance_sat().await.unwrap(),
-        receiver_amount_sat
-    );
-
+    // For BOLT12 MRH payments, prepare_send_payment estimates a submarine swap fee.
+    // At send_payment time the SDK detects the MRH and routes via a direct Liquid tx,
+    // so the actual fee is lower. Read back from the Liquid Esplora indexer.
```
hydra-yse, line 165: Very good observation! Do you know if we're already checking for this on the SDK side?

hydra-yse (commented):
Looks really well thought-out, great job. It's probably even better than what I had in mind for the indexer API, I like the simplicity. @dangeross @danielgranhao Wdyt?

lib/core/tests/regtest/bolt11.rs
maggo83: Great idea! Added in 97c81fa6.

lib/core/tests/regtest/utils.rs
maggo83: In the regtest setup, helper HTTP calls currently go through the local SSL proxy for all test modes (WASM, native, node) to keep one common path.
Because of that, the swapproxy URL is passed as a value in X-Proxy-URL, and reqwest.query() would only affect the outer proxy request, not the inner swapproxy target URL.

So in this specific helper: manual construction of the inner URL is intentional.
But looking further: making helpers environment-aware could be a good follow-up:

-keep proxy path for WASM
-use direct swapproxy calls in non-WASM tests (->lower artificial latency, closer to production behavior)

Would you prefer this in the current PR or as a separate follow-up?

lib/core/tests/regtest/bolt12.rs
maggo83: You mean "does the SDK detect the lower fees for MRH and stores them in the swap"?
-> not that I could find. pay_liquid_onchain() verifies (similar to the test) that the fees are lower or equal than the original estimate. But they are not persisted/written back (instead the original estimate is stored/returned).

(Not sure if that is a "quirk" or should be considered a bug... would depend on how that value is used. If it is advertised as "real paid fee" it is a bug. If it is more a "upper limit" behavior would be fine)

hydra-yse: It's more of an upper limit. So long as we're not paying more, it's fine

PR #1099, opened 2026-06-16
Upgrade dependencies: lwk 0.18.0 and boltz-client 0.4.0

danielgranhao:
Upgrades the SDK's two major dependencies and refreshes transitive deps.

## Commits
- **Upgrade lwk to 0.18.0** — bumps the `breez/lwk` fork to `dffb6574` (rebased onto the upstream `wollet_0.18.0` tag; see fork branch `breez-sdk-liquid-0.18.0`).
- **Upgrade boltz-client to 0.4.0** — bumps to SatoshiPortal `2d52ad54`.
- **Update transitive dependencies** — `cargo update` on both lockfiles; pinned git revs and direct constraints unchanged.
- **regtest: update boltz submodule checkout**.

## lwk 0.18.0 migration
- `ElementsNetwork` → `Network` (`LiquidTestnet` → `TestnetLiquid`; regtest → `CustomElements(ElementsParams)`).
- **Wallet cache persistence** reworked: lwk removed the index-based `Persister`, so the cache now goes through a `DynStore` over a new `wallet_cache_kv` SQLite table, with lwk owning update serialization + encryption. The legacy `wallet_updates` table is dropped — existing installs rebuild the wallet cache on first sync after upgrade (rebuildable; the SDK already wipes-and-rescans on cache errors).
- Wollet construction moved to `WolletBuilder`; `PsetBalance::fee` → `fees_in()`; `WalletTx.balance` is now `SignedBalance`; `EsploraClientBuilder::build()` now returns `Result`.
- New required `Signer::sign_message` is stubbed (not reachable from SDK flows).
- The fork carries our node.js `async_sleep`, local path-imports, and no-change `TxBuilder` patches; the old 256-input cap is dropped in favour of upstream's `TooManyInputs`.

## boltz-client 0.4.0 migration
- Implement the new `LiquidClient`/`BitcoinClient::get_tx` trait methods.
- `Preimage::new()` → `Preimage::random()`.
- `get_chain_claim_tx_details()` now returns `Option` (None once the server has claimed) — handled.
- MuSig types moved from `secp256k1_zkp` to `secp256k1::musig`; added the matching `secp256k1_musig` dep to name them (boltz exposes them publicly but doesn't re-export the crate).
- Handle the new `Error::HTTPStatusNotSuccess` variant explicitly.

## Validation
- `make clippy` clean; builds on native + wasm (`node-js` + `browser`); bindings build.
- 126/126 unit tests pass.
- **Native regtest integration: 10/10 pass** (Electrum + Esplora — bolt11, bolt12, liquid, chain swaps).
- **Node.js wasm regtest integration: 4/4 pass** (Esplora — bolt11, bolt12, liquid, chain swaps; exercises the node-compatible `async_sleep`).

## Notes for reviewers
- The companion `breez/lwk` fork branch (`breez-sdk-liquid-0.18.0`) should be reviewed alongside this.

### DIFF
--- a/lib/core/Cargo.toml
+++ b/lib/core/Cargo.toml
@@ -23,8 +23,8 @@ derivative = "2.2.0"
 env_logger = "0.11"
 futures = "0.3.31"
 log = { workspace = true }
-lwk_common = { git = "https://github.com/breez/lwk", rev = "57f78f778038e716979c1cb3cd3c710d46ed657f" }
-lwk_signer = { git = "https://github.com/breez/lwk", rev = "57f78f778038e716979c1cb3cd3c710d46ed657f", default-features = false }
+lwk_common = { git = "https://github.com/breez/lwk", rev = "dffb6574c3c076cd984d327bd6a0e537fd323b87" }
+lwk_signer = { git = "https://github.com/breez/lwk", rev = "dffb6574c3c076cd984d327bd6a0e537fd323b87", default-features = false }
 mockall = "0.13.1"
 tokio = { version = "1", default-features = false, features = ["rt", "macros"] }
 tokio_with_wasm = { version = "=0.8.2", features = [
@@ -70,7 +70,7 @@ electrum-client = { version = "0.21.0", default-features = false, features = [
     "use-rustls-ring",
     "proxy",
 ] }
-lwk_wollet = { git = "https://github.com/breez/lwk", rev = "57f78f778038e716979c1cb3cd3c710d46ed657f" }
+lwk_wollet = { git = "https://github.com/breez/lwk", rev = "dffb6574c3c076cd984d327bd6a0e537fd323b87" }
 prost = "^0.11"
 tonic = { version = "^0.8", features = ["tls", "tls-webpki-roots"] }
 uuid = { version = "1.8.0", features = ["v4"] }
@@ -87,7 +87,7 @@ rusqlite = { git = "https://github.com/breez/rusqlite", branch = "rusqlite-v0.38
 [target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
 console_log = "1"
 sqlite-wasm-rs = { git = "https://github.com/breez/sqlite-wasm-rs", branch = "rusqlite-v0.38.0", default-features = false }
-lwk_wollet = { git = "https://github.com/breez/lwk", rev = "57f78f778038e716979c1cb3cd3c710d46ed657f", default-features = false, features = [
+lwk_wollet = { git = "https://github.com/breez/lwk", rev = "dffb6574c3c076cd984d327bd6a0e537fd323b87", default-features = false, features = [
     "esplora",
 ] }
 prost = "^0.13"
--- a/lib/core/Cargo.toml
+++ b/lib/core/Cargo.toml
@@ -23,8 +23,8 @@ derivative = "2.2.0"
 env_logger = "0.11"
 futures = "0.3.31"
 log = { workspace = true }
-lwk_common = { git = "https://github.com/breez/lwk", rev = "dffb6574c3c076cd984d327bd6a0e537fd323b87" }
-lwk_signer = { git = "https://github.com/breez/lwk", rev = "dffb6574c3c076cd984d327bd6a0e537fd323b87", default-features = false }
+lwk_common = { git = "https://github.com/breez/lwk", rev = "5f76812020e1e8f987c6bf26d4faf4ba429b1312" }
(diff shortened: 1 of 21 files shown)


lib/core/src/wallet/mod.rs
```diff
@@ -224,27 +220,23 @@ impl LiquidOnchainWallet {
...
+        let build_wollet = |persister: persister::LwkPersister| {
+            lwk_wollet::WolletBuilder::new(network, descriptor.clone())
+                .with_updates_store(persister)
+                .build()
+        };
+        let wollet_res = build_wollet(wallet_cache_persister.get_lwk_persister()?);
         match wollet_res {
             Ok(wollet) => Ok(wollet),
             res @ Err(
-                lwk_wollet::Error::PersistError(_)
```
dangeross, line 237: Now that `lwk_wollet::Error::PersistError` is gone, are the other cache-access failures it used to cover (store read/write errors, or decrypt/deserialize of a persisted update) mitigated some other way? The remaining arms only match `UpdateHeightTooOld`/`UpdateOnDifferentStatus`, so I wasn't sure whether the rest are expected to just propagate now.

roeierez (approved):
LGTM

lib/core/Cargo.toml
```diff
@@ -63,18 +63,21 @@ esplora-client = { git = "https://github.com/breez/rust-esplora-client", branch
...
+secp256k1_musig = { package = "secp256k1", version = "0.32.0-beta.0" }
 
 # Non-Wasm dependencies
 [target.'cfg(not(all(target_family = "wasm", target_os = "unknown")))'.dependencies]
 electrum-client = { version = "0.21.0", default-features = false, features = [
     "use-rustls-ring",
     "proxy",
 ] }
-lwk_wollet = { git = "https://github.com/breez/lwk", rev = "57f78f778038e716979c1cb3cd3c710d46ed657f" }
+lwk_wollet = { git = "https://github.com/breez/lwk", rev = "dffb6574c3c076cd984d327bd6a0e537fd323b87" }
```
dangeross: This bump also drops the fork's `Take max 256 UTXOs as inputs` patch (the `.take(256)` cap on lwk's UTXO selection) in favour of upstream's `TooManyInputs`. That changes behavior for wallets with more than 256 L-BTC UTXOs.

The #874 selection (`select_wallet_utxos` then `set_wallet_utxos`) bounds the input count for L-BTC sends, so that path is unaffected. But `build_drain_tx` and non-L-BTC asset sends never set the selected utxos, so `finish()` falls into lwk's `// FIXME: always add all L-BTC inputs` loop, which no longer caps at 256. For an asset send the L-BTC inputs are only for the fee, so a wallet with many small L-BTC UTXOs would now fail with `TooManyInputs` (surfaced as `PaymentError::Generic`) even though the tx needs only a couple of fee inputs, where the dropped cap previously let it build.

For amounts that genuinely need more than 256 inputs the error is correct, since 256 is the surjection-proof input limit, so this is only about the paths that skip selection. Is that acceptable, or worth bounding the drain and asset-fee paths too (or at least surfacing a clearer error than `Generic`)?

dangeross (commented):
Looks good. Just wondering about these error cases + the cache. Does it become unrecoverable?

lib/core/src/wallet/mod.rs
danielgranhao: Good point. I went over lwk, and now it produces generic errors for corruption issues. I've added a commit on our fork to use `StoreError` instead (good candidate to upstream, but leaving for later) and wired that error on our wipe-and-retry logic

lib/core/Cargo.toml
danielgranhao: Right! 

I've added a more descriptive error for the case where amounts genuinely need more than 256 inputs. Still used generic as I didn't want to change API/bindings in this PR.

For the asset send case, added a select_asset_and_fee_utxos that selects L-BTC utxos for fees.

lib/core/src/wallet/mod.rs
```diff
@@ -614,7 +702,7 @@ impl OnchainWallet for LiquidOnchainWallet {
                 if matches!(
                     e,
                     lwk_wollet::Error::UpdateHeightTooOld { .. }
-                        | lwk_wollet::Error::PersistError(_)
```
dangeross, line 617: Should we handle the StoreError here also?

dangeross (approved):
LGTM, just the other StoreError to check

lib/core/src/wallet/mod.rs
danielgranhao: Yes, done

PR #1102, opened 2026-07-03
Fix stale service-timeout Runnable disconnecting the shared SDK

### LINKED ISSUE #1101: Android notification plugin: stale service-timeout Runnable survives service destruction and disconnects the shared SDK of a newer service instance
Version: 0.12.4 (code unchanged on main)

TL;DR: ForegroundService.shutdown() calls stopSelf() without cancelling the pending 3-minute serviceTimeoutRunnable. That timer outlives the destroyed service instance, and when it fires it disconnects BreezSdkLiquidConnector — a process-wide singleton — out from under a newer service instance that's mid-job. Result: intermittent "Liquid SDK instance is not running" failures posted to the LNURL reply URL, so payers see failed payments.

How we hit it: We run a self-hosted LNURL-pay service (webhook → FCM → notification plugin). Payments to our users failed intermittently with {"status":"ERROR","reason":"Generic error: Liquid SDK instance is not running"} at the reply URL. LNURL-pay's two-push shape triggers it reliably enough to notice:

T+0: lnurlpay_info push starts the service; the job finishes in seconds; onStartCommand has armed the timeout for T+180s.
~T+65s: shutdownRunnable fires → disconnect + stopSelf(). The instance is destroyed, but its timeout callback stays queued on the main looper (Handler callbacks aren't tied to the Service lifecycle, and nothing removes them).
T+70–180s: the payer confirms the amount → lnurlpay_invoice push creates a new service instance with a fresh SDK connection.
T+180s: the dead instance's timer fires, calls shutdown() → shutdownSDK() on the shared connector (and stopSelf() against the live component). The invoice job's next SDK call throws, and the error is posted to the reply URL.
It only reproduces when the second push lands inside a dead instance's leftover window — hence "sometimes".

Workaround we ship: in our ForegroundService subclass, set a destroyed flag in onDestroy and override shutdown() to no-op once destroyed. This eliminated the failures for us.

Suggested fix: remove all pending callbacks from both handlers in shutdown() (or in onDestroy), and/or guard shutdown() so a destroyed instance can't touch BreezSdkLiquidConnector.

dangeross:
`ForegroundService` armed a 3-minute `serviceTimeoutRunnable` on every `onStartCommand` but never cancelled it on teardown (no `onDestroy`; `shutdown()` → `stopSelf()` left it queued). Because `Handler` callbacks live on the shared main `Looper` and aren't tied to the `Service` lifecycle, a timer from a **destroyed** instance would later fire and call `shutdownSDK()` on `BreezSdkLiquidConnector` — a process-wide singleton — tearing the SDK out from under a *newer* instance mid-job. This surfaced as intermittent `Liquid SDK instance is not running` failures (e.g. the LNURL-pay info→invoice two-push sequence).

### Changes
- **`onDestroy()`**: cancel both handlers' pending callbacks and the `serviceScope`, and set a `@Volatile destroyed` flag — so a dead instance's timers/coroutines can never touch the shared connector, even on teardown paths that don't go through `shutdown()`.
- **`shutdown()`**: no-op when `destroyed`, and clear both handlers up front so one timer's shutdown cancels its sibling (kills the latent double-shutdown).
- **Handler tidy-up**: post `serviceTimeoutRunnable` on `serviceTimeoutHandler` (it was mis-posted to `shutdownHandler`, leaving that handler unused).
- **Concurrency**: guard the remaining off-main-thread `jobs` accesses (`add`, `isEmpty` check) with `synchronized(sdkListener)` to match the existing lock discipline.

[Android build](https://github.com/breez/misty-breez/actions/runs/28673250290)

Fixes #1101

### DIFF
--- a/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/ForegroundService.kt
+++ b/lib/bindings/langs/android/lib/src/main/kotlin/breez_sdk_liquid_notification/ForegroundService.kt
@@ -37,6 +37,7 @@ import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 import kotlin.io.path.Path
 
@@ -55,6 +56,9 @@ abstract class ForegroundService :
     protected var logger: ServiceLogger = ServiceLogger()
     private var jobs: MutableList<Job> = arrayListOf()
 
+    @Volatile
+    private var destroyed = false
+
     companion object {
         private const val TAG = "ForegroundService"
     }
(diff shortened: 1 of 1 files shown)


danielgranhao (approved):

erdemyerebasmaz (approved):
👏

PR #1103, opened 2026-07-30
Recover from a stale wallet UTXO cache

dangeross:
Send swap lockups fail permanently with:

```
sendrawtransaction RPC error: {"code":-25,"message":"bad-txns-inputs-missingorspent"}
```

A reported wallet had 4 UTXOs that lwk listed as unspent but which a single confirmed tx had spent ~286 days earlier. Coin selection is deterministic branch-and-bound, so it re-picked the same coin on every attempt, meaning small payments could never succeed again.

The cache was internally inconsistent rather than merely stale: its *transaction* set was correct (the tx-derived balance matched the chain) and only `Wollet::utxos()` was wrong.

### Root cause

In lwk `Cache::update_unspent` re-adds the outputs of every tx in `txid_height_new`, filtered only by txid, with no check for whether a tx already in the cache spends them:

```rust
self.unblinded.keys().filter(|op| txids_new.contains(&op.txid))   // no spent-check
```

`inputs_new`, which removes spent outpoints, is computed from that same delta, so an existing spender is never consulted. Re-announcing a funding tx therefore resurrects its spent outputs permanently. The spender stays in `heights`, so balances stay correct and nothing surfaces it.

**Trigger:** anything that changes a funding tx's cached height once its outputs are spent, most commonly an output spent while unconfirmed whose funding tx then confirms. `utxos()` returns height-`None` outputs and our selection has no confirmation filter, so ordinary zero-conf change spending is enough.

### Fix

Detect the lost invariant, that no unspent outpoint may be spent by a known tx, and repair it:
1. **Local repair on rejection**: The spending tx is normally already held, so re-applying it makes lwk drop the stale inputs. Since that re-adds the spender's own outputs, it iterates until nothing changes, which resolves chains of spends. Instant, no network. The payment still fails, but the cache is corrected before the error returns, so a retry succeeds.
2. **Wipe otherwise**: If the tx set cannot explain the drift, schedule a cache wipe for the start of the next scan, which repopulates it. Left to the background, since a cold rescan takes minutes and must not stall a payment.
3. **Proactive check on startup**: So a wallet corrupted in an earlier session is repaired before a payment discovers it. Both entry points are needed: the trigger is a scan outcome, so corruption appears mid-session as well as across restarts.

Plus `apply_broadcast_tx` after every successful broadcast, closing the window where a second payment re-selects coins already in flight. This uses `Wollet::apply_transaction` to notify lwk of a transaction broadcast before the sync so it can update its UTXO cache.

Note that `test_lwk_resurrects_outputs_of_reannounced_funding_tx` asserts the lwk bug is *present*, reproducing it against a real `Wollet`. It will fail once lwk fixes this, which is deliberate: the failure message says so and points at the recovery code to reassess.

### DIFF
--- a/lib/core/src/send_swap.rs
+++ b/lib/core/src/send_swap.rs
@@ -23,7 +23,9 @@ use crate::prelude::{PaymentTxData, PaymentType, Swap};
 use crate::recover::recoverer::Recoverer;
 use crate::swapper::Swapper;
 use crate::utils;
-use crate::wallet::OnchainWallet;
+use crate::wallet::{
+    handle_stale_cache_broadcast_error, should_retry_after_cache_repair, OnchainWallet,
+};
 use crate::{
     error::PaymentError,
     model::{PaymentState, Transaction as SdkTransaction},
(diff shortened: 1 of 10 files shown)


lib/core/src/send_swap.rs
```diff
@@ -244,33 +246,51 @@ impl SendSwapHandler {
...
+                    if should_retry_after_cache_repair(
+                        &*self.onchain_wallet,
+                        &err,
+                        &mut repaired_cache,
+                    )
+                    .await?
+                    {
+                        continue;
+                    }
+                    return Err(handle_stale_cache_broadcast_error(err));
```
roeierez: Since this is clearly an issue in lwk wallet that needs to be solved (prevent override of spent tx in the cache) and we should expect such fix soon. Perhaps it is sufficient to leave the repair in full_scan (also restart) only rather than in the retry mechanism? WDYT Ross?

roeierez (changes requested):
Well analyzed and looks good!

lib/core/src/wallet/mod.rs
```diff
@@ -662,73 +773,166 @@ impl OnchainWallet for LiquidOnchainWallet {
...
                     &self.signer,
                     self.wallet_cache_persister.clone(),
                 )
                 .await?;
-                client
-                    .full_scan_to_index(&mut new_wallet, index_with_buffer)
-                    .await?;
-                *wallet = new_wallet;
-                Ok(())
+                Some(FlagGuard(&self.recovery_scan_in_progress))
```
danielgranhao: If clear_cache or create_wallet fail, we return and `recovery_scan_in_progress` stays true until SDK restart right? That will cause tx building to fail until the SDK is restarted. Should be solved by moving the guard creation to right after the `recovery_scan_in_progress.store()` call.

Also, if create_wallet fails after clear_wallet succeeds, the old wallet with stale utxo(s) will remain. Shouldn't we only clear `needs_cache_clear` after both?

lib/core/src/wallet/mod.rs
```diff
@@ -757,6 +961,384 @@ mod tests {
...
+        let resurrected = has_utxo(&w, a0)?;
+        let spender_known = w.transactions()?.iter().any(|t| t.txid == spend.txid());
+
+        // Asserts the bug is still PRESENT, so this fails if lwk ever fixes it. That is the
+        // point: it is the signal to revisit the recovery code, not a regression in this crate.
+        assert!(
+            resurrected && spender_known,
+            "lwk no longer resurrects outputs of a re-announced funding tx \
+             (resurrected={resurrected}, spender_known={spender_known}). If this failed after an \
+             lwk bump, the upstream bug is likely fixed: see LWK_ISSUE.md and reassess whether \
```
danielgranhao: LWK_ISSUE.md was probably a temporary file?

lib/core/src/wallet/mod.rs
```diff
@@ -662,73 +773,166 @@ impl OnchainWallet for LiquidOnchainWallet {
...
+        let remaining = find_spent_utxos(&wallet.transactions()?, &wallet.utxos()?);
+        if remaining.is_empty() {
+            info!("Wallet cache repaired locally, no rescan needed");
+            Ok(true)
+        } else {
+            warn!("Local wallet cache repair left {remaining:?} unresolved, flagging for a rescan");
+            // Only a wipe fixes this now; schedule it rather than stall the call in flight.
+            self.needs_cache_clear.store(true, Ordering::Relaxed);
+            Ok(false)
+        }
```
danielgranhao: When we re-apply a spender, we might also be adding a change output that was already spent, so we end up reintroducing the issue. We could repeat the repair step until nothing changes to avoid needing to clear the cache.

lib/core/src/wallet/mod.rs
```diff
@@ -212,9 +277,43 @@ impl LiquidOnchainWallet {
...
+            return Ok(());
+        }
+
+        error!(
+            "Wallet cache is inconsistent: {} utxo(s) are already spent by known txs. {spent:?}",
+            spent.len()
+        );
+
+        // The local repair usually suffices; if not it schedules a wipe for the next scan.
+        self.repair_cache().await?;
```
danielgranhao, line 339: Could we log error and ignore? Otherwise we skip syncing payments because full scan is the first thing on `sync_inner`

danielgranhao (approved):
LGTM

PR #1104, opened 2026-07-30
Fix clippy lints from newer toolchain

dangeross:
CI runs clippy against current stable, 1.97 at time of writing, while the repo pins no toolchain. Two lints that did not exist when this code was written now fail the build:

- `useless_borrows_in_formatting`: a redundant `&` in a log or format argument (8 sites)
- `for_kv_map`: `for (_, v) in map.iter()` should be `map.values()` (2 sites)

All ten sites are pre-existing and unrelated to any in-flight feature work, so this is split out rather than folded into whichever PR happened to surface it. No behaviour change.

Surfaced by the CI run on #1103, which fails on these same lints despite touching none of the lines.

Note this could not be verified locally: `useless_borrows_in_formatting` does not exist in the newest toolchain installed here (1.95), so the changes are clippy's own suggestions applied verbatim. Tests, fmt and the older clippy all pass; CI is the real check.

Worth considering separately: a `rust-toolchain.toml` would stop CI drifting under contributors, at the cost of needing deliberate bumps.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/core/src/chain_swap.rs
+++ b/lib/core/src/chain_swap.rs
@@ -357,12 +357,12 @@ impl ChainSwapHandler {
                         if let Err(e) = self.handle_amountless_update(swap).await {
                             // In case of error, we log the error but don't mark as refundable,
                             // letting the recovery logic handle the state.
-                            error!("Failed to accept the quote for swap {}: {e:?}", &swap.id);
+                            error!("Failed to accept the quote for swap {}: {e:?}", swap.id);
                         }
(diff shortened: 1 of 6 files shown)


danielgranhao:
I think it's worth adding the `rust-toolchain.toml` file. That's the approach we now have on the spark repo already.

danielgranhao (approved):

